PR #2, opened 2026-01-31
refactor(email): drop iterable provider

GuiBibeau:
## Summary
- remove Iterable email provider and related env/config
- keep email provider abstraction (resend/console only)
- drop Iterable defaults from wrangler config

## Testing
- pnpm exec biome check --changed --since=origin/main --no-errors-on-unmatched


GuiBibeau:
@jacobcreech. i tried using this technique to trigger emails from our verified addresses and i get an unverified error. 

Does this align with what you guys do when you use Iterable on non marketing campaings?

jacobcreech:
> Does this align with what you guys do when you use Iterable on non marketing campaings?

Usually I just get 200ok. Let me read a doc on this

PR #8, opened 2026-02-03
feat(api): draft payments openapi schema

GuiBibeau:
## Summary
- add draft Payments OpenAPI schemas (wallets, transfers, Solana Pay, ramps) using Zod
- register draft Payments paths/tags in the generated spec
- mark every Payments endpoint as DRAFT / not implemented for review

## Testing
- not run


apps/sdp-api/src/openapi/paths/payments.ts
```diff
@@ -0,0 +1,384 @@
...
+    method: "get",
+    path: "/v1/payments/transfers",
+    tags: ["Payments"],
+    summary: "List transfers",
+    operationId: "listPaymentTransfers",
+    description: withDraft("Lists transfers for the organization."),
+    security: [{ apiKeyAuth: [] }],
+    request: {
+      query: z.object({
+        wallet: z.string().optional().openapi({ description: "Filter by wallet ID." }),
```
gitteri, line 248: would be good to have a filter for wallet addresses too

apps/sdp-api/src/openapi/paths/payments.ts
```diff
@@ -0,0 +1,384 @@
...
+    method: "get",
+    path: "/v1/payments/transfers/{transferId}",
+    tags: ["Payments"],
+    summary: "Get transfer",
+    operationId: "getPaymentTransfer",
+    description: withDraft("Retrieves details for a specific transfer."),
+    security: [{ apiKeyAuth: [] }],
+    request: {
+      params: z.object({
+        transferId: transferIdParamSchema,
```
gitteri, line 303: is this a db id or onchain id?

apps/sdp-api/src/openapi/paths/payments.ts
```diff
@@ -0,0 +1,384 @@
...
+    path: "/v1/payments/fees/quote",
+    tags: ["Payments"],
+    summary: "Get fee quote",
+    operationId: "getPaymentFeeQuote",
+    description: withDraft("Retrieves a fee quote for a transfer."),
+    security: [{ apiKeyAuth: [] }],
+    request: {
+      query: z.object({
+        token: z.string().openapi({ description: "Fee token symbol or mint address." }),
+        amount: z.string().openapi({ description: "Transfer amount." }),
```
gitteri: why does amount impact fees?

apps/sdp-api/src/openapi/paths/payments.ts
```diff
@@ -0,0 +1,384 @@
...
+  // ═══════════════════════════════════════════════════════════════════════════
+  // Ramps
+  // ═══════════════════════════════════════════════════════════════════════════
+
+  registry.registerPath({
+    method: "post",
+    path: "/v1/payments/ramps/quote",
+    tags: ["Payments"],
+    summary: "Get ramp quote",
+    operationId: "getPaymentRampQuote",
```
gitteri: we likely want to split on/off-ramp into two separate apis (this seems to be standard w/on & offramp providers apis). you can take a look at some of them

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+  .object({
+    provider: custodyProviderSchema,
+    config: z
+      .record(z.unknown())
+      .optional()
+      .openapi({ description: "Provider-specific configuration payload." }),
+  })
+  .openapi({ description: "Custody configuration." });
+
+export const createWalletRequestSchema = z
```
gitteri, line 36: item for the backlog:
* wallet policies - configure specific rulesets for a given wallet (destination allowlist etc)

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+
+export const walletBalancesSchema = z
+  .object({
+    walletId: walletIdParamSchema,
+    address: solanaAddressSchema.openapi({ description: "Wallet address." }),
+    balances: z.array(tokenBalanceSchema).openapi({ description: "Token balances." }),
+  })
+  .openapi({ description: "Wallet balances payload." });
+
+export const gaslessConfigSchema = z
```
gitteri: what's this for?

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+export const createTransferRequestSchema = z
+  .object({
+    source: z.string().openapi({ description: "Source wallet ID." }),
+    destination: solanaAddressSchema.openapi({ description: "Destination wallet address." }),
+    token: z
+      .string()
+      .optional()
+      .openapi({ description: "Token symbol or mint address (omit for SOL)." }),
+    amount: tokenAmountSchema,
+    gasless: gaslessConfigSchema.optional(),
```
gitteri: 1. let's not use the term "gasless"
2. everything should just be free in the SDP to start. in the future we can add a "pay for user fee" option which will automatically add a fee in the sending token to cover the transaction fee

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+    memo: z
+      .string()
+      .max(256)
+      .optional()
+      .openapi({ description: "Optional memo for the transfer." }),
+  })
+  .openapi({ description: "Create transfer request payload." });
+
+export const priorityFeeSchema = z
```
gitteri, line 153: this should go into the fee estimate api endpoint instead of amount

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+  .openapi({ description: "Priority fee level.", example: "auto" });
+
+export const prepareTransferRequestSchema = z
+  .object({
+    source: z.string().openapi({ description: "Source wallet ID or pubkey." }),
+    destination: solanaAddressSchema.openapi({ description: "Destination wallet address." }),
+    token: z
+      .string()
+      .optional()
+      .openapi({ description: "Token symbol or mint address (omit for SOL)." }),
```
gitteri: i think this should just be required

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+    amount: tokenAmountSchema,
+    memo: z
+      .string()
+      .max(256)
+      .optional()
+      .openapi({ description: "Optional memo for the transfer." }),
+    reference: solanaAddressSchema
+      .optional()
+      .openapi({ description: "Optional reference key for tracking." }),
```
gitteri: nit: reference address account? i know key is what solana pay uses but that term always confused me

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+    organizationId: orgIdParamSchema,
+    type: transferTypeSchema,
+    status: transferStatusSchema,
+    signature: z
+      .string()
+      .nullable()
+      .openapi({ description: "Solana transaction signature.", example: "sig_example" }),
+    serializedTx: base64Schema.nullable().openapi({
+      description: "Base64-encoded transaction payload, if available.",
+      example: "AQID",
```
gitteri: AQID?

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+  .enum(["pending", "processing", "confirmed", "finalized", "failed"])
+  .openapi({ description: "Transfer status.", example: "confirmed" });
+
+export const transferSchema = z
+  .object({
+    id: transferIdParamSchema,
+    organizationId: orgIdParamSchema,
+    type: transferTypeSchema,
+    status: transferStatusSchema,
+    signature: z
```
gitteri: signature = tx id yea? not a transaction signature? i'm debating if we should call this transaction hash / id

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+      .string()
+      .nullable()
+      .openapi({ description: "Solana transaction signature.", example: "sig_example" }),
+    serializedTx: base64Schema.nullable().openapi({
+      description: "Base64-encoded transaction payload, if available.",
+      example: "AQID",
+    }),
+    params: z
+      .record(z.unknown())
+      .openapi({ description: "Operation parameters captured for audit." }),
```
gitteri: ?

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+      .nullable()
+      .openapi({ description: "Slot number, if confirmed.", example: 123456 }),
+    blockTime: isoDateTimeSchema
+      .nullable()
+      .openapi({ description: "Block time, if confirmed.", example: "2025-01-01T00:00:00.000Z" }),
+    fee: z
+      .number()
+      .int()
+      .nullable()
+      .openapi({ description: "Transaction fee in lamports.", example: 5000 }),
```
gitteri, line 260: debating if we want this in usd / what we charged the end user

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+    createdAt: isoDateTimeSchema.openapi({
+      description: "Creation timestamp.",
+      example: "2025-01-01T00:00:00.000Z",
+    }),
+    updatedAt: isoDateTimeSchema.openapi({
+      description: "Last update timestamp.",
+      example: "2025-01-02T00:00:00.000Z",
+    }),
+  })
+  .openapi({ description: "Transfer transaction record." });
```
gitteri, line 291: we're missing memo

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+    fee: z
+      .number()
+      .int()
+      .nullable()
+      .openapi({ description: "Transaction fee in lamports.", example: 5000 }),
+    error: z.string().nullable().openapi({
+      description: "Error message if the transaction failed.",
+      example: "Signature failed",
+    }),
+    initiatedByKeyId: apiKeyIdParamSchema
```
gitteri: if it was initiated by a UI user what goes here?

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+  .openapi({ description: "Transfer type.", example: "transfer" });
+
+export const transferStatusSchema = z
+  .enum(["pending", "processing", "confirmed", "finalized", "failed"])
+  .openapi({ description: "Transfer status.", example: "confirmed" });
+
+export const transferSchema = z
+  .object({
+    id: transferIdParamSchema,
+    organizationId: orgIdParamSchema,
```
gitteri, line 233: are you ready for project level as well? or we can add that later

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+  })
+  .openapi({ description: "Create confidential transfer request payload." });
+
+export const createPaymentRequestRequestSchema = z
+  .object({
+    recipient: z.string().openapi({ description: "Recipient wallet ID." }),
+    amount: tokenAmountSchema,
+    token: z.string().openapi({ description: "Token symbol or mint address." }),
+    label: z.string().max(64).optional().openapi({ description: "Merchant/recipient name." }),
+    message: z
```
gitteri: is this the memo? if so let's call it that

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+      .string()
+      .max(256)
+      .optional()
+      .openapi({ description: "Payment request description." }),
+    expiresIn: z
```
gitteri: i like this. note that you'll have to use nonce accounts for anything more than 60s...

fyi the flow here is:
1. create payment request
2. backend creates the transaction and signs it as fee payer (add optional fee to fee payer to cover costs)
3. receive payment request 
4. forward to sender
5. have them sign and submit

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+      .optional()
+      .openapi({ description: "Expiration time in seconds.", example: 3600 }),
+  })
+  .openapi({ description: "Create payment request payload." });
+
+export const paymentRequestStatusSchema = z
+  .enum(["pending", "fulfilled", "expired", "cancelled"])
+  .openapi({ description: "Payment request status.", example: "pending" });
+
+export const paymentRequestSchema = z
```
gitteri: i'm wondering if we cut this from v1. basic payment requests can't be gasless (unless we update solana pay to my v1.2 spec https://github.com/solana-foundation/solana-pay/pull/270)

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+      example: "2025-01-01T01:00:00.000Z",
+    }),
+  })
+  .openapi({ description: "Payment request details." });
+
+export const feeQuoteSchema = z
+  .object({
+    feeToken: z.string().openapi({ description: "Fee token symbol.", example: "USDC" }),
+    feeAmount: tokenAmountSchema,
+    feeInLamports: z.number().int().openapi({ description: "Fee in lamports.", example: 5000 }),
```
gitteri: i think we just do fee token and fee amount

apps/sdp-api/src/openapi/schemas/payments.ts
```diff
@@ -0,0 +1,403 @@
...
+
+export const rampDirectionSchema = z
+  .enum(["onramp", "offramp"])
+  .openapi({ description: "Ramp direction.", example: "onramp" });
+
+export const rampProviderSchema = z
+  .enum(["bridge", "moonpay", "ramp", "transak"])
+  .openapi({ description: "Ramp provider.", example: "moonpay" });
+
+export const rampQuoteRequestSchema = z
```
gitteri: note: ignoring this section because we should split this into two endpoints (on & off). the flows, requests, responses are all pretty different

gitteri (changes requested):

gitteri:
We need general ability to track inbound deposits as well. this can go on the transfers endpoint or a new one. not sure yet which is better so let's just put it on transfers w/a filter & pagination

note: for we want to include a risk score for inbound / outbound transfers

apps/sdp-api/src/openapi/paths/payments.ts
GuiBibeau: i removed it.

apps/sdp-api/src/openapi/paths/payments.ts
GuiBibeau: i plit them into:
POST /v1/payments/ramps/onramp/quote
POST /v1/payments/ramps/onramp/execute
POST /v1/payments/ramps/offramp/quote
POST /v1/payments/ramps/offramp/execute

apps/sdp-api/src/openapi/schemas/payments.ts
GuiBibeau: added to the schema to know to implement

apps/sdp-api/src/openapi/schemas/payments.ts
GuiBibeau: removed

apps/sdp-api/src/openapi/schemas/payments.ts
GuiBibeau: made a better example

apps/sdp-api/src/openapi/schemas/payments.ts
GuiBibeau: i clarified in the description. I'm thinking it's worth a whole discussion since changing signature here would mean changing in other areas also. Might be good to discuss on a specific issue

apps/sdp-api/src/openapi/schemas/payments.ts
GuiBibeau: The idea was to keep something flexible but removing for now. We can get deeper on implementation details

apps/sdp-api/src/openapi/schemas/payments.ts
GuiBibeau: for simpler accounting?

apps/sdp-api/src/openapi/schemas/payments.ts
GuiBibeau: good point, i'm going to make that a bit more flexible with system / userId / api. to be flexible

apps/sdp-api/src/openapi/schemas/payments.ts
GuiBibeau: Not quite ready yet. As we solidify the Auth layer i think adding in would be quite simple. I'm using Clerk now to avoid us doing the full auth as i did initially. This should make it quite easy.

GuiBibeau: added project for now and when implemnting i'll add it. Not too worried about complexity

apps/sdp-api/src/openapi/schemas/payments.ts
GuiBibeau: Good point, I captured it in a comment for when we do implementation

apps/sdp-api/src/openapi/schemas/payments.ts
GuiBibeau: yeah good point. I think pay is also right now pretty in flux in between commerce kit pay gill pay kit? It would be good to get it up to speed maybe as a collateral from our work on SDP.

I'll cut it for now

apps/sdp-api/src/openapi/schemas/payments.ts
gitteri: yeah. we have nothing else in sol. maybe both?

gitteri (approved):

PR #20, opened 2026-02-11
Payments PR2: transfer history list/get endpoints

GuiBibeau:
## Scope (Payments Part 2)
Refactors Payments transfer history to an RPC-first model while keeping D1 for pending transfer state and wallet policies.

### Endpoints covered
- `GET /v1/payments/transfers`
- `GET /v1/payments/transfers/{transferId}`
- `POST /v1/payments/transfers/prepare` (response shape update)
- `POST /v1/payments/transfers` (execution response alignment)

## What this PR adds
- RPC-native transfer history listing/get behavior with ownership enforcement.
- Dedicated RPC history helper module to keep route handlers readable:
  - `apps/sdp-api/src/routes/payments/rpc-history.ts`
- Exact pagination totals (`meta.total`) after filters are applied.
- Batch transaction hydration (`getTransaction`) with compatibility fallback.
- Security hardening:
  - rejects unowned `walletAddress` filters
  - returns `NOT_FOUND` when a signature does not involve org-owned wallets
- Policy enforcement improvements:
  - `maxDailyAmount` now uses same-day outbound on-chain totals (RPC)
  - plus pending transfer totals from D1 (`pending`/`processing`)

## Data model decisions
- Keep D1 as **pending transfer state** + **wallet policy storage**.
- Transfer history/confirmed execution is sourced from RPC.
- Remove non-essential transfer lifecycle update/read paths that caused complexity and bugs.
- Keep DB schema non-destructive in this PR (no table drops).

## API / OpenAPI changes
- `POST /v1/payments/transfers/prepare`
  - no longer returns `data.transfer`
  - returns `preparedTransaction` (+ optional `simulation`)
- `GET /v1/payments/transfers/{transferId}`
  - accepts on-chain signature
  - preserves pending transfer lookup via `xfr_*` IDs
- Executed/history transfer IDs are signature-based (`id === signature`)
- OpenAPI schemas/paths updated accordingly.

## Tests
- Updated unit coverage in:
  - `apps/sdp-api/src/routes/payments.test.ts`
- Added devnet integration coverage in the shared integration suite:
  - `packages/sdp-api-integration/src/tests/payments-history.test.ts`
- Integration suite is unified under one command (includes Kora tests):
  - `pnpm test:integration`

## Validation
- `pnpm --filter @sdp/api exec vitest run src/routes/payments.test.ts`
- `pnpm --filter @sdp/api exec biome check src/routes/payments.test.ts vitest.config.ts`
- `pnpm --filter @sdp/api-integration exec biome check src/tests/payments-history.test.ts`
- `pnpm test:integration` (requires devnet/Kora env vars)


apps/sdp-api/src/routes/payments/handlers.ts
```diff
@@ -469,43 +1042,17 @@ export async function getWalletBalances(c: AppContext) {
...
-    walletId: wallet.walletId,
-    mode: row.mode,
-    destinationAllowlist: parseDestinationAllowlist(row.destination_allowlist),
-    ...(row.max_transfer_amount ? { maxTransferAmount: row.max_transfer_amount } : {}),
-    ...(row.max_daily_amount ? { maxDailyAmount: row.max_daily_amount } : {}),
-    createdAt: row.created_at,
-    updatedAt: row.updated_at,
-  };
+  const rows = await repository.getWalletPoliciesByCustodyWalletId(wallet.id);
+  const payload = buildWalletPolicyPayload(wallet.walletId, rows, wallet.createdAt);
```
gitteri, line 1211: an easier way here would be to just get latest unique type policies from the db to avoid this "build policy" thing and then return them as individual policies

apps/sdp-api/src/routes/payments/handlers.ts
```diff
@@ -755,3 +1251,230 @@ export async function createTransfer(c: AppContext) {
...
+    merged = merged.filter((row) => {
+      const iso = unixSecondsToIso(row.blockTime);
+      return iso ? Date.parse(iso) <= maxDate : false;
+    });
+  }
+
+  const offset = (page - 1) * pageSize;
+  const slicedPlusOne = merged.slice(offset, offset + pageSize + 1);
+  const hasMore = slicedPlusOne.length > pageSize;
+  const pageItems = hasMore ? slicedPlusOne.slice(0, pageSize) : slicedPlusOne;
```
gitteri: this function is wild. we're getting transfers from an rpc and matching them against our db and getting details there? there's gotta be a better way. this is to avoid indexing deposits? let's chat for 2m about this

apps/sdp-api/src/routes/payments/handlers.ts
```diff
@@ -755,3 +1251,230 @@ export async function createTransfer(c: AppContext) {
...
+
+  let filteredTransfers = transfers;
+  if (token) {
+    filteredTransfers = filteredTransfers.filter((t) => t.token === token);
+  }
+  if (direction) {
+    filteredTransfers = filteredTransfers.filter((t) => t.direction === direction);
+  }
+
+  const total = hasMore ? offset + filteredTransfers.length + 1 : offset + filteredTransfers.length;
```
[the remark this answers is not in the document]
GuiBibeau: Addressed. Token and direction filters are now applied before pagination and meta.total is computed from the fully filtered set in apps/sdp-api/src/routes/payments/handlers.ts (requiresHydratedFiltering branch). Added regression coverage in apps/sdp-api/src/routes/payments.test.ts with the exact-total filtered list test.

apps/sdp-api/src/routes/payments/handlers.ts
```diff
@@ -755,3 +1251,230 @@ export async function createTransfer(c: AppContext) {
...
+  const selectedWalletAddresses = new Set<string>();
+  if (wallet) {
+    selectedWalletAddresses.add(resolveWallet(scope.wallets, wallet).publicKey);
+  }
+  if (walletAddress) {
+    selectedWalletAddresses.add(assertValidAddress(walletAddress, "walletAddress"));
+  }
+  if (!wallet && !walletAddress) {
+    for (const w of scope.wallets) selectedWalletAddresses.add(w.publicKey);
+  }
```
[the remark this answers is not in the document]
GuiBibeau: Addressed. walletAddress is now validated against org-owned addresses and rejected with BAD_REQUEST when unowned in apps/sdp-api/src/routes/payments/handlers.ts. Covered by the regression test in apps/sdp-api/src/routes/payments.test.ts: rejects listTransfers walletAddress when address is not owned by authenticated org.

apps/sdp-api/src/routes/payments/handlers.ts
GuiBibeau: Addressed via refactor to an RPC-first history flow. Transfer-history hydration and parsing moved into apps/sdp-api/src/routes/payments/rpc-history.ts, and list/get handlers no longer merge confirmed history from D1. D1 is retained for pending transfer state and wallet policies.

apps/sdp-api/src/routes/payments/handlers.ts
```diff
@@ -755,3 +1251,230 @@ export async function createTransfer(c: AppContext) {
...
+  const rpc = createRpc(c.env);
+
+  const tx = await getTransactionJsonParsed(rpc, transferId);
+
+  if (!tx) {
+    throw new AppError("NOT_FOUND", "Transfer not found");
+  }
+
+  const walletAddress = scope.wallets[0]?.publicKey ?? "";
+  const details = inferOnChainTransferDetails(tx, walletAddress);
```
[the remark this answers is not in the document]
GuiBibeau: Addressed. Signature-based getTransfer now enforces ownership by requiring the transaction to touch an org-owned wallet before returning data (touchesOwnedWallet check in apps/sdp-api/src/routes/payments/handlers.ts). Added regression coverage in apps/sdp-api/src/routes/payments.test.ts: returns NOT_FOUND when transfer signature does not involve any org wallet.

apps/sdp-api/src/routes/payments/handlers.ts
```diff
@@ -755,3 +1251,230 @@ export async function createTransfer(c: AppContext) {
...
+  const transferRowsBySignature = await getTransferRowsBySignatures(
+    c,
+    pageItems.map((row) => row.signature),
+    scope.auth.organizationId,
+    scope.auth.projectId
+  );
+
+  const transfers = await Promise.all(
+    pageItems.map(async (row) => {
+      const tx = await getTransactionJsonParsed(rpc, row.signature);
```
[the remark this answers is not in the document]
GuiBibeau: Addressed. Transaction hydration now uses batched JSON-RPC calls (getTransactionsJsonParsedBatch) with chunking and compatibility fallback in apps/sdp-api/src/routes/payments/rpc-history.ts, and list handlers call this batch path rather than per-item sequential fetches.

apps/sdp-api/src/routes/payments/handlers.ts
```diff
@@ -261,6 +263,156 @@ function getUtcDayWindow(now: Date): { start: string; end: string } {
...
+
+      if (rows.length === 0) {
+        break;
+      }
+
+      for (const row of rows) {
+        const blockTimeIso = unixSecondsToIso(row.blockTime);
+        if ((minDate !== null || maxDate !== null) && !blockTimeIso) {
+          continue;
+        }
```
gitteri, line 320: i'm confused why this is needed. if theres no block time feels like we should error out?

apps/sdp-api/src/routes/payments/handlers.ts
```diff
@@ -261,6 +263,156 @@ function getUtcDayWindow(now: Date): { start: string; end: string } {
...
+
+        if (blockTimeIso) {
```
gitteri, line 322: is there any reasonable expectation that a blocktime can be nill on a row in the response of this api?

apps/sdp-api/src/routes/payments/handlers.ts
```diff
@@ -261,6 +263,156 @@ function getUtcDayWindow(now: Date): { start: string; end: string } {
...
+          const blockTimeDate = Date.parse(blockTimeIso);
+          if (minDate !== null && blockTimeDate < minDate) {
+            continue;
+          }
+          if (maxDate !== null && blockTimeDate > maxDate) {
+            continue;
```
gitteri, line 328: does listSigs api return in block order? if so we can break here instead

apps/sdp-api/src/routes/payments/handlers.ts
```diff
@@ -261,6 +263,156 @@ function getUtcDayWindow(now: Date): { start: string; end: string } {
...
+            ...row,
+            queriedAddress: address,
+          });
+        }
+      }
+
+      const last = rows[rows.length - 1];
+      if (!last) {
+        break;
+      }
```
gitteri, line 344: why?

apps/sdp-api/src/routes/payments/handlers.ts
```diff
@@ -261,6 +263,156 @@ function getUtcDayWindow(now: Date): { start: string; end: string } {
...
+  options: { from?: string; to?: string }
+): Promise<ScopedSignatureRow[]> {
+  const minDate = options.from ? Date.parse(options.from) : null;
+  const maxDate = options.to ? Date.parse(options.to) : null;
+  const bySignature = new Map<string, ScopedSignatureRow>();
+
+  for (const address of addresses) {
+    let before: string | undefined;
+
+    while (true) {
```
gitteri, line 306: we can end up in an infinite loop a few different ways i think here including if min is null. worth rethinking this function and remove a while true condition here

gitteri (changes requested):

gitteri:
@greptileai review the `listScopedSignaturesForAddresses` function for correctness and to find cases where you can be stuck in an infinite loop

GuiBibeau:
going back to drawing board

PR #170, opened 2026-04-10
fix: show wallet deposits in recent transactions

GuiBibeau:
## Summary
- Include owned SPL token accounts when discovering wallet-scoped transfer signatures.
- Reuse the existing observed-transfer parser so custom token deposits without SDP DB rows appear in wallet recent activity.
- Add a regression test for a custom-token deposit whose signature is only discoverable through the destination token account.

## Linear
- PRO-1101

## Validation
- pnpm --filter @sdp/api lint
- pnpm --filter @sdp/api typecheck
- pnpm --filter @sdp/api test -- src/routes/payments.test.ts -t "list transfers"


GuiBibeau:
@greptileai please re-review after bounding token-account signature history fanout.

GuiBibeau:
@greptileai please re-review after parallelizing token-account discovery.

GuiBibeau:
@greptileai please re-review after adding observed Token-2022 mint deposits to wallet activity.

PR #239, opened 2026-04-29
fix(web): use configured RPC for devnet faucet

GuiBibeau:
## Summary
- route the devnet faucet `requestAirdrop` call through SDP's `/v1/rpc` relay
- let the faucet inherit the configured organization/project RPC provider from settings
- include the resolved RPC provider in faucet error messages when the upstream rejects the request

## Root cause
The faucet server action was hardcoded to `https://api.devnet.solana.com`. A direct probe showed that endpoint returned a JSON-RPC `Internal error` for `requestAirdrop` while `getHealth` still returned `ok`, so the faucet was bypassing the configured company RPC and depending on the public devnet endpoint.

## Validation
- `pnpm --filter sdp-web typecheck`
- `git diff --check`

## Notes
- `pnpm --filter sdp-web lint` is currently blocked before it reaches source files because generated `.next-playwright` output triggers an existing ESLint/plugin compatibility error.

### DIFF
--- a/apps/sdp-web/src/app/dashboard/custody/actions.ts
+++ b/apps/sdp-web/src/app/dashboard/custody/actions.ts
@@ -434,18 +434,15 @@ export async function requestDevnetSolanaFaucetAction(
       return { status: "error", message: "Sign in to request devnet SOL." };
     }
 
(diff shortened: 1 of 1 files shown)


GuiBibeau:
@greptileai review

GuiBibeau:
@greptileai review latest commit

PR #242, opened 2026-04-29
fix(api): render Worker vars from Doppler

GuiBibeau:
## Summary
- render committed Worker runtime vars from the fetched Doppler environment into the generated Wrangler config
- fail deploy rendering when production Worker vars still contain placeholder Kora/RPC URLs
- include SDP_DEPLOYMENT_MODE in the shared non-secret Worker var contract

## Context
Production was still able to deploy with placeholder values from wrangler.toml, including KORA_RPC_URL and provider RPC URL templates. That makes Kora-backed write actions fail even though the workflow fetched Doppler and synced secrets.

## Verification
- pnpm exec biome check scripts/render-wrangler-config.mjs
- pnpm check:env-contract
- pnpm --filter @sdp/api typecheck
- render script fails when production placeholders remain
- render script succeeds and writes Doppler-style production values into [env.production.vars]


GuiBibeau:
@greptileai review

PR #268, opened 2026-05-05
feat: use stable layout when transitioning

multipletwigs:
Use stable layouts for the animation so the layout doens't break

https://github.com/user-attachments/assets/91ce36a9-4323-4a39-b59e-9b327a085cac


multipletwigs:
@greptileai can you review again?

GuiBibeau (approved):
Thanks for fixing my slop 😅

PR #295, opened 2026-05-12
feat(sdp-web): HOO-490 expose tokenId and token selector in issuance API playground

eldarik:
## Summary
 - Fix the API playground shell's path-fields renderer to honor `kind: "select"`. The playground config already authored `tokenId` as a select with `Name (SYMBOL)` options and a
   sane default, but the path-fields branch in `api-playground-shell.tsx` always fell back to `<Input>` — pure asymmetry with the body-fields branch that handled select
  correctly. The fix is a ~15-line addition mirroring the body-fields renderer. Benefits all 17 endpoints with a `{tokenId}` or `{entryId}` path param.
  - Surface the token's internal id on the token detail header. New `Token ID:` row sits under the mint address with the full `tok_…` value and a copy button. Wired via new
  `tokenId` + `onCopyTokenId` props on `TokenManagementHeader`; `TokenManagementWorkspace` adds a `handleCopyTokenId` mirroring the existing address-copy helper and fires a
  `"Token ID copied"` Sonner toast. Since the create-token modal redirects to the detail page, the new token's id is visible and copyable immediately after creation.

  ## Testing
  - `pnpm typecheck` — clean across `sdp-web`, `sdp-api`, `sdp-docs`.
  - Two new Playwright e2e cases in `apps/sdp-web/playwright/tests/issuance.e2e.spec.ts`:
    - token-id row visible on the open-token detail page, copy button fires `"Token ID copied"` toast.
    - playground `{tokenId}` renders as a `<select>`, options include each seeded token id, default is one of the seeded ids, selecting `open.id` updates the control.
  - Manual: `/dashboard/issuance?tab=playground&endpoint=get-token` shows a populated `{tokenId}` dropdown with `Name (SYMBOL)` labels; switching selection updates the URL
  preview.
  - Manual regression: zero-tokens project — path field falls back to a text input (`buildSelectBackedField` already returns a plain field when `options.length === 0`).

  ## Out of scope
  - Moving the playground off the hand-maintained `issuance-playground-config.ts` onto OpenAPI is a separate, much larger ticket.
  - Sibling tickets from the same review (HOO-471 key selection error, HOO-472 mint API 500s, HOO-473 amount units) — addressed separately.
  - "Copy token ID" on the token list card — AC1 already satisfied by the detail-page row; can be added as a follow-up if requested.


multipletwigs (approved):
LGTM thanks for the select fix.

PR #301, opened 2026-05-13
chore(sdp-api): clean up openapi types for payments

multipletwigs:
## Summary

This PR cleans up the payments OpenAPI input schemas so they map back to the runtime Zod schemas in `routes/payments/schemas.ts` instead of redefining validation in the OpenAPI layer.

## Before

- Payment request bodies were manually rewritten in `openapi/schemas/payments.ts`, so docs could drift from runtime validation.
- Some path/query inputs were built inline in `openapi/paths/payments.ts` instead of using route schemas.
- Payments OpenAPI still exposed stale shapes for old or unused surfaces like payment wallet create/list/detail, confidential transfer requests, and fee quotes.
- Shared input values like priority fee, transfer direction, and transfer status were duplicated across runtime and OpenAPI code.

## After

- Active payment request bodies now import the runtime schemas as `*Base` and only attach OpenAPI metadata with `withOpenApi`.
- Wallet ID params, transfer ID params, and list-transfer query params now come from `routes/payments/schemas.ts`.
- Stale unused payments OpenAPI schemas and response helpers were removed.
- Shared payment input enums/options are defined once in route schemas and reused by OpenAPI.

## Intentional behavior change

- Zero transfer amounts are now rejected at request-schema validation time, matching the existing no-zero business rule but returning the standard `Invalid request body` validation shape with `errors.amount`.

## Validation

- Ran focused Biome checks for the touched payments OpenAPI/runtime schema files.
- Ran `pnpm -C apps/sdp-api exec vitest run src/routes/payments.test.ts`.
- Ran `pnpm --filter @sdp/api typecheck`.

### DIFF
--- a/apps/sdp-api/src/openapi/paths/payments.ts
+++ b/apps/sdp-api/src/openapi/paths/payments.ts
@@ -1,20 +1,15 @@
 import type { OpenAPIRegistry } from "@asteasolutions/zod-to-openapi";
-import { z } from "zod";
 
 import {
   createTransferRequestSchema,
   errorResponseSchema,
   executeOfframpRequestSchema,
   executeOnrampRequestSchema,
-  isoDateTimeSchema,
-  pageQuerySchema,
-  pageSizeQuerySchema,
+  paymentListTransfersQuerySchema,
+  paymentTransferIdParamsSchema,
+  paymentWalletIdParamsSchema,
   prepareTransferRequestSchema,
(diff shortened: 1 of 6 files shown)


multipletwigs:
@greptile-apps can we re-review here?

GuiBibeau (approved):
🔥 🚢

PR #304, opened 2026-05-14
fix(sdp-api): HOO-507 harden NodeBackgroundRunner for SIGTERM drain                                                       

G1de0n:
NodeBackgroundRunner is the Node-side replacement for ctx.waitUntil. It tracks in-flight background promises so the SIGTERM handler from [HOO-511](https://linear.app/solana-fndn/issue/HOO-511/phase-2-nodedocker-nodejs-entrypoint-serverts) can drain them before process.exit, which makes its edge-case behavior important for graceful shutdown in self-hosted Kubernetes/Docker deployments.

Three latent issues fixed before the drain gets wired up:

- Rejecting tracked promises were unhandled rejections until awaitAll's Promise.allSettled absorbed them, and were never absorbed at all if the process exited before drain. Now absorbed at registration via .catch(() => undefined).finally(...).
- Concurrent awaitAll() calls each flipped a boolean flag, so the first drain to finish cleared draining while the other was still in flight (any run() in that window was tracked but not awaited by the remaining drain's snapshot). Replaced with a drainCount refcount.
- run() during drain previously labeled the late task "dropped" and silently swallowed any rejection of that untracked promise. The warning now says "task not tracked (side effects already in flight)" and the rejection is logged.

Also exposes readonly draining: boolean on the BackgroundRunner interface for diagnostics. CF impl is false (drain is platform-managed).

Test coverage: 4 → 7 (drain-guard text, untracked-rejection logging, concurrent awaitAll refcount).

### DIFF
--- a/apps/sdp-api/src/index.ts
+++ b/apps/sdp-api/src/index.ts
@@ -35,6 +35,7 @@ import payments from "@/routes/payments";
 import projects from "@/routes/projects";
 import rpc from "@/routes/rpc";
 import webhooks from "@/routes/webhooks";
+import { WorkersBackgroundRunner } from "@/runtime/background-cf";
 import { getSentryOptions, isSentryEnabled } from "@/runtime/observability";
 import { cloudflareObservability, withSentry } from "@/runtime/observability-cf";
 import { trackPendingTransfers } from "@/services/jobs/track-pending-transfers";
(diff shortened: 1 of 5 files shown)


G1de0n:
@greptile-apps review

pashpashkin (dismissed):

multipletwigs (dismissed):
sick pr lets go

G1de0n:
@greptile-apps review

pashpashkin (approved):

PR #307, opened 2026-05-15
refactor(sdp-api): HOO-508 unify Sentry across runtimes via observability module

G1de0n:
Prepares src/index.ts for Phase 2 (the Node entrypoint, src/server.ts in [HOO-510](https://linear.app/solana-fndn/issue/HOO-510/phase-2-nodedocker-rediskvstore-implementation)) by routing every Sentry call-site through a runtime-neutral Observability interface. Without this split, the Node bundle can't import @sentry/cloudflare (workerd-only) and the CF bundle can't import @sentry/node (needs native modules).

- src/runtime/observability.ts: Observability/ObservabilityScope/MonitorOptions/SentryOptions types + getSentryOptions(env) (moved from index.ts)
- src/runtime/observability-cf.ts: cloudflareObservability + withSentry export (the CF-only top-level wrapper has no Node analog)
- src/runtime/observability-node.ts: nodeObservability + initNodeSentry helper (Node initialises at startup; wired up by HOO-510 when server.ts lands)
- src/index.ts: imports the observability module instead of @sentry/cloudflare; SENTRY_PENDING_TRANSFERS_MONITOR slug preserved verbatim; the scheduled handler's SENTRY_DSN gate now trims to match getSentryOptions (was: truthy-only)
- apps/sdp-api/package.json: +@sentry/node@^10.51.0

Tests: 17 new unit tests across the three modules. observability-cf.test.ts is smoke-only: under vitest's isolate:false config, a module-level vi.mock of @sentry/cloudflare leaks through src/index.ts and breaks the rest of the suite.

### DIFF
--- a/apps/sdp-api/package.json
+++ b/apps/sdp-api/package.json
@@ -33,6 +33,7 @@
     "@react-email/render": "2.0.8",
     "@sdp/types": "workspace:*",
     "@sentry/cloudflare": "^10.51.0",
+    "@sentry/node": "^10.51.0",
     "@solana-program/system": "catalog:",
     "@solana-program/token-2022": "catalog:",
     "@solana/addresses": "catalog:",
--- a/apps/sdp-api/src/index.ts
+++ b/apps/sdp-api/src/index.ts
@@ -5,7 +5,6 @@
  * Built with Hono, Postgres, Hyperdrive, and KV
  */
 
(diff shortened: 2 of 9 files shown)


G1de0n:
@greptile-apps review

multipletwigs (approved):
nicee

PR #312, opened 2026-05-18
refactor(sdp-api): HOO-509 split index.ts → app.ts + extract cron function

G1de0n:
Unblocks [HOO-511](https://linear.app/solana-fndn/issue/HOO-511/phase-2-nodedocker-nodejs-entrypoint-serverts) : Node's `serve({ fetch })` and the node-cron runner both need the Hono app and reconciliation logic, but the old index.ts welded them to ExecutionContext / waitUntil / @sentry/cloudflare. Without this split server.ts would have to duplicate every middleware mount, route registration, and error handler — then keep both copies in sync by hand.

The KV factory still returns the Workers impl unconditionally; that dispatch is [HOO-510](https://linear.app/solana-fndn/issue/HOO-510/phase-2-nodedocker-rediskvstore-implementation) scope (`runtime/factory.ts:4-5` documents the intentional staging) and is gated behind `createKVStoreSet`, so app.ts works as the seam without it.

- `apps/sdp-api/src/app.ts` (new): `createApp({ observability })` factory. Owns the Hono instance, middleware stack (request-id → tracing → secure headers → CORS → prettyJSON → logger → kvStoreMiddleware → skipRateLimitPaths, KV_FREE_PATHS shared between the last two), routes, onError, notFound, plus helpers (mapSigningError, getFireblocksBlockedError, captureUnexpectedError). captureUnexpectedError now takes `observability` as a parameter instead of importing the CF Sentry instance directly. AC: no cloudflare / wrangler / ExecutionContext / KVNamespace text in this file.
- `apps/sdp-api/src/cron/pending-transfers.ts` (new): `runPendingTransfersReconciliation({ env, bg, observability? })` — literal extraction of the old `scheduled` body. Exports PENDING_TRANSFERS_MONITOR / PENDING_TRANSFERS_CRON. When observability is supplied, wraps trackPendingTransfers in a Sentry cron monitor; otherwise hands the bare promise to bg.run.
- `apps/sdp-api/src/index.ts`: shrinks from 369 → 56 lines. Now a thin CF entrypoint: instantiates `createApp({ observability: cloudflareObservability })`, builds the worker { fetch, scheduled, request } object, wraps it with `withSentry`. `scheduled` delegates to `runPendingTransfersReconciliation`, passing `cloudflareObservability` only when `isSentryEnabled(runtimeEnv)`. 17 test files keep importing `app from "@/index"` via the `request` method on the default export — shape preserved.


G1de0n:
@greptile-apps please review

eldarik (approved):
LGTM

PR #315, opened 2026-05-19
feat(pro-1202): add test mode indicator and toggle

multipletwigs:
## Summary

Adds a UI surface for the upcoming sandbox / production split and wires the current backend behavior to stay sandbox-safe from the dashboard. The dashboard now shows a `Sandbox` badge next to the avatar, and the sidebar has a new **Mode** section that groups a disabled `Sandbox mode` toggle alongside the existing API Debug Logs toggle.

Sandbox mode is **always on and disabled** for now, you cannot toggle it to Production yet. Requests made through `sdp-web` default to sandbox behavior, while direct API requests continue to resolve environment from the API key (`sk_test` -> sandbox, `sk_live` -> production). 

CleanShot 2026-05-19 at 16 57 16@2x

## What changed

- **`Sandbox` badge in the dashboard top bar** (`dashboard-shell.tsx`)
  - Sits to the right of the `UserButton` with a thin vertical divider.
  - Renders in both the default and centered-title top bar variants.
  - Bound to dashboard workspace environment state, so it can disappear automatically once production mode is supported.

- **New `Mode` section in the sidebar** (`dashboard-shell.tsx`)
  - Single uppercase `MODE` header (matching `CREATE` / `MANAGE` styling) with two toggle rows underneath.
  - `Sandbox mode` row: a disabled toggle locked in the on position, with a tooltip ("Only Sandbox mode supported at the moment") anchored to the right of the label.
  - `API Debug Logs` row: the existing network-debug toggle, restyled to a single nav-row to match.

- **Dashboard workspace environment state** (`dashboard-workspace-context.tsx`)
  - Replaces the boolean `isSandbox` with `sdpEnvironment`, currently initialized to `sandbox`.
  - Scopes dashboard SWR/persisted cache keys by environment so future production support will not mix sandbox and production data.

- **Sandbox backend behavior for dashboard calls** (`ramps.ts`)
  - API-key requests use the API key environment.
  - Non-API-key/dashboard requests default to sandbox while that is the only supported dashboard mode.

- **Network debug toggle restyled** (`network-debug-panel.tsx`)
  - Dropped the bordered container, "Development mode" sub-label, and info tooltip. Now matches the same nav-row visual as the Sandbox toggle: switch + label + optional pending-count pill on the right.
  - Removed the unused `NetworkDebugSwitch`, `InfoIcon`, and `Tooltip*` imports.

- **Shared `Badge` UI component** (`components/ui/badge.tsx`)
  - Thin wrapper over `@solana/design-system/badge` with semantic variants (`default`, `success`, `warning`, `danger`, `info`). The Sandbox badge uses the neutral `default` variant.
  

## Project attachment

The current status quo is that there is only ever one effective dashboard project: the first active project available to the user, falling back to an auto-provisioned "Default Project" when none exists. That default project is created with environment = 'sandbox', and dashboard/session requests do not currently select or switch projects based on the sandbox/production UI state.

## Notes

- The dashboard does not expose production mode yet; `sdp-web` remains sandbox-only for this PR.
- Direct API users still select sandbox or production through their API key prefix/environment.
- Project-based environment switching is intentionally left for a follow-up once the product direction is settled.

## Test plan

- [ ] Sandbox badge appears next to the avatar on every dashboard route (incl. centered-title pages like `/dashboard/payments/send`).
- [ ] Sidebar shows the `MODE` section with both toggles; Sandbox toggle is visually on, not clickable, and shows tooltip on hover of the label.
- [ ] API Debug Logs toggle still enables/disables the debug panel and pending-count badge as before.
- [ ] Dashboard ramp/provider requests without an API key use sandbox provider credentials.
- [ ] Direct API requests continue to resolve sandbox/production from the API key environment.
- [ ] `pnpm --filter sdp-web typecheck` and targeted Biome checks pass on the changed files.


multipletwigs:
@greptile-apps please do another review! thanks

pashpashkin (approved):

PR #317, opened 2026-05-19
fix(sdp-api): HOO-461 Auto-add mint destinations to on-chain allowlist

eldarik:
## Problem

Per Ilan's review:

> j.i — this doesn't actually use token acl and instead is a db based allowlist.
nothing actually blocking the token holder from having access to the token.
>
> j.ii — default account state is set to initialized. allowlist tokens should have
this set to frozen

These are two surface symptoms of the same root cause. Verified on devnet pre-fix:
allowlist tokens deployed with `defaultAccountState=Initialized`, no on-chain ABL
enforcement. The allowlist was a DB-only assertion
(`assertDestinationAllowedByControlList`); a third party could receive the token
without ever being on the list.

**Trace**:

1. `apps/sdp-api/src/routes/issuance/handlers/access-control.ts` —
`shouldEnableOnChainAcl` returned `true` for allowlist tokens **only when** `network
=== "mainnet-beta"`. Every other network → on-chain ABL off.
2. `apps/sdp-api/src/services/mosaic/service.ts` — `enableSrfc37 = requestedSrfc37 &&
freezeAuthority !== null`. With `requestedSrfc37=false` on non-mainnet, sRFC-37 stayed
 off.
3. Mosaic SDK templates (`stablecoin`, `arcade`, `tokenized-security`) call
`withDefaultAccountState(aclMode === 'blocklist' || !useSrfc37)` — with
`useSrfc37=false`, allowlist tokens resolved to `Initialized`.

The mainnet carve-out existed because a fresh destination wouldn't be on the on-chain
allowlist yet, and minting to a frozen ATA fails with `0x11 AccountFrozen`. Fixing the
 default state required also fixing the mint flow.

## Solution

Two changes, shipped together. Together they close both HOO-461 and HOO-469.

### 1. Drop the mainnet-only carve-out in `shouldEnableOnChainAcl`

```diff
-  if (mode === "blocklist") return true;
-  return network === "mainnet-beta";
+  return getTokenAccessControlMode(token) !== "disabled";
```

Once sRFC-37 fires on every network for allowlist tokens, the Mosaic SDK templates
emit defaultAccountState=Frozen automatically. HOO-461 closes by side effect.

2. Auto-add mint destinations to the on-chain allowlist

New helper getOnChainAllowlistMutationForMint(token, network) returns the ABL list
address when:
- token is allowlist mode, and
- on-chain ABL is active for the token+network, and
- ablListAddress is populated.

prepareMint and executeMint use it as follows:

- Helper returns null (blocklist, or allowlist without ablListAddress — e.g. legacy
tokens deployed pre-PR) → preserve the existing strict assertion via
assertDestinationAllowedByControlList.
- Helper returns a list address → skip the strict assertion, run a DB-first /
on-chain-second sync:
  a. tokenService.addAllowlistEntry(...) (DB mirror).
  b. mosaic.addToList(...) (on-chain).
  c. If on-chain throws → revokeAllowlistEntry rollback → re-throw.
  d. Then proceed to prepareMintTo / mintTo.

Idempotent: ADDRESS_ALREADY_ALLOWLISTED is swallowed as a benign race. Audit log
records addedToAllowlist: boolean.

The DB-first ordering mirrors the existing /allowlist POST handler — guarantees the
two layers can't drift if the DB insert fails after the on-chain write succeeded. (Per
 Copilot review feedback on the initial commit.)

This unblocks the SDK's permissionless-thaw step: it checks the on-chain ABL before
unfreezing the destination ATA. With the wallet now on-list, the bundled mint path
succeeds. HOO-469's structural complaint closes — the allowlist actually gates
on-chain receipt now.

Why two txs instead of bundling

Separate addToList then mintTo rather than bundling:

- Same correctness outcome on the user-visible path.
- No @solana/mosaic-sdk signature changes needed (the wrapper's payer plumbing isn't
exposed via the .d.ts).
- If addToList succeeds and mintTo fails, the wallet stays allowlisted — correct retry
 state, not a corrupt one.

Single-tx bundling can land as a follow-up if Kora cost / latency forces it.

Mosaic SDK patch fixes

patches/@solana__mosaic-sdk@0.1.1.patch extended to fix two upstream bugs surfaced
once sRFC-37 fires on devnet:

- walletAddress → wallet key drift in getAddWalletInstructions /
getRemoveWalletInstructions. The upstream @solana/mosaic-sdk@0.1.1 calls
findWalletEntryPda({ walletAddress, listConfig }), but the underlying
@solana/token-acl-gate-sdk@0.2.0 reads seeds.wallet. Result: addToList threw Cannot
read properties of undefined (reading 'length') for every allowlist mutation.
- Hunk header line count fix so the patch actually applies during install. The pre-fix
 hunk declared @@ -58,7 +59,7 @@ but had 8 lines per side; pnpm rejected the patch
with hunk header integrity check failed.

Plus the existing payer plumbing additions and inspect-token.js freeze-authority
detection (kept).

Diagnostics

app.ts catch-all unexpected-error logger now includes context and cause for
SolanaError instances. Surfaced the upstream findWalletEntryPda bug above; useful for

patches/@solana__mosaic-sdk@0.1.1.patch extended to fix two upstream bugs surfaced once sRFC-37 fires on devnet:

- walletAddress → wallet key drift in getAddWalletInstructions / getRemoveWalletInstructions. The upstream @solana/mosaic-sdk@0.1.1 calls findWalletEntryPda({ walletAddress,
listConfig }), but the underlying @solana/token-acl-gate-sdk@0.2.0 reads seeds.wallet. Result: addToList threw Cannot read properties of undefined (reading 'length') for every
allowlist mutation.
- Hunk header line count fix so the patch actually applies during install. The pre-fix hunk declared @@ -58,7 +59,7 @@ but had 8 lines per side; pnpm rejected the patch with
hunk header integrity check failed.

Plus the existing payer plumbing additions and inspect-token.js freeze-authority detection (kept).

Diagnostics

app.ts catch-all unexpected-error logger now includes context and cause for SolanaError instances. Surfaced the upstream findWalletEntryPda bug above; useful for future devnet
simulation failures.

Dependency overrides

- `fast-uri` pinned to `^3.1.2` in the root `pnpm.overrides`. The Dependency Review job flagged two high-severity advisories on the transitively-pulled `fast-uri@3.1.0`: `GHSA-q3j6-qgpj-74h6` (path traversal via percent-encoded dot segments) and `GHSA-v39h-62p7-jpjc` (host confusion via percent-encoded authority delimiters). The override is the upstream-recommended fix; no code changes needed on our side.

Out of scope

- Transfer-flow auto-allowlisting (Alice → fresh Bob). Different product question — needs a stakeholder call. Tracked separately.
- Seize / force-transfer (handlers/seize.ts:53–60, :169–176) still uses strict DB assertion and never calls addToList. Distinct stakeholder review around force-transfer policy.
- Burn / force-burn — no ACL check at all; unrelated.
- Backfill of pre-PR allowlist tokens with ablListAddress=NULL. Mint falls back to DB-only assertion for those — preserves the old behavior for old tokens, not a new
regression. Ops task; devnet-era tokens can be re-deployed.
- Upstream Mosaic SDK PRs for blocklist default state and freeze extra-metas. Independent concerns, tracked separately.

Testing

Automated

- pnpm --filter @sdp/api typecheck — clean on touched files. Pre-existing unrelated errors on main untouched.
- vitest run src/routes/issuance.test.ts — 73 pass, 9 skipped. Includes 4 new handler-level integration tests for the auto-add mint behavior:
  - auto-adds destination to on-chain allowlist on prepare mint
  - auto-adds destination to on-chain allowlist on execute mint
  - skips on-chain allowlist sync on prepare mint when destination already on list
  - skips on-chain allowlist sync on execute mint when destination already on list
- access-control.test.ts — unit coverage for shouldEnableOnChainAcl across 3 templates × 4 networks + getOnChainAllowlistMutationForMint shapes.
- pnpm biome check on touched files — clean.

Manual end-to-end on devnet

Verified the structural fix on devnet token 243EfRs6GgotaWsWvFbFoRuSuH1GKvP2FC8Ncfd4hpm7:

1. Deploy fresh allowlist stablecoin → on-chain defaultAccountState=Frozen ✓, freezeAuthority is Token ACL mintConfig PDA (owned by TACLkU6CiCdkQN2MjoyDkVg2yAH9zkxiHDsiztQ52TP)
 ✓, ablListAddress populated ✓.
2. Add wallet via /allowlist → on-chain walletEntry PDA created (owned by ABL program GATEzzqxhJnsWF6vHRsgtixxSB8PaQdcqGEVTEHWiULz), listConfig.walletsCount increments ✓.
3. Mint to fresh wallet → addedToAllowlist: true, ATA thawed, tokens land.
4. Mint to same wallet again → addedToAllowlist: false, no redundant addToList.
5. Negative check: transfer from allowlisted wallet → non-allowlisted wallet outside SDP fails with AccountFrozen. (This is the HOO-469 regression guard.)


eldarik:
@greptile-apps review

eldarik:
@greptile-apps review

eldarik:
@greptile-apps review

eldarik:
@greptile-apps review

eldarik:
@greptile-apps review

eldarik:
@greptile-apps review

eldarik:
@greptile-apps review

eldarik:
@greptile-apps review

GuiBibeau (approved):

PR #318, opened 2026-05-19
feat(sdp-api): HOO-510 RedisKVStore for Node runtime

G1de0n:
## Summary
- New `RedisKVStore` + `createRedisKVStoreSet` for the Node runtime: single shared `ioredis` client, four prefixed stores (`apiKeys` / `rateLimits` / `cache` / `sessions`).
- Factory now branches on `getRuntime(env)`: `node` → Redis, `cloudflare` → existing Workers KV (unchanged).
- Adds `ioredis ^5.4.1`, a sibling `vitest.node.config.ts` (plain Node pool for `*.node.test.ts`), and a `redis:7-alpine` service container in the Unit Tests CI job.

## Why a separate Node-pool vitest config
The existing config uses `@cloudflare/vitest-pool-workers`, which runs tests inside a simulated Workers runtime. `ioredis` opens raw TCP sockets via `node:net` and won't run there. Rather than fight either runtime, this PR adds a tiny sibling config that picks up `*.node.test.ts` files and runs them in plain Node. The Workers config now `exclude`s the same pattern so neither config crosses over.

## Spec mapping (per Linear ticket)
- TTL via `SET key value PX <ttl_seconds * 1000>` — ticket says PX explicitly; multiplies seconds because the interface inherits CF's `expirationTtl: seconds`.
- Key listing via `SCAN MATCH <prefix>:* COUNT 100` until cursor=0 (non-blocking, paginated).
- No-TTL `put` uses plain `SET` so `rpc:relay:stats:*` and the round-robin cursor preserve their current CF KV semantics (persist indefinitely). Any change to that would be a behavior change in callers, not in HOO-510.

## Watch-out for cutover
Redis doesn't serve stale reads after TTL expiry. Cloudflare KV has up to ~60s grace; the Node path won't. Anything that accidentally relies on the grace window will surface.

## Drive-by fix
`env.d.ts` was missing `LIGHTSPARK_GRID_API_BASE_URL` and `BVNK_API_TOKEN` even though `PROCESS_ENV_FALLBACK_KEYS` in `runtime-env.ts` references them. CI's typecheck job filters by changed packages, so the breakage only surfaces on PRs that touch `sdp-api` — like this one. Adding the two optional string declarations is the minimum needed for CI to go green.

## Test plan
- [x] 13 integration tests against real Redis pass locally (round-trip, JSON overload, `SET PX` PTTL semantics, no-TTL persists, SCAN-based list with prefix strip, cross-store isolation, factory error paths for missing/whitespace `REDIS_URL`).
- [x] `pnpm --filter @sdp/api typecheck` clean.
- [x] `biome check` clean on PR-scoped files (matches CI's diff-only lint).
- [x] CI run: Unit Tests job picks up the new `redis:7-alpine` service and "Node-runtime tests" step.
- [ ] Follow-up (HOO-511): graceful `client.quit()` on SIGTERM lives in the Node entrypoint, not here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       DATABASE_URL: postgresql://sdp:sdp@127.0.0.1:5432/sdp
+      REDIS_URL: redis://127.0.0.1:6379
     services:
       postgres:
         image: postgres:16-alpine
@@ -89,6 +90,18 @@ jobs:
           - 5432:5432
         options: >-
           --health-cmd "pg_isready -U sdp -d sdp"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 20
+      # HOO-510: Redis backs the node-runtime KVStore. Node-pool tests
+      # (*.node.test.ts) connect via REDIS_URL above. Workers-pool tests
+      # don't touch it.
+      redis:
+        image: redis:7-alpine
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "redis-cli ping"
           --health-interval 5s
           --health-timeout 5s
           --health-retries 20
@@ -129,6 +142,12 @@ jobs:
           DOPPLER_TOKEN: ${{ secrets.DOPPLER_TOKEN_CI }}
           DOPPLER_CONFIG: dev_ci
         run: pnpm test
+
+      - name: Node-runtime tests (Redis-backed KVStore)
+        env:
+          DOPPLER_TOKEN: ${{ secrets.DOPPLER_TOKEN_CI }}
+          DOPPLER_CONFIG: dev_ci
+        run: pnpm --filter @sdp/api test:node
 
   build_api:
     name: Build API
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,9 +93,6 @@ jobs:
           --health-interval 5s
           --health-timeout 5s
           --health-retries 20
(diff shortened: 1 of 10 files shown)


G1de0n:
Both findings addressed across `b1aa7ae` + `a2ad016`:

- **P1 (connection leak).** `createRedisKVStoreSet` shares one cached `Promise<Redis>` per URL via the module-level `clientPromisesByUrl` map; `kvStoreMiddleware`'s per-request factory calls all land on the same promise. ioredis itself is lazy-loaded inside `ensureClient` via `await import("ioredis")` so the static graph from `factory.ts` doesn't include it — that also unblocked the CF Workers test pool (miniflare's transformer otherwise choked on ioredis's `debug.js`). Rejected promises self-evict (with an `===` guard) so a transient malformed-URL or module-load failure doesn't permanently poison the cache. Regression test pins reference equality on `clientPromise` across two factory calls. `closeAllRedisClients()` exported for HOO-511's SIGTERM handler.
- **P2 (misleading `lazyConnect` comment).** Rewrote — ioredis starts the TCP handshake immediately but **asynchronously**; an unreachable host doesn't throw from `new Redis(...)`. The comment now reflects that.
- **`REDIS_URL` not in `PROCESS_ENV_FALLBACK_KEYS`.** Added it (plus `DATABASE_URL` for symmetry, since `env.d.ts` groups both as "Node runtime equivalents"). Noted in the commit body that `DATABASE_URL` is technically redundant today — `db/client.ts:316` already falls back to `process.env.DATABASE_URL` directly — but keeping it consistent prevents future consumers from having to remember the exception.

G1de0n:
@greptile-apps review please

pashpashkin (approved):

PR #327, opened 2026-05-20
feat(sdp-api): HOO-511 Node.js entrypoint (server.ts)

G1de0n:
## Summary

- Adds `src/server.ts` — a Node.js process entrypoint that wires the runtime-neutral seams from earlier commits (process-env Env fallback, KVStore + Redis factory, NodeBackgroundRunner with SIGTERM drain, nodeObservability, createApp / runPendingTransfersReconciliation) into a Hono server bound to `@hono/node-server`, with a node-cron reconciliation tick and a graceful SIGTERM/SIGINT drain. Cloudflare Workers entrypoint (`src/index.ts`) untouched.
- Adds `src/runtime/shutdown-node.ts` — pure `shutdown(deps)` helper. Order: `server.close → cron.stop → bg.awaitAll → closeAllRedisClients → closeDatabasePools`. Lives next to `background-node` / `observability-node` so unit tests don't pull the full Hono app graph through `@solana/mosaic-sdk`, which has a directory-import resolution that trips the plain Node test pool.
- Adds `src/cron/runner.ts` — `startCron` returns `null` when `DISABLE_CRON ∈ {"true","1"}` (single-replica strategy until a distributed lock lands), otherwise schedules `PENDING_TRANSFERS_CRON` through node-cron and routes the tick through the shared `runPendingTransfersReconciliation`. A `stopping` closure flag short-circuits any tick that fires concurrent with shutdown so no late work gets registered after `awaitAll()`'s snapshot.

### Behaviour notes

- `main()` hardcodes `SDP_RUNTIME=node` after `withProcessEnvFallback({} as Env)` — the entrypoint IS the runtime; don't let a stray env value flip the KV factory back to Cloudflare mode.
- `assertRequiredEnv` fails loud at boot if `ENVIRONMENT`, `API_VERSION`, `DATABASE_URL`, or `REDIS_URL` is unset.
- `PORT` defaults to `8787` for parity with `wrangler dev`; non-numeric values (`"8787abc"`) reject via `Number()`-not-`parseInt`.
- SIGTERM and SIGINT route through a cached promise so a second signal is a no-op.
- `unhandledRejection` takes the graceful path; `uncaughtException` exits fast since V8 documents the process state as potentially corrupt.
- `import.meta.url === pathToFileURL(process.argv[1]).href` guard keeps `main()` from running on import.

## Test plan

- [x] `pnpm --filter @sdp/api typecheck` — green
- [x] `pnpm exec biome check` on changed files — green
- [x] `pnpm --filter @sdp/api test` (workers pool) — 440 passed, 14 skipped, 0 failed
- [x] `pnpm --filter @sdp/api test:node` on the new files — 12/12 passed
- [x] `pnpm --filter @sdp/api build` (`wrangler deploy --dry-run`) — green; CF entrypoint still builds
- [x] Smoke: `tsx src/server.ts` boots, `GET /health` → 200, `GET /` → 302, SIGTERM produces ordered drain log
- [x] Smoke: missing `DATABASE_URL` / `REDIS_URL` exits with `Fatal startup error: …`; `PORT=8787abc` exits with `Invalid PORT: 8787abc`
- [ ] CI green on this PR

### DIFF
--- a/apps/sdp-api/src/cron/runner.ts
+++ b/apps/sdp-api/src/cron/runner.ts
@@ -28,10 +28,25 @@ export interface CronHandle {
 }
 
 const TRUTHY_DISABLE_CRON: ReadonlySet<string> = new Set(["true", "1"]);
+const FALSY_DISABLE_CRON: ReadonlySet<string> = new Set(["false", "0"]);
 
+// Strict whitelist so a typo (`DISABLE_CRON=treu`) fails loudly instead of
+// silently enabling cron and double-firing across replicas.
 function isCronDisabled(env: Env): boolean {
-  const raw = env.DISABLE_CRON?.trim().toLowerCase();
-  return raw !== undefined && TRUTHY_DISABLE_CRON.has(raw);
+  const raw = env.DISABLE_CRON;
+  if (raw === undefined) {
+    return false;
+  }
+  const normalised = raw.trim().toLowerCase();
+  if (TRUTHY_DISABLE_CRON.has(normalised)) {
+    return true;
+  }
+  if (FALSY_DISABLE_CRON.has(normalised)) {
+    return false;
+  }
+  throw new Error(
+    `Invalid DISABLE_CRON: ${JSON.stringify(raw)} (expected 'true', 'false', '1', or '0')`
+  );
 }
 
 export function startCron(deps: CronDeps): CronHandle | null {
--- a/apps/sdp-api/src/cron/runner.ts
+++ b/apps/sdp-api/src/cron/runner.ts
@@ -1,0 +1,65 @@
+/**
+ * node-cron wrapper for the Node runtime. Schedules the same reconciliation
+ * job that the Cloudflare `scheduled` handler triggers, going through the
+ * shared `runPendingTransfersReconciliation` so observability and background
+ * tracking are wired identically across runtimes.
+ *
+ * `DISABLE_CRON=true` skips registration entirely, leaving the process free
+ * to run as one of many web replicas without firing the reconciliation N
+ * times per minute. A distributed lock would let every replica schedule
+ * safely; until that lands, single-replica + DISABLE_CRON elsewhere is the
+ * agreed-upon strategy.
+ */
+
+import { type ScheduledTask, schedule } from "node-cron";
+import type { BackgroundRunner } from "@/runtime/background";
+import type { Observability } from "@/runtime/observability";
+import type { Env } from "@/types/env";
+import { PENDING_TRANSFERS_CRON, runPendingTransfersReconciliation } from "./pending-transfers";
+
+export interface CronDeps {
+  env: Env;
+  bg: BackgroundRunner;
+  observability?: Observability;
+}
+
+export interface CronHandle {
+  stop(): void | Promise<void>;
+}
+
+const TRUTHY_DISABLE_CRON: ReadonlySet<string> = new Set(["true", "1"]);
+
+function isCronDisabled(env: Env): boolean {
+  const raw = env.DISABLE_CRON?.trim().toLowerCase();
+  return raw !== undefined && TRUTHY_DISABLE_CRON.has(raw);
+}
+
+export function startCron(deps: CronDeps): CronHandle | null {
+  if (isCronDisabled(deps.env)) {
+    return null;
+  }
+
+  // node-cron's `task.stop()` halts future scheduling but doesn't promise
+  // to interrupt a tick already mid-flight. A `stopping` flag short-circuits
+  // any callback that fires concurrent with shutdown so no extra background
+  // work gets registered after the awaitAll() snapshot is taken.
+  let stopping = false;
+
+  const task: ScheduledTask = schedule(PENDING_TRANSFERS_CRON, () => {
+    if (stopping) {
+      return;
+    }
+    runPendingTransfersReconciliation({
+      env: deps.env,
+      bg: deps.bg,
+      observability: deps.observability,
+    });
+  });
+
+  return {
+    stop() {
+      stopping = true;
+      return task.stop();
+    },
+  };
+}
--- a/apps/sdp-api/src/runtime/shutdown-node.ts
+++ b/apps/sdp-api/src/runtime/shutdown-node.ts
@@ -1,0 +1,60 @@
+/**
+ * Node SIGTERM/SIGINT shutdown sequence.
+ *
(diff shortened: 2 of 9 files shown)


G1de0n:
@greptile-apps review please

apps/sdp-api/src/server.ts
```diff
@@ -0,0 +1,193 @@
...
+import { withProcessEnvFallback } from "@/lib/runtime-env";
+import { NodeBackgroundRunner } from "@/runtime/background-node";
+import { getSentryOptions, isSentryEnabled } from "@/runtime/observability";
+import { initNodeSentry, nodeObservability } from "@/runtime/observability-node";
+import { shutdown } from "@/runtime/shutdown-node";
+import type { Env } from "@/types/env";
+
+const DEFAULT_PORT = 8787;
+const DEFAULT_SHUTDOWN_TIMEOUT_MS = 25_000;
+const TRUTHY_ENV: ReadonlySet<string> = new Set(["true", "1"]);
```
multipletwigs, line 26: ditto on concern for cron env

apps/sdp-api/src/cron/runner.ts
```diff
@@ -0,0 +1,65 @@
...
+  env: Env;
+  bg: BackgroundRunner;
+  observability?: Observability;
+}
+
+export interface CronHandle {
+  stop(): void | Promise<void>;
+}
+
+const TRUTHY_DISABLE_CRON: ReadonlySet<string> = new Set(["true", "1"]);
```
multipletwigs, line 30: Should we avoid truthy env vars? A narrower contract and just maybe accept "true" should be better in terms of maintainability. Clearest contract is `DISABLE_CRON=true` or `DISABLE_CRON=false`

apps/sdp-api/src/cron/runner.ts
```diff
@@ -0,0 +1,65 @@
...
+
+function isCronDisabled(env: Env): boolean {
```
multipletwigs, line 35:
```
  DISABLE_CRON="treu"  // raw "treu", has => false, cron enabled
  DISABLE_CRON="yes"   // raw "yes", has => false, cron enabled
```
I think because startCron only branches on `isCronDisabled()`, any `DISABLE_CRON` value other than "true" or "1" falls through and starts cron. That means typos like DISABLE_CRON=treu, also triggers it. Maybe stricter contract here is good.

apps/sdp-api/src/runtime/shutdown-node.ts
```diff
@@ -0,0 +1,71 @@
...
+export interface ShutdownDeps {
+  server: Closable;
+  cron: CronHandle | null;
+  bg: BackgroundRunner;
+  log: (msg: string) => void;
+}
+
+export async function shutdown(deps: ShutdownDeps): Promise<void> {
+  deps.log("closing HTTP listener");
+  await new Promise<void>((resolve, reject) => {
```
multipletwigs, line 48: Thinking about the order of shutdown here, is it better to shutdown the cron first before draining the http listener? Stopping cron first would shut down the internal producer of background work before we wait on HTTP drain, then `bg.awaitAll()` can drain the bounded set of work that was already registered. 

Avoids unexpected new work at the moment of shutdown.

apps/sdp-api/src/server.ts
```diff
@@ -0,0 +1,193 @@
...
+  }
+}
+
+async function main(): Promise<void> {
+  // The Node entrypoint IS the Node runtime — don't let a stray process.env
+  // value flip the KV factory back to Cloudflare mode. Set this before
+  // assertRequiredEnv so downstream code (e.g. getRuntime) sees the override.
+  const env = withProcessEnvFallback({} as Env);
+  env.SDP_RUNTIME = "node";
+  assertRequiredEnv(env);
```
multipletwigs, line 109: we assert the requirement it exists but perhaps we should also validate? Any typo like prod, prd, staging, or Production makes the API behave as non-production and allow all origins.

multipletwigs (commented):
just a few questions

apps/sdp-api/src/server.ts
G1de0n: Good catch — this had real CORS implications. Fixed in 7763cc6: `assertRequiredEnv` now rejects any `ENVIRONMENT` outside `{development, production}` with an explicit error, so typos like `prod` or `Production` fail at startup instead of opening dev-only paths.

apps/sdp-api/src/cron/runner.ts
G1de0n: Agreed. Fixed in 7763cc6: `DISABLE_CRON` now has an explicit whitelist (`true`/`false`/`1`/`0`, case-insensitive after trim); anything else throws at startup. Tests cover the typo cases (`treu`, `yes`) plus blank-string and whitespace/casing normalisation.

apps/sdp-api/src/cron/runner.ts
G1de0n: Done in 7763cc6 — `DISABLE_CRON` now accepts only `true`/`false`/`1`/`0` and throws on everything else.

apps/sdp-api/src/server.ts
G1de0n: The only consumer of `TRUTHY_ENV` in `server.ts` is `shouldShutdownOnUnhandledRejection`, which already throws on any unrecognised value (including blank). So the `server.ts` env parsing here is already strict; The matching `DISABLE_CRON` fix landed in 7763cc6.

apps/sdp-api/src/runtime/shutdown-node.ts
G1de0n: Good question — I went back and forth on this one too. The thing that landed me on the current order is that `bg.run` actually has two producers, not just cron: in-flight HTTP requests also fire-and-forget background work. So stopping cron first wouldn't really close the registration window — HTTP requests would still be opening it. Closing HTTP first drains those, and the `stopping` flag in `runner.ts` handles a cron tick that races `stop()`.

The file header in `shutdown-node.ts` has the full reasoning if you want to dig in. Let me know if I'm missing the scenario you had in mind though — happy to revisit.

multipletwigs: Oh nice! Yeah, it makes sense, good rationales

multipletwigs (approved):
very nice! love this pr, thanks a lot!

PR #360, opened 2026-05-27
feat: list all provider wallets by default

GuiBibeau:
## Summary

- List wallets across all active providers by default from `/v1/wallets`, while preserving `includeAllProviders=false` for default-provider-only results.
- Update the custody OpenAPI/Postman copy and multi-provider route coverage for the new default.
- Add wallet ID copy support and full-value tooltips for wallet address and wallet ID in the wallet overview cards.

## Validation

- `pnpm --filter @sdp/api typecheck`
- `pnpm --filter sdp-web typecheck`
- `pnpm exec vitest run src/routes/custody-multi-provider.test.ts --config vitest.workers.config.ts`

Note: `pnpm --filter @sdp/api test -- src/routes/custody-multi-provider.test.ts` currently runs the full Workers suite before the targeted Node test and fails locally in `counterparties.test.ts` because the local test DB is missing the `status` column. The directly targeted custody test passes.


multipletwigs (approved):
lets go!

PR #372, opened 2026-05-29
feat: confidential transfers extension + create-token flow page

GuiBibeau:
## Summary

Adds Token-2022 **confidential transfers** support to token issuance and reworks the create-token wizard into a dedicated page. Also fixes a height/alignment issue in the payments send form.

Screenshot 2026-05-29 at 4 02 17 PM

## Issuance
- Allow the `confidentialTransfers` extension override on the **stablecoin**, **tokenized-security**, and **custom** templates (devnet-only), wired through `createTokenSchema`, the create handler, and `MosaicService`.
  - For stablecoin/tokenized-security the mosaic SDK already enables confidential balances unconditionally; this just stops the API from rejecting the override as "Invalid template overrides".
- Move the create-token wizard from a modal into a dedicated `/dashboard/issuance/create` page (template selection → identity → features), with a matching Suspense loading skeleton.
- Flatten the step layouts to match the rest of the dashboard forms (no card wrapper), neutralize the template icon pills, and add the confidential-transfers toggle (checkbox style).
- Submit is now driven explicitly from the button, so landing on the features step (e.g. via Enter on a prior step) no longer auto-submits.

## Payments
- Fix the Amount/Asset (and Reference/Asset) row so the input and select share the same height (`size="lg"`) and vertically align, by using the design-system field `label` consistently across both columns.

## Test plan
- [ ] Create a stablecoin draft on devnet with confidential transfers enabled — no longer 400s.
- [ ] Create-token wizard steps render flat/consistent; features step doesn't auto-submit on load.
- [ ] Payments send form: Amount and Asset fields are the same height and aligned.


multipletwigs:
sorry @GuiBibeau :p one playground bug

eldarik (approved):
LGTM

PR #375, opened 2026-05-29
fix: HOO-585 enforce project boundary in token reads and project scope

eldarik:
## Summary
- Scopes `TokenService.getToken` to `organization_id + project_id` at the query level — renamed to `getToken({ tokenId, organizationId, projectId })`, returning `null` for tokens outside the caller's scope. This closes a cross-project read gap where a project-scoped caller could fetch tokens (and their allowlist/authority/freeze/supply/transaction sub-resources) belonging to another project in the same org.
- Adds a private `_getTokenById` for service-internal callers that already trust the id (update/supply flows), so internal lookups are unchanged.
- Updates every token-resolving issuance handler (allowlist, authority, burn, deploy, force-burn, freeze, mint, pause, seize, supply, tokens, transactions) to resolve `{ projectId, orgId }` via `requireProjectScope(c)` and pass them into the scoped lookup, dropping the manual `getAuth` + post-fetch org/project comparison. Out-of-scope tokens now 404. `requireProjectScope` is now synchronous.
- Drops the `?projectId=` query-param fallback in `projectContextMiddleware` — project scope must come from the `x-project-id` header (API-key callers stay pinned to their key's project).
- Part of HOO-585: hardening the production/sandbox `project_id` boundaries introduced in #362.

## Test plan
- New `middleware/project-context.test.ts`: header scope for a member project, 403 for a same-org non-member project, 400 when the header is missing (dropped query fallback), API-key pinning that ignores the header, and 401 with no auth.
- New cross-project-isolation regression tests (404/exclusion) for tokens, counterparties, custody wallets, custody configs, and payment transfers belonging to a different project in the same org.
- `cd apps/sdp-api && pnpm test`


multipletwigs (dismissed):
LGTM

multipletwigs (dismissed):
lgtm

GuiBibeau (dismissed):

GuiBibeau (approved):

PR #391, opened 2026-06-02
fix: batch Cloudflare Worker secret uploads

GuiBibeau:
## Summary
- Add a `cloudflare-batches` projection mode that writes Worker secrets in JSON batches.
- Update the SDP API deploy workflow to upload each batch separately.
- Cover batching and empty-secret behavior in the script tests.

## Why
Cloudflare now rejects `wrangler secret bulk` requests with more than 25 secrets. The latest main deploy had 59 populated Worker secrets, so syncing everything in one request failed before deployment.

## Validation
- `pnpm test:scripts`
- `pnpm exec biome check --write scripts/project-secrets.mjs scripts/project-secrets.test.mjs .github/workflows/deploy-sdp-api.yml`
- `git diff --check`


G1de0n (approved):
LGTM

PR #403, opened 2026-06-04
feat(sdp-web): HOO-491 Surface why mint/burn is disabled in token modal

eldarik:
## Summary

  Surfaces a **human-readable reason next to the Mint and Burn submit buttons** in the issuance token modal when the action is disabled, so users no longer have to scan the form
  fields to work out *why* they can't submit (e.g. the signer authority isn't a controlled wallet, or the destination is denylisted).

  Previously the button just sat there greyed out with no explanation. Now a "Mint unavailable — …" / "Burn unavailable — …" note appears directly above the button, derived from
  the same conditions that drive the disabled state.

Mint modal showing the disabled-reason note above the button

  ## What changed

  - **`token-action-primary-forms.tsx`** — derives `mintDisabledReason` / `burnDisabledReason` (`signerUnavailableReason || validationReason`) and renders a note above each
  submit button whenever a reason exists and the form isn't pending. The button's `disabled` condition reuses the same derived value, so the note and the button state can't drift
  apart.
  - **`token-validation-message.tsx`** — adds an `announce` prop (default `true`) that controls whether the `aria-live="polite"` region is attached. The submit-area note passes
  `announce={false}` so screen readers don't announce the same error twice (once at the field, once at the button).

  Frontend only (`apps/sdp-web/.../issuance/[tokenId]/`). No API changes.

  ## Accessibility

  The submit-area note restates a message that's already announced under the relevant field (the validation reason is derived from the same field errors, which render with
  `aria-live`). It's rendered as a non-live `<p>` to stay visible without duplicating screen-reader output. `reserveSpace={false}` keeps it from shifting layout when absent.


multipletwigs (dismissed):
lets go

GuiBibeau (dismissed):

GuiBibeau (approved):

PR #437, opened 2026-06-11
feat(ramps): show recent transaction history for the selected counterparty

multipletwigs:
## Summary

QOL: once a counterparty is picked in the send/deposit wizard, a compact **Recent Transfers** section renders under the picker so you can see your history with them before continuing.

CleanShot 2026-06-11 at 23 01 42@2x


multipletwigs:
Addressed the pagination/filtering concern: the transfers list API now accepts a comma-separated status list (`status=completed,confirmed,finalized`), so the component fetches its 5 rows already filtered server-side instead of post-filtering a fixed page of 20. A counterparty whose recent transfers are all failed/pending now correctly surfaces older successful ones. Covered by a new handler test; the single-status form is unchanged.

GuiBibeau (approved):

PR #440, opened 2026-06-12
chore: rename missingApiKeys to avoid CodeQL clear-text-logging false positive

G1de0n:
## What

Rename the `missingApiKeys` variable in `scripts/check-env-contract.mjs` to `missingApiEnvVars`.

## Why

CodeQL's `js/clear-text-logging` rule flags the `console.error(report)` / `console.log(report)` calls in this script as "clear-text logging of sensitive information". It's a false positive: the variable only ever holds env-var **names** — it's built from `extractApiBindingVars()`, which greps the source for `env.X` references and returns the identifiers (e.g. `TRITON_API_KEY`, `SIGNING_PROVIDERS`). It never reads any env **values**. `API_LOCAL_ENV_KEYS` is only used inside a `.includes()` predicate and is never logged.

The alert fires purely because CodeQL's sensitive-data heuristic matches the `Keys` token in the variable name. Renaming it away from that pattern clears the alert with no behavior change.

## Changes

- `missingApiKeys` → `missingApiEnvVars` (3 references). No logic change.
- `node --check` passes.


multipletwigs (approved):
sick thanks!

PR #471, opened 2026-06-18
refactor(sdp-api): remove provisioning from on-ramp quotes

multipletwigs:
createOnrampQuote is now pure: it reads provisioned state from the counterparty's provider_data instead of running ensure*. Lightspark reads the resolved Grid customerId; BVNK reads the customer + rule entry via bvnkOnrampStatusFromProviderData. When the counterparty isn't fully provisioned, both throw counterpartyNotProvisioned (409) directing the caller to complete POST /counterparties/:id/requirements first.

Provisioning is now owned solely by the requirements advance endpoint. persistRampQuoteTransfer is unconditional since every successful on-ramp quote is fully provisioned.


multipletwigs:
@greptile-apps can u review again? ensured the proper error is thrown in case anything is missing in provider_data

eldarik (dismissed):

eldarik (approved):

PR #476, opened 2026-06-18
feat(sdp-api): route BVNK calls through static-egress proxy

G1de0n:
## Why

BVNK whitelists outbound traffic by source IP. To satisfy that, sdp-api's BVNK calls are routed through a reverse proxy with a static egress IP (`BVNK_API_BASE_URL` points at the proxy). Two coupled changes make this work without breaking BVNK's Hawk auth or the existing direct (no-proxy) mode.

## What

**1. Decouple Hawk signing host from request host**
`buildBvnkHawkAuthorizationHeader` previously derived the MAC host/port from the request URL. With a proxy in front, the request host is the proxy, so the signature would be computed for the wrong host → 401. The MAC is now signed with `BVNK_SIGNING_HOST` (default: the real sandbox/prod BVNK hostname) on a fixed port `443`, while the fetch still targets `BVNK_API_BASE_URL`. `resource` (path+query) and payload-hash are untouched — the proxy preserves them.

**2. Proxy shared-secret header**
The proxy returns `403` unless it sees `X-Proxy-Auth: <PROXY_SHARED_SECRET>`. The header is attached to outbound BVNK requests **only when the base URL is an explicit proxy override**:
- `request()` — gated on `BVNK_API_BASE_URL` being set
- `_discoverRails` — gated on `BVNK_RAMP_RAILS_API_BASE_URL` being set

This avoids ever sending the shared secret to the real BVNK host when no proxy is configured.

Headers are not part of the Hawk MAC canon, so `X-Proxy-Auth` does not affect the signature.

## Behavior preserved

- No `BVNK_API_BASE_URL` → signed host unchanged, no `X-Proxy-Auth`, direct access works as before.
- Defaults track mode: sandbox → `api.sandbox.bvnk.com`, prod → `api.bvnk.com`.

## New env

`BVNK_SIGNING_HOST`, `PROXY_SHARED_SECRET` — declared in `env.d.ts`, `runtime-env.ts`, `turbo.json`, and documented in `.dev.vars.example`. Both optional; behavior is unchanged when unset.

## Verification

- `pnpm typecheck` — clean on touched files (5 pre-existing unrelated module-resolution errors remain on `main`).
- `pnpm lint` — clean.


apps/sdp-api/src/lib/ramps/providers/bvnk.ts
```diff
@@ -139,16 +141,30 @@ function readBvnkConfig(env: Record<string, string | undefined>, mode: SdpEnviro
...
     throw new AppError("INTERNAL_ERROR", "BVNK API URL configuration is invalid.");
   }
 
-  return { auth: { authId, secretKey }, walletId, apiBaseUrl };
+  // Hawk signs the real BVNK host. When BVNK_API_BASE_URL points at a
+  // static-egress proxy, the request host differs from the signed host, so the
+  // signing host is configured separately (default tracks the sandbox/prod host).
+  const signingHost =
+    env.BVNK_SIGNING_HOST?.trim() ||
+    new URL(mode === "sandbox" ? BVNK_SANDBOX_API_URL : BVNK_PRODUCTION_API_URL).hostname;
```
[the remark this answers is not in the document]
G1de0n: Good catch — fixed. `BVNK_SIGNING_HOST` now goes through `new URL(...).hostname`, so an accidental `:443` (or any port suffix) is stripped before it reaches the Hawk canonical string.

Also addressed the related summary point: a misconfigured-proxy 403 was surfacing as a Hawk-credential error. Split the 401/403 branch in `mapBvnkErrorStatus` so a 403 now also points at the `PROXY_SHARED_SECRET` / `X-Proxy-Auth` setup when the request is routed through the egress proxy.

apps/sdp-api/src/lib/ramps/providers/bvnk.ts
```diff
@@ -323,10 +336,16 @@ function mapBvnkErrorStatus(
       `BVNK request was blocked at the edge (CloudFront/WAF, status ${status}) before reaching the API. This is typically IP rate-limiting, not a credential issue; retry shortly or from a different egress.`
     );
   }
-  if (status === 401 || status === 403) {
+  if (status === 401) {
     return new AppError(
       "PROVIDER_NOT_CONFIGURED",
```
multipletwigs: let's use the error from error.ts if we have it for provider not configured, if not create one

apps/sdp-api/src/lib/ramps/providers/bvnk.ts
```diff
@@ -139,16 +141,27 @@ function readBvnkConfig(env: Record<string, string | undefined>, mode: SdpEnviro
...
-    (mode === "sandbox" ? BVNK_SANDBOX_API_URL : BVNK_PRODUCTION_API_URL);
+    apiBaseUrlOverride || (mode === "sandbox" ? BVNK_SANDBOX_API_URL : BVNK_PRODUCTION_API_URL);
   try {
     new URL(apiBaseUrl);
   } catch {
     throw new AppError("INTERNAL_ERROR", "BVNK API URL configuration is invalid.");
   }
 
-  return { auth: { authId, secretKey }, walletId, apiBaseUrl };
+  const signingHostInput =
```
multipletwigs, line 153: great - this makes sure both self hosted or having a proxy works

multipletwigs (dismissed):
one nit but ill approve it

apps/sdp-api/src/lib/ramps/providers/bvnk.ts
G1de0n: Done — switched these to the `providerNotConfigured()` / `providerUnavailable()` helpers from `@/lib/errors` (they already existed and were imported). Converted the adjacent edge-block branch too so the whole function uses the helpers consistently; `RATE_LIMITED` / `INTERNAL_ERROR` stay on `new AppError` since there is no dedicated helper for those.

Also pushed a follow-up in the same branch: the two new env vars (`PROXY_SHARED_SECRET`, `BVNK_SIGNING_HOST`) were missing from the Cloudflare secret-sync allowlist (`scripts/secret-keys.mjs`), so the Worker would have started without them — added there.

multipletwigs (approved):
amazing, thank u

PR #478, opened 2026-06-18
feat: standardize ramp webhook verification & processing + capture settlement economics

multipletwigs:
## Summary

Standardizes how ramp webhooks (MoonPay, Lightspark, BVNK) are **verified** and **processed**, and captures the actual settled economics so on-ramp transactions render correctly.

Two logical commits:
1. `feat` — verification standardization + settlement economics modeling
2. `refactor` — standardized processing (fast-ack + consolidation)

Fees / exchange rate info exposed within provider data
CleanShot 2026-06-18 at 21 14 16@2x

## Before → After

| Area | Before | After |
|---|---|---|
| **Signature verify** | 3 hand-rolled blocks (`timingSafeEqual`, `createVerify`, `crypto.subtle`) duplicated per provider | one `verifyWebhookSignature` util (HMAC + ECDSA, constant-time, raw-body); providers just pass a descriptor |
| **Replay protection** | none — captured webhooks replayable forever | 300s tolerance window across all 3, from each provider's **signed** timestamp (MoonPay header `t`, Lightspark/BVNK body `timestamp`); single `WEBHOOK_TIMESTAMP_TOLERANCE_SECONDS` const |
| **Processing** | settlement applied inline before the 200 ack | ack 200 immediately, apply in `waitUntil` — slow DB never delays the ack |
| **Failure handling** | `try/catch` that logged and still returned 200 → events silently lost, never retried | fail-loud: errors surface via the dispatch `.catch`; no false 200 |
| **Handlers** | two byte-identical settlement handlers (moonpay.ts, lightspark.ts) | one `handleRampSettlementWebhook(provider, payload)` |
| **Settled amount** | on-ramp recorded the *quote estimate* (or null) → modal showed "Recipient gets —" | records the **actual** settled amount (on-ramp crypto / off-ramp fiat) |
| **Economics** | dropped on the floor (`provider_data: {}`) | per-provider raw settlement (`RampTransferSettlement`, both terminal states) captured in `provider_data`, surfaced per-provider in the transaction detail modal |


multipletwigs:
@greptile-apps can we check everything here and ensure that all the comments are already handled? thanks

G1de0n (approved):
LGTM

PR #481, opened 2026-06-19
fix(web): [PRO-1396] redirect counterparty not-found to list + branded 404 page

multipletwigs:
## What

- **Counterparty not-found redirect** — the counterparty detail page now redirects to `/dashboard/payments/counterparty` when the counterparty can't be found (e.g. after switching workspaces, where the ID belongs to another org), instead of rendering a 404.
- **SDP-branded 404** — adds a root `app/not-found.tsx` to replace the default Next.js/Vercel 404. Server component, matches the landing aesthetic (gradient + Solana logo), with **Back to dashboard** / **Go home** CTAs. As the root `not-found.tsx` it covers both unmatched URLs and any `notFound()` call.

## Why

Switching workspaces left users stranded on a 404 for a counterparty that simply belongs to a different org. Redirecting to the list is the expected behavior, and the branded 404 gives a consistent fallback for genuinely missing routes.

## Test

- `tsc --noEmit` + `biome check` pass on both files.

### DIFF
--- a/apps/sdp-web/src/app/dashboard/payments/counterparty/[counterpartyId]/page.tsx
+++ b/apps/sdp-web/src/app/dashboard/payments/counterparty/[counterpartyId]/page.tsx
@@ -1,5 +1,5 @@
 import { auth } from "@clerk/nextjs/server";
-import { notFound, redirect } from "next/navigation";
+import { redirect } from "next/navigation";
 import { getAuthEntryPath } from "@/lib/auth-entry";
 import { withDashboardPageTrace } from "@/lib/dashboard-page-trace";
 import { fetchCounterpartyDetail } from "../counterparty-detail.data";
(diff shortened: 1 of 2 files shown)


eldarik (approved):
LGTM ✅
Nits, none blocking:
- **[P3 — scope]** `issuance/[tokenId]` and `custody/[walletId]` have the same "ID belongs to another org after workspace switch" footgun but still `notFound()` without redirecting. They'll now show the new branded 404 (an improvement), but if redirect-to-list is the intended UX they'd want the same treatment — worth a follow-up.
- **[P3 — observability]** The trace `catch` logs intentional `redirect()`/`notFound()` throws as `{ ok: false, error: "NEXT_REDIRECT" }`. Pre-existing (not introduced here), but consider special-casing `isRedirectError`/`isHTTPAccessFallbackError` so control-flow throws don't pollute error traces.
- **[P3 — test]** Neither path is covered by a test. Low value for the static 404; the redirect branch is trivial enough to skip.

PR #482, opened 2026-06-19
feat(payments): confirm + cancel ramp transfers

multipletwigs:
## What

Adds a confirmation step + real cancellation for ramp wizard transfers. Once the `/quote` call persists a `payment_transfers` row, hitting **Cancel** in the wizard now (1) prompts a confirmation dialog and (2) actually flips the persisted row to `canceled` instead of silently navigating away.

### Problem
On/off-ramps check counterparty requirements, then call `/quote`, which **persists a transfer row**. Before this, the wizard's Cancel button just navigated away — leaving a live `pending`/`awaiting_payment` row orphaned, and it was far too easy to hit by accident.

CleanShot 2026-06-19 at 18 17 47@2x

## Changes

**API**
- `POST /v1/payments/ramps/transfers/cancel` (`payments:write`) — looks up the transfer by `provider` + `providerReference` (org/project scoped), guards that status is **cancelable** (`pending` / `awaiting_payment`), then sets it to `canceled` via the shared `updateTransferRecord` helper. Returns the updated transfer summary.
- New `canceled` value on `PaymentTransferStatus` (no migration — `payment_transfers.status` is free-text `TEXT`, no CHECK constraint).
- `canceled` added to `transferStatusSchema` (query filter) and the `listTransfers` non-chain status set so canceled rows still surface in listings.

**Web**
- `CancelTransactionDialog` (DS `Modal`) — "Keep going" / "Cancel transaction".
- `RampWizardShell` gains `confirmSecondary`; when set, the secondary button opens the dialog instead of firing immediately. Both rails pass `confirmSecondary={wizard.onTransactionStage}` (same flag that already drives the "Cancel" label).
- `useRampWizard.handleSecondary` routes the transaction stage through a new `cancelTransfer()` → calls the cancel endpoint, toasts, then exits.
- `isTerminalRampTransferStatus` now treats `canceled` as terminal, so the transfer-status SWR poll stops after cancel.
- Next proxy route for the new endpoint.

## Design notes
- **Local-only cancel.** Flips *our* row to `canceled`; does **not** call the provider's cancel API. An abandoned, unfunded quote just expires provider-side. Provider-side teardown can be a follow-up if needed.
- **Status guard fails loudly** — a transfer already `processing`/`settling`/`completed` returns 400 rather than faking a cancel on in-flight money.
- Cancel is keyed by `provider`+`providerReference` (what the wizard has client-side), under `/ramps/`.

## Out of scope / follow-ups
- Single source of truth for the transfer-status lists (`PaymentTransferStatus` / zod enum / non-chain array / terminal helper currently parallel) — pre-existing debt this change touches.
- This branch also carries in-progress **BVNK off-ramp dedicated-wallet provisioning** (`ensureBvnkOfframpWallet`), which is interleaved in `handlers/ramps.ts` and rides along here.

## Verification
- `tsc --noEmit` green on `sdp-api` and `sdp-web` (pre-existing unrelated errors aside); `biome check` clean. (ESLint is broken repo-wide.)
- Provider calls 503 locally without Doppler — endpoint logic verified by types + review, not a live ramp run.

### DIFF
--- a/apps/sdp-api/src/db/repositories/payments.repository.postgres.ts
+++ b/apps/sdp-api/src/db/repositories/payments.repository.postgres.ts
@@ -222,6 +222,27 @@ export function createPostgresPaymentsRepository(db: AppDb): PaymentsRepository 
       return getTransferByIdInternal(db, input.transferId);
     },
 
+    async updateTransferStatusGuarded(input) {
+      const scope = buildTransferScopeWhere({
+        organizationId: input.organizationId,
+        projectId: input.projectId,
+        extraClauses: ["id = ?", "status = ANY(?)"],
+        extraValues: [input.transferId, [...input.fromStatuses]],
+      });
+
+      const row = await db
+        .prepare(
+          `UPDATE payment_transfers
+           SET status = ?, updated_at = ?
+           WHERE ${scope.where}
+           RETURNING *`
+        )
+        .bind(input.toStatus, input.updatedAt, ...scope.values)
+        .first<Record<string, unknown>>();
+
+      return row ? mapTransferRow(row) : null;
+    },
+
     async getTransferById(params) {
       const scope = buildTransferScopeWhere({
         organizationId: params.organizationId,
--- a/apps/sdp-api/src/db/repositories/payments.repository.test.ts
+++ b/apps/sdp-api/src/db/repositories/payments.repository.test.ts
@@ -1,0 +1,166 @@
+import { afterAll, beforeAll, beforeEach, describe, expect, it } from "vitest";
+import { getDb } from "@/db";
+import { TEST_ORG, TEST_USER } from "@/test/fixtures/organizations";
+import { env } from "@/test/helpers/env";
+import { clearTestDatabase, seedTestDatabase } from "@/test/mocks/db";
+import type { PaymentsRepository } from "./payments.repository";
+import { createPostgresPaymentsRepository } from "./payments.repository.postgres";
+
+const TEST_PROJECT_ID = "prj_payments_repo_test";
+const OTHER_PROJECT_ID = "prj_payments_repo_test_other";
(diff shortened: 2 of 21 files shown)


eldarik (dismissed):
**Approve with nits — but please split or live-verify the BVNK off-ramp work** 🟡

The cancel feature itself looks solid: org/project-scoped lookup, a loud status guard (`pending`/`awaiting_payment` → 400 otherwise), reuse of the shared `updateTransferRecord`/`mapTransferRow`, `canceled` threaded consistently through the type/zod/non-chain/terminal lists, and `provider_reference` is persisted as `quote.id` so the wizard's `{provider, quote.id}` lookup matches. 👍

Main asks:
- **[Risk — scope]** This PR also bundles the BVNK off-ramp dedicated-wallet feature (`ensureBvnkOfframpWallet`, optional `customerReference`, `BvnkExecuteOfframpInput` reshape, `providerCustomer` → `externalCustomerId`/`walletId`). That changes off-ramp money-movement and is separately testable — strong preference to land it as its own PR.
- **[Risk — untested]** The off-ramp now sends `customerId: counterparty.external_id ?? counterparty.id` where it used to send the BVNK customer reference (and threw if missing). With provider calls 503 locally, this is types-only verified — please confirm with a live sandbox off-ramp before merge.
- **[P2]** With `customerReference` omitted, `getFiatWalletProfile` filters by `currency` only — confirm the merchant profile is uniquely resolvable per currency.
- **[P2]** Add a handler test for the cancel status guard (cancelable set + 400/404); the route-coverage line is a registration check, not behavior.
- **[P3]** Cancel read-then-update isn't transactional (harmless/idempotent in practice); and the four parallel status lists remain a latent footgun.

GuiBibeau (approved):

PR #488, opened 2026-06-22
fix(sdp-api): create the test database before running pnpm test

moviendome:
While setting up a fresh clone to keep contributing, `pnpm test` failed before running a single test. This chains the one missing setup step so the documented flow works, and corrects a diagnostic that pointed the wrong way.

## What happens today

On a clean checkout, the documented setup:

```
pnpm db:postgres:up
pnpm --filter @sdp/api db:postgres:bootstrap
pnpm --filter @sdp/api test
```

fails right away:

```
Caused by: error: database "sdp_test" does not exist
Serialized Error: { severity: 'FATAL', code: '3D000', routine: 'InitPostgres' }
```

Every workers test file fails the same way.

## Root cause

`db:postgres:bootstrap` creates and migrates `sdp`, but nothing creates `sdp_test`. The workers pool (`vitest.workers.config.ts`) binds `sdp_test` but doesn't create it, and the `test` script (`pnpm test:workers && pnpm test:node`) never ran the step that would. That step already exists: `db:migrate:test` creates and migrates `sdp_test` via `ensureDatabaseExists`. It just wasn't chained in.

## What this changes

`apps/sdp-api/package.json` — chain the existing step into `test`:
`"test": "pnpm db:migrate:test && pnpm test:workers && pnpm test:node"`

`db:migrate:test` is idempotent: `ensureDatabaseExists` guards `CREATE DATABASE` behind a `pg_database` existence check, and migrations report "up to date" when there's nothing new. So on a machine that already has `sdp_test`, the chained step is a fast no-op.

## Testing

Verified end to end on a fresh clone (v0.31.0):

- before: `pnpm test` fails every workers file with `3D000` / "database sdp_test does not exist"
- after: the same command creates `sdp_test` and runs green past the DB layer (727 passed)
- re-run: `db:migrate:test` prints "Test postgres migrations are up to date" and the suite runs again with no DB error, confirming the no-op path

`biome check` on the changed file is clean.

## CI

No CI changes needed. The fork-safe runner (`scripts/run-workspace-tests.mjs`) already runs `db:migrate:test` before the suite, so chaining it into `test` is a duplicate idempotent no-op there; on the Doppler path it adds the same idempotent step against the CI Postgres service. Either way it's the no-op described above, and the PR's own run exercises it.


multipletwigs (approved):

multipletwigs:
@moviendome sweet, thanks

GuiBibeau:
working on getting this merged

moviendome:
> working on getting this merged

Great! Thx 🚀

GuiBibeau (approved):
Approved after green CI.

PR #490, opened 2026-06-22
feat: set max transferrable amount in the UI for clarity

multipletwigs:
CleanShot 2026-06-22 at 15 35 59@2x

### DIFF
--- a/apps/sdp-web/src/app/dashboard/payments/ramps/components/amount-balance-readout.tsx
+++ b/apps/sdp-web/src/app/dashboard/payments/ramps/components/amount-balance-readout.tsx
@@ -1,0 +1,35 @@
+import { cn } from "@/lib/utils";
+
+export function AmountBalanceReadout({
+  available,
+  assetLabel,
+  exceeds,
+  onMax,
+}: {
+  available: string;
+  assetLabel: string;
+  exceeds: boolean;
+  onMax?: () => void;
+}) {
+  return (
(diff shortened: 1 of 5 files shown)


eldarik (approved):

multipletwigs:
Re: the flagged "shows 0 USDC / trips exceeds-balance during the wallet-select → balance-fetch gap" concern — that gap isn't reachable, so no loading state is needed here.

All four entry points (`pay`, `send`, `receive`, `deposit`) render the action page with `wallets={[]}`. In `useOnchainSendWizard`, that makes `fallbackData: wallets.length > 0 ? wallets : undefined` resolve to `undefined` — there are never balance-less fallback wallets.

So `liveWallets` is only ever:
- `[]` while the fetch is pending → the wallet combobox shows its own `isLoading` state and nothing is selectable, or
- the fully-fetched list from `fetchWallets({ includeBalances: true })` → balances present.

There's no window where you can select a wallet whose `balances` haven't loaded. The `availableAmount = 0` branch only fires when a *loaded* wallet genuinely holds 0 of the asset, which is the intended "0 USDC, nothing to send" readout.

One forward-looking note: the `wallets.length > 0` ternary is built to accept non-empty SSR-seeded wallets. If a page ever starts seeding wallets *without* balances as fallback, the gap becomes real — at which point the correct signal is `balances === undefined && isValidating` (not bare `balances === undefined`, which would shimmer forever on a genuinely empty wallet).

PR #509, opened 2026-06-23
ci(kora): STARTED + SUCCESS/FAILED Slack messages on deploy

niks3089:
Mirrors the **tokens** repo deploy notifications for the kora deploy.

**Before:** one message at the end (`[DONE]`/`[FAILED]`).
**After:** a `[STARTED]` message when the deploy job begins, then `[SUCCESS]`/`[FAILED]` when it finishes — same Block Kit format (Service, Version, Triggered by, Run link).

- `scripts/notify-slack.sh`: add `started` status (amber), rename `[DONE]` → `[SUCCESS]`
- `.github/workflows/deploy-kora.yml`: add a "Notify Slack — deploy started" step after the tag resolves; the existing `if: always()` step posts the result


niks3089:
Good catch @greptile — fixed in 6ce2574.

- **`Notify Slack — deploy started`** → `continue-on-error: true`. As flagged, it runs before GCP auth + the deploy, so a Slack 429/5xx under `set -euo pipefail` would have blocked the deployment.
- Also hardened **`Notify Slack — deploy result`** with `continue-on-error: true`. It's last + `if: always()`, but a Slack blip there would still flip an otherwise-successful job to *failed* — and on devnet that skips mainnet via the fail-fast matrix. Now Slack outcomes are fully decoupled from the deploy result.

Same change mirrored in sdp-infra PR #2.

waldruupi (approved):

PR #515, opened 2026-06-24
feat(payments): request-for-payment links with Solana Pay QR

multipletwigs:
## Request for Payments

Single-use payment-request links payable over Solana Pay.

> [!NOTE]
> **Settlement / reconciliation is deferred to a follow-up PR.** This PR ships request creation + the public pay surface. Detecting the on-chain payment and flipping a request to `paid` (cron sweep, on-demand status, live page update) lands separately, built on `@solana/pay`'s `findReference` + `validateTransfer`. The design is diagrammed below for context.

Payment Link Creation
CleanShot 2026-06-25 at 00 04 37@2x

CleanShot 2026-06-25 at 00 05 30@2x

Send this link to anyone to initiate payment! 
CleanShot 2026-06-25 at 00 05 56@2x

^ Tested on my devnet phantom mobile QR, works as expected!

### API (`sdp-api`)
- `POST /v1/payments/requests` — create a fixed-amount request (wallet, token, amount, optional counterparty + expiry); scoped to the caller's api-key wallet binding.
- `GET /v1/payments/requests` — paginated list.
- `GET /pay/:token` — public request metadata + a Solana Pay **transfer-request** URL (`solana:<recipient>?amount&spl-token&reference`), built with `@solana/pay`'s `encodeURL`. The wallet builds and submits the transfer itself; the `reference` rides in the URL so the payment stays reconcilable on-chain.
- The reference is minted in the repository alongside `id`/`public_token`.

> **Now built on the `@solana/pay` library (transfer requests).** The earlier hand-rolled **transaction-request** flow (`GET`/`POST /pay/:token/tx` + the local `solana-pay.ts` wrapper and its server-side transaction building) was removed in favor of the library — the wallet now constructs the transfer from the URL rather than the API serializing it.

### Web (`sdp-web`)
- Public `/pay/:token` page: a branded Solana Pay QR (rendered with `@solana/qr-code-styling`, including the embedded Solana mark), amount, and status panels.
- Dashboard **Requests** tab (create + list).

### Deferred to follow-up (not in this PR)
- Reconciliation: `@solana/pay` `findReference` (locate the tx by reference) → `validateTransfer` (confirm recipient + amount) → settle `paid`; plus the cron sweep, the on-demand `/pay/:token/status` reconcile, and the live page update.

### Verification
- `tsc --noEmit` clean on `sdp-api`; feature files + `biome` clean on `sdp-web`.

### End-to-end sequence (this PR)

```mermaid
sequenceDiagram
    autonumber
[...]

PR #522, opened 2026-06-25
feat(payments): settle payment requests on read via Solana Pay reference

multipletwigs:
## What

Payment requests now **settle on read**: when an awaiting request is viewed (public `GET /pay/:token` or the dashboard list), its Solana Pay `reference` is checked on-chain and, if a valid payment is found, the request is marked `paid` with a linked inbound transfer recorded.

This is the settlement path #515 shipped-then-reverted, rebuilt on `@solana/pay` instead of a hand-rolled validator.

## How

- `services/payments/payment-requests.ts` → `reconcilePaymentRequest(env, row)`: short-circuit non-awaiting/expired, then `findReference` → `validateTransfer` (`@solana/pay`, kit-native), then `createTransfer` (inbound) + `markPaymentRequest(paid)`. Wired into `pay.ts` and the list handler.
- Repo refactor (separate commit): `createTransfer` owns the `id` + timestamps and takes `signature`/`slot`, returning via `RETURNING` — one insert, no create-then-update.

## Decisions worth knowing

- **Settle-on-read, not push.** A request only flips to `paid` when someone looks; fine for now. `@solana/pay`'s `watchReference` is the push upgrade path if a non-human consumer ever needs it.
- **Reuse `@solana/pay`** for validation — it handles the failed-tx selection + SOL/SPL precision that #515 was reverted over.
- **Settlement persists a real inbound `payment_transfers` row** — required by the `CHECK (status != 'paid' OR fulfilled_by_transfer_id IS NOT NULL)`.
- **No KV coalesce**; the `/pay` route is already rate-limited. Add a cache/cap only if RPC cost spikes.

## Testing

`services/payments/payment-requests.test.ts` covers settle (SOL + SPL), not-found, invalid, rethrow-on-unexpected, and both short-circuits. `tsc`/`biome` clean; 123/123 across the 5 affected suites (DB tests: `--config vitest.workers.config.ts`).

Deferred (non-blocking): endpoint rename `/pay/:token` → `/payment-requests/:token`.

### DIFF
--- a/apps/sdp-api/src/db/repositories/payments.repository.postgres.ts
+++ b/apps/sdp-api/src/db/repositories/payments.repository.postgres.ts
@@ -10,6 +10,7 @@ import type {
   UpdatePaymentTransferInput,
   UpsertPaymentWalletPolicyInput,
 } from "./payments.repository";
+import { generatePaymentTransferId } from "./payments.repository";
 
 function buildInClause(length: number): string {
   return Array.from({ length }, () => "?").join(", ");
@@ -87,18 +88,6 @@ function buildTransferScopeWhere(params: {
   };
 }
 
-async function getTransferByIdInternal(
-  db: DatabaseExecutor,
-  transferId: string
-): Promise<PaymentTransferRow | null> {
-  const row = await db
-    .prepare("SELECT * FROM payment_transfers WHERE id = ?")
-    .bind(transferId)
-    .first<Record<string, unknown>>();
-
-  return row ? mapTransferRow(row) : null;
-}
-
 async function getWalletPoliciesInternal(
   db: DatabaseExecutor,
   custodyWalletId: string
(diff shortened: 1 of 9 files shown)


multipletwigs:
@greptileai re-review please. Addressed the concurrency finding: `reconcilePaymentRequest` now catches the `payment_transfers.signature` UNIQUE violation (via `isPostgresUniqueViolation`) on concurrent reconciles of the same request, converges on the transfer the winning caller recorded, and re-reads the request when `markPaymentRequest` finds it already settled — so both callers return the paid state and exactly one transfer row exists. Added a concurrent-reconcile test.

multipletwigs:
@greptileai re-review please. Two follow-ups since the last review:

1. **Concurrent settlement (prior 3/5 finding):** `reconcilePaymentRequest` now catches the `payment_transfers.signature` UNIQUE violation on concurrent reconciles of the same request, converges on the transfer the winner recorded, and re-reads the request when `markPaymentRequest` loses the race (no more stale `awaiting_payment`). Added a concurrent-reconcile test.

2. **List 500 on a single RPC failure (P1):** added `reconcilePaymentRequestBestEffort`; the list (`Promise.all` over up to 100 rows) and the public pay page now degrade to the stored row and log on unexpected RPC errors instead of failing the whole response. Added a best-effort degradation test.

The strict `reconcilePaymentRequest` primitive is unchanged for settle callers.

eldarik (approved):

PR #526, opened 2026-06-26
fix(observability): kora-devnet metrics + dashboard, drop deprecated kora-sdp

niks3089:
The Kora paymaster dashboard's `service` variable mapped **`devnet → kora-sdp`** (the pre-cutover service), so the devnet view showed the **old** fee-payer (`CsXg…`) instead of the live `kora-devnet` payer (`GSDYH3…`).

**Fix:** repoint `devnet → kora-devnet` (every panel filters by `$service`, so this is the only change). Bumped dashboard version 2→3.

**Already deployed live** via `grafana-sync.sh` (doppler `prd`) — `dashboard kora.json: 200` — so https://sdp.grafana.net/d/kora-paymaster now shows the correct devnet payer. This PR brings the repo source of truth in line.

### DIFF
--- a/alert-rules/kora-5xx-ratio-high.json
+++ b/alert-rules/kora-5xx-ratio-high.json
@@ -3,96 +3,101 @@
   "title": "kora: 5xx ratio > 1% for 5m",
   "ruleGroup": "kora-paymaster",
   "folderUID": "sdp",
-  "condition": "C",
+  "condition": "E",
   "for": "5m",
   "noDataState": "OK",
   "execErrState": "Error",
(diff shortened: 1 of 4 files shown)


niks3089:
Superseded by #532. Rather than fix the kora dashboard/alerts here, kora observability is being removed from this repo entirely and lives only in sdp-infra (the corrected versions are in solana-foundation/sdp-infra#8, and already pushed live to Grafana). Closing without merge — no revert needed since the live dashboards/alerts already carry these fixes.

PR #531, opened 2026-06-29
fix: standardize UI for lightspark/bvnk and ramps in general

multipletwigs:
## What

Standardize the ramp wizard UI for onramp/offramp (lightspark + bvnk). The two step-content files had drifted into divergent copies of the same flow — offramp was missing a success screen, onboarding copy was inline and provider names leaked. This pulls the shared structure into reusable components and moves per-provider copy into per-provider files.

## Before / after file structure

**Before** — two divergent files, copy inline:

```
ramps/components/
  onramp-step-content.tsx     status copy + onboarding panels + complete screen, all inline
  offramp-step-content.tsx    divergent copy of the same — NO complete screen, drifting copy
```

**After** — shared components + per-provider copy files; routers go thin:

```
ramps/components/
  providers/
    index.ts                  onboardingCopy() / provisioningDetail() / simulateActionLabels() resolvers
    bvnk.ts                   bvnk onboarding copy (per status, w/ icon) + provisioning detail + simulate labels
    lightspark.ts             lightspark onboarding copy + provisioning detail + simulate labels
  ramp-status-panel.tsx       transfer status copy/icon — shared, switched by direction
  ramp-onboarding-panel.tsx   onboarding lifecycle panel — shared, + pulsing provisioning dot
  ramp-complete-screen.tsx    success screen — shared, now used by offramp too
  onramp-step-content.tsx     thin router
  offramp-step-content.tsx    thin router
```

The only per-provider difference now lives in `providers/*`; direction copy lives in the shared components. Offramp gains the success screen it was missing.

### DIFF
--- a/apps/sdp-web/src/app/dashboard/payments/ramps/components/offramp-step-content.tsx
+++ b/apps/sdp-web/src/app/dashboard/payments/ramps/components/offramp-step-content.tsx
@@ -1,8 +1,8 @@
 "use client";
 
 import { getCryptoRailAssetLabel } from "@sdp/types/payment-rails";
-import { CheckCircle2Icon, SendIcon, WalletIcon, XCircleIcon } from "lucide-react";
-import { useEffect, useMemo, useState } from "react";
+import { SendIcon, WalletIcon } from "lucide-react";
+import { useMemo } from "react";
 import { Combobox } from "@/components/ui/combobox";
 import { OFFRAMP_PAIRS, RAMP_PROVIDER_OPTIONS, toRampCryptoToken } from "@/lib/ramps";
 import type { OfframpWizard } from "../hooks/use-offramp-wizard";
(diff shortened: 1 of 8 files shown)


multipletwigs:
@greptileai please re-review — addressed both findings in 41b81b2:

- **P1 (Ready Providers Crash):** added `hasOnboardingLifecycle(provider)` (bvnk/lightspark only) and gated both onboarding-panel branches on it. MoonPay/MoneyGram report `status: "ready"` and now fall through to the quote-loading skeleton instead of hitting `onboardingCopy`'s throw.
- **P2 (Unknown Provider Uses Lightspark Labels):** `simulateActionLabels` is now a `switch` with an exhaustive `never` default that throws, instead of a ternary that silently returned Lightspark labels.

multipletwigs:
@greptileai re-review please — addressed the manual-instructions crash finding in cafabdb. `simulateActionLabels` now returns `null` for providers without simulate support (no throw, no silent Lightspark default), and the onramp manual-instructions branch omits the optional simulate action when labels are absent, keeping the quote screen usable.

eldarik (approved):

PR #533, opened 2026-06-29
fix(sdp-web): compliance screening with provider risk cards

multipletwigs:
https://github.com/user-attachments/assets/e07d88c0-7943-46b8-af04-d21c7c2ce9d8

- Animate per-provider screening on the add-external-account form: staggered checking -> pass/flag reveal with muted provider logos; modal height grows in
- Replace the risky-confirm modal with an inline flow: Add account when clear, hold-to-add (new HoldButton) when any provider flags or cannot be screened, with an animated warning line
- Disable inputs while screening so the screened address cannot desync from the submitted one
- HoldButton (components/ui): reuses the design-system button via asChild, pointer + keyboard hold, clip-path fill
- Strongly type ComplianceProviderResult.provider as ComplianceProviderId
- Counterparty account rows: fixed-height address line, Solana logo + label
- Show the full address (not middle-truncated) in the destination dropdown
- turbo.json: pass compliance provider keys through Turbo strict env mode so screening runs


multipletwigs:
@greptileai please re-review — addressed the 4/5 findings (empty-results stuck state, unknown-provider logo crash, duplicate-submit guard) plus simplification cleanups.

arseniy-nikitochkin (approved):

PR #535, opened 2026-06-29
feat(sdp-api): batch payment transfers to multiple counterparties

multipletwigs:
## Summary

Implements **batch payments** for `sdp-api` (PRO-1446, phase 8 part 2): send **SOL** or an **SPL / token-2022** token from a single SDP custody wallet to **N counterparty crypto-wallet accounts** in one API call. Recipients are size-packed into one or more **sponsored** Solana transactions, executed and confirmed server-side, with full per-recipient state tracking.

Builds on the persistence layer (table + repository scaffold) already in `main`.

## Endpoints

| Method | Path | Permissions | Purpose |
|---|---|---|---|
| `POST` | `/v1/payments/transfer-batches/estimate` | `payments:read`, `wallets:read`, `counterparties:read` | Dry-run: recipient/transaction counts + fee & rent estimate. No DB writes, no signing. |
| `POST` | `/v1/payments/transfer-batches` | `payments:write`, `wallets:read`, `counterparties:read` | Create + execute a batch. |
| `GET` | `/v1/payments/transfer-batches` | `payments:read` | List batches (wallet/token/status/externalId filters, paginated). |
| `GET` | `/v1/payments/transfer-batches/:batchId` | `payments:read` | Batch detail + recipients + per-chunk transfers. |

Request shape (`createTransferBatchSchema`): `{ source, token, recipients[1..500]{ counterpartyId, counterpartyAccountId, amount, externalId? }, externalId?, options?{ maxRecipientsPerTransaction(1..64), preflight, priorityFee } }`.

## The flow

**1. Validate & resolve (`resolveBatchRequest`)**
- Parse body; resolve auth scope; assert the input `projectId` matches the key's project.
- Resolve the `source` custody wallet and assert the API key is scoped to it.
- Resolve **token context**: `SOL` → 9 decimals; otherwise resolve the mint's token program + the source token account (and its decimals) on-chain.
- Resolve **recipients** concurrently — each recipient's counterparty + counterparty-account are fetched (org/project-scoped), the account is required to be a `crypto_wallet` with `details.network = "solana"`, and the destination address is extracted. Amounts are parsed to base units (bigint) and summed for the batch total.

**2. Policy enforcement (`enforceBatchPolicies`, create only)**
- Per recipient: destination-allowlist + per-transfer max-amount.
- Once for the aggregate: the per-transaction max-amount against the **batch total** (so it can't be split-bypassed) and the **daily limit** against the day's running outbound total. Denials are recorded via the legacy wallet-policy denial path.

**3. Build instructions (`buildInstructionGroups`)**
- **SOL**: one `transferSol` instruction per recipient.
- **SPL/token-2022**: `createAssociatedTokenIdempotent` (destination ATA) + `transferChecked` per recipient. ATAs are derived via `findAssociatedTokenPda`.

**4. Pack into transactions (`chunkInstructionGroups`)**
- Greedy bin-packing under **two limits**: a recipient-count cap (`maxRecipientsPerTransaction`, default 20) and the real constraint — the **1232-byte** v0 transaction size limit, checked by compiling each candidate and calling `isTransactionWithinSizeLimit`. A single recipient that can't fit alone is a `400`.
- Each resulting chunk is a self-contained transaction message (fee payer = sponsor, lifetime = recent blockhash, source as signer).

**5. Persist (create only)**
- Insert one `payment_transfer_batches` row (`processing`) and **bulk-insert** all `payment_transfer_recipients` rows (`pending`) in a single statement.

**6. Execute each chunk (`executeChunk`, sequential)**
- Partially sign with the source signer; create a `payment_transfers` row (`type: transfer_batch`, `processing`) for the chunk; bulk-update the chunk's recipients to `processing`.
- Optional **preflight** simulation (`options.preflight`, default on; `sigVerify: false`).
- Hand the bytes to the fee-payment adapter (`signAndSend`) — the sponsor co-signs and submits — then confirm.
- On success: bulk-mark recipients + transfer `confirmed`. On failure: bulk-mark `failed` with the error. Chunks run sequentially so the source wallet emits one transaction at a time.

**7. Finalize**
- Batch status derived from the chunk transfers: all confirmed → `confirmed`, none → `failed`, mixed → `partially_failed`.
- Response: `{ batch, recipients[], transfers[] }`.

The **estimate** path runs steps 1, 3, 4 with a noop signer, then reports recipient/transaction counts, network fee (`getFeeForMessage` per chunk), and missing-ATA rent — no DB writes, no signing.

## Data model

- `payment_transfer_batches` — one row per batch (source wallet, token, totals, counts, status).
- `payment_transfer_recipients` — one row per recipient, linked to its chunk's `transfer_id`, with its own status/error.
- `payment_transfers` — one row per **chunk** (`type: transfer_batch`); `counterpartyId`/`destination` are set only for single-recipient chunks, and `providerData.recipientIds` always links the recipients. Reconciliation by counterparty goes through `payment_transfer_recipients`.

## Efficiency

- Recipient resolution and fee/rent estimation **fan out** (`Promise.all`) instead of serial loops.
- Recipient create and per-chunk status updates use **bulk SQL** (one multi-row `INSERT`, one `UPDATE ... WHERE id IN (...)`) instead of N round-trips.
- Chunk *execution* stays sequential on purpose (one on-chain tx at a time from the source wallet; deterministic partial-failure accounting).

## Testing

`transfer-batches.test.ts` (workers pool, real Postgres): SOL estimate; multi-chunk SOL create (asserts 2 transactions, confirmed batch, recipient/transfer rows, detail + list); SPL create for both `spl-token` and `token-2022`. `tsc --noEmit` and biome clean.

## Known limitations (devnet, deferred)

- **No idempotency** — a retried create with the same `externalId` re-sends the batch. Devnet only for now; fix later by promoting the `(organization_id, external_id)` partial index to `UNIQUE`.
- **`options.priorityFee` is accepted but not yet applied** (no compute-budget instruction; estimate reports `0`).
- If a chunk throws *outside* `executeChunk`'s inner try (e.g. signing), the batch is left `processing` rather than finalized — no outer reconciliation yet.

### DIFF
--- a/apps/sdp-api/src/db/repositories/payment-transfer-batches.repository.postgres.ts
+++ b/apps/sdp-api/src/db/repositories/payment-transfer-batches.repository.postgres.ts
@@ -16,6 +16,7 @@ import type {
   PaymentTransferRecipientRow,
   UpdatePaymentTransferBatchInput,
   UpdatePaymentTransferRecipientInput,
+  UpdatePaymentTransferRecipientsStatusInput,
   UpsertPaymentTransferBatchInput,
   UpsertPaymentTransferRecipientInput,
 } from "./payment-transfer-batches.repository";
@@ -325,6 +326,12 @@ export function createPostgresPaymentTransferBatchesRepository(
       if (input.walletId) {
         extraClauses.push("source_wallet_id = ?");
         extraValues.push(input.walletId);
+      } else if (input.walletIds) {
+        if (input.walletIds.length === 0) {
+          return { rows: [], total: 0 };
+        }
+        extraClauses.push(`source_wallet_id IN (${input.walletIds.map(() => "?").join(", ")})`);
+        extraValues.push(...input.walletIds);
       }
       if (input.token) {
         extraClauses.push("token = ?");
@@ -425,6 +432,54 @@ export function createPostgresPaymentTransferBatchesRepository(
       return mapPaymentTransferRecipientRow(row);
     },
 
+    async createTransferRecipients(inputs: CreatePaymentTransferRecipientInput[]) {
+      if (inputs.length === 0) {
+        return [];
+      }
+
+      const rowPlaceholder = `(${Array(12).fill("?").join(", ")})`;
+      const values = inputs.flatMap((input) => [
+        generatePaymentTransferRecipientId(),
+        input.batchId,
+        input.organizationId,
+        input.projectId,
+        input.transferId ?? null,
+        input.externalId ?? null,
+        input.counterpartyId,
+        input.counterpartyAccountId,
+        input.destinationAddress,
+        input.amount,
+        input.status ?? "pending",
+        input.error ?? null,
+      ]);
+
+      const result = await db
+        .prepare(
+          `INSERT INTO payment_transfer_recipients (
+             id,
+             batch_id,
+             organization_id,
+             project_id,
+             transfer_id,
+             external_id,
+             counterparty_id,
+             counterparty_account_id,
+             destination_address,
+             amount,
+             status,
+             error
+           ) VALUES ${inputs.map(() => rowPlaceholder).join(", ")}
+           RETURNING *`
+        )
+        .bind(...values)
+        .all<Record<string, unknown>>();
+
+      if (result.results.length !== inputs.length) {
+        throw internalError("payment_transfer_recipients bulk INSERT returned an unexpected count");
+      }
+      return result.results.map(mapPaymentTransferRecipientRow);
+    },
+
     async upsertTransferRecipient(input: UpsertPaymentTransferRecipientInput) {
       const recipientId = input.recipientId ?? generatePaymentTransferRecipientId();
       const row = await db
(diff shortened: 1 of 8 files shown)


multipletwigs:
@greptileai the two P1s are addressed in 4b7c8c1: blockhash is now refreshed per chunk before signing, and a post-submit confirmation timeout keeps the transfer `processing` with its signature (only a definitive on-chain error marks it `failed`). Added failure-path tests. Please re-review.

multipletwigs:
@greptileai both remaining P1s (ambiguous `signAndSend` failure on line 773, and batch/recipient rows staying `processing` after a future reconciliation on line 789) are the same exactly-once-delivery problem: safely resolving a submission whose response was lost, or a chunk that times out and later confirms, requires **sponsor-side idempotency plus a reconciler** that re-checks pending signatures and propagates status back to the batch and recipient rows.

That reconciliation + idempotency layer does not exist in the platform yet and is intentionally **out of scope** for this PR. This endpoint is **devnet-only** for now, so there is no real-funds exposure, and the work is tracked separately.

What this PR already does on that path: refreshes the blockhash per chunk (your earlier P1), and on a confirmation timeout keeps the chunk `processing` **with its signature persisted** rather than marking it `failed` — so a future reconciler has the signature to resolve it. The two cases above are knowingly accepted until the reconciliation layer lands.

Re-review requested with that context.

arseniy-nikitochkin (approved):

PR #549, opened 2026-06-30
feat(payments): batch send — recipient selection, bulk import, custom mints

multipletwigs:
## Summary

Phase 8 batch payments for all providers (PRO-1335). Adds an onchain batch-send flow to Payments v2: pick a source wallet + asset, choose recipients from a counterparty-grouped wallet table or bulk-import them by pasting rows, review, and send to many counterparties in one batch. Supports any SPL token by mint, not just well-known symbols.

## API (`sdp-api`)

- **Crypto-recipients listing** — `GET /v1/counterparties/crypto-recipients` returns active Solana `crypto_wallet` accounts (counterparty name + address + label), with search and pagination. Adds an `ids` filter so a specific set of wallet ids can be resolved in one query (powers bulk import).
- **Counterparty account repo** — `getCounterpartyAccountByIdInProject` and `listCounterpartyAccountsByIdsInProject`, scoped by org + project.
- **Batch transfers — optional `counterpartyId`** — `recipients[].counterpartyId` is now optional. When omitted, the account is resolved within org/project scope and the counterparty is derived from the account row (authoritative, unspoofable). When supplied, the verified counterparty-scoped lookup is kept (defense in depth). This lets the paste flow send with just `{ counterpartyAccountId, amount }`.

## Web (`sdp-web`)

- **Batch send rail + recipients step** — counterparty-grouped, expandable wallet table; source-wallet/asset selection with the available balance shown inline in the wallet trigger (shakes once + turns red when the total exceeds balance); single/batch send-mode toggle wired through the shared ramp wizard shell.
- **Bulk import dialog** — Vercel-style paste/rows UI: paste `counterparty_wallet_id, currency-or-mint, amount` lines into any field to explode them into editable rows, or add rows manually. Resolves wallet ids to recipients, enforces one currency per batch, and surfaces per-row + unresolved errors.
- **Custom mint support** — currency column accepts a well-known symbol or a raw base58 mint; resolves against the source wallet's balances.

## Testing

- `tsc --noEmit` clean (both apps), `biome check` clean on touched files.
- `sdp-api` batch transfer suite green (7/7), including a new test for the derive-counterpartyId path.

## Notes

- Legacy send/receive is untouched; this is additive to the v2 ramp flow.
- Batch idempotency remains out of scope (devnet demo).


multipletwigs:
@greptileai re-review please — addressed both findings:
- `resolveRecipients` now does a single bulk `listCounterpartyAccountsByIdsInProject` lookup instead of 1–2 queries per recipient (no more N concurrent queries)
- added `accountIds` IN-clause filter coverage to `listBatchRecipients` repo tests (incl. search+ids bind-order guard)

multipletwigs:
Addressed all 3 P2 findings:
- **`pageSize: 1` hack** — `fetchBatchRecipients` now treats `page`/`pageSize` as optional and id-resolution omits them, so the call-site no longer depends on the server's silent limit override (and avoids the `pageSize > 100` validation trap a literal `ids.length` would hit).
- **>500 silent dead-end** — `bulkImport` now throws a clear `A batch can have at most 500 recipients…` message (surfaced in the dialog) instead of silently disabling Review; extracted `MAX_BATCH_RECIPIENTS`.
- **React key collision** — bulk-import error list now uses positional keys.

@greptileai re-review please.

G1de0n (dismissed):

multipletwigs:
@greptileai please re-review — pushed `60ddaf6e`.

Changes since last review:
- **`MAX_BATCH_RECIPIENTS`** raised from the page size (6) to the API cap (500).
- **Single asset-options API**: batch send now reuses `walletBalanceAssetOptions` (shared with onchain-send + recurring); deleted the batch-only `resolveWalletAssetOptions` / `findWalletBalanceForDisplayToken`.
- **Asset keyed on mint** (like onchain-send); mint→symbol translation for display goes through the shared well-known token registry.
- New reusable `@sdp/types` helpers: `wellKnownMint`, `isWellKnownTokenSymbol`, `CLUSTER_BY_SDP_ENVIRONMENT`; bulk-import resolves `currency`→mint by project cluster (sandbox=devnet, production=mainnet-beta).
- Fixed a merge-botched `resolveAggregateBalanceDisplayToken` (duplicate block + dropped `.toUpperCase()`).

multipletwigs:
@greptileai please re-review — pushed `7f0963ea`.

Addressed the review findings:
- **bulkImport partial-state (the 4/5 blocker)**: now returns `{ unresolved }` *before* mutating any state, so a cancel on a partial resolve can't leave ghost recipients in the batch.
- **Currency-change merge (wrong-asset payment)**: switching the imported currency now *replaces* the batch instead of merging, so recipients selected under a different asset can't be paid in the new one. Comparison is against the resolved mint.
- **Unsendable currency**: throws a clear error when a well-known symbol has no mint on the active cluster (e.g. USDT on devnet).
- **>500 recipients**: explicit "A batch can have at most 500 recipients" message instead of a silently disabled Review button.

Notes on the other two inline comments: the duplicate-error-key issue is already fixed (`bulk-import-dialog.tsx` uses `key={index}` with a biome-ignore), and the ID-resolution call already omits `pageSize` (`fetchBatchRecipients({ ids })`), matching the suggested approach.

G1de0n (approved):

PR #574, opened 2026-07-02
fix(api): surface typed errors from on-chain submit and confirm paths

### LINKED ISSUE #567: Issuance token deploy returns a generic 500 for every template (priority)
**Area:** API / issuance 

Token creation succeeds, but the deploy step fails with a generic `500` and no detail, for both the `stablecoin` and `custom` templates. The token stays in the `pending` state and cannot be minted.

**Reproduce**
```bash
curl -X POST "$BASE/v1/issuance/tokens" -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"name":"X","symbol":"XX","template":"stablecoin","signingWalletId":"$WALLET_ID"}'
#   201, status pending
curl -X POST "$BASE/v1/issuance/tokens/$TOKEN_ID/deploy" -H "Authorization: Bearer $KEY" -d '{}'
```
**Deploy response:** `500`
```json
{"error":{"code":"INTERNAL_ERROR","message":"An internal error occurred"}}
```
Reproduced across two tokens and three attempts; a subsequent mint returns `400 TOKEN_NOT_ACTIVE`. The signing wallet holds SOL for fees, and the same key created the token, so this is neither a funding nor a permission problem.

**Impact:** the create, deploy, and mint path cannot complete, and the error is opaque. If the failure is not specific to a single project, it would affect issuance broadly.
**Suggested fix:** return an actionable error identifying what failed (signer resolution, RPC, extension initialization), investigate the custody deploy path, and confirm whether the failure is environment-wide.

multipletwigs:
## Problem

`POST /v1/issuance/tokens/:id/deploy` failed with a bare `500 {"code":"INTERNAL_ERROR"}` for every template, leaving tokens stuck in `pending` (mint then 400s `TOKEN_NOT_ACTIVE`) with no way for the caller to see what broke. Investigation on #567.

CleanShot 2026-07-02 at 15 12 54

Errors are now shown exactly as what they are instead of internal error 

```
{
  "ok": false,
  "status": 503,
  "statusText": "Service Unavailable",
  "body": {
    "error": {
      "code": "PROVIDER_UNAVAILABLE",
      "message": "The signing provider is temporarily unavailable. Try again."
    },
    "meta": {
      "requestId": "web_b66d8e8826754acda775e7c6102c2f59:01"
    }
  }
}
```

Closes #567

## Root cause

The custody deploy path's on-chain submit/confirm steps threw **plain `Error`s**, and the global error handler in `app.ts` maps typed errors to actionable responses but collapses any plain `Error` to `INTERNAL_ERROR`. The real failure (on-chain program error or confirm timeout) was already logged server-side under the trace ID — it just never reached the response.

## Changes

- **`services/mosaic/service.ts`** — both `confirmation.err` sites in `signAndSubmit` now throw `AppError("TRANSACTION_FAILED", ...)` carrying the full stringified on-chain error.
- **`services/solana/rpc.ts`** — both confirm-timeout sites (`sendAndConfirmTransaction`, `confirmTransaction`) now throw `AppError("SOLANA_RPC_ERROR", ...)` with signature + timeout; also aligned the two messages (one previously omitted the timeout).
- Both error codes already existed in `lib/errors.ts` — no new codes, no changes to the global handler's catch-all, no change to the deploy handler (its re-throw and DB error-recording pass the typed errors through cleanly).

**Blast radius (intentional):** `signAndSubmit`/`confirmTransaction` are shared by payments transfers/batches, custody, and recurring payments — all those callers now surface the same detail instead of opaque 500s. Their suites pass unchanged.

**Not changed:** `app.ts` fee-payment mapping (already implemented upstream via `mapFeePaymentError`); mosaic-SDK build errors (deferred per plan).

## Tests

- New unit test: `signAndSubmit` with an on-chain error → `AppError` `TRANSACTION_FAILED`/400 with the exact `InstructionError` detail.
- New `rpc.test.ts`: both confirm helpers → `SOLANA_RPC_ERROR`/502 with signature + timeout in the message.
- New route-level deploy tests: on-chain failure → `400 TRANSACTION_FAILED`, confirm timeout → `502 SOLANA_RPC_ERROR`, detail intact in `error.message`.

## Verification

`tsc --noEmit` clean; biome clean; vitest: mosaic + rpc + app 18/18, issuance 107 pass, payments + transfer-batches 84/84, custody + recurring + fee adapters 39/39.

Note: this PR makes the deploy failure *visible*; diagnosing and fixing the underlying devnet deploy failure (per the trace-ID runbook in the investigation) is follow-up work on #567.

### DIFF
--- a/apps/sdp-api/src/lib/errors.ts
+++ b/apps/sdp-api/src/lib/errors.ts
@@ -212,6 +212,10 @@ export function transactionFailed(message?: string, details?: Record<string, unk
   return new AppError("TRANSACTION_FAILED", message, details);
 }
 
+export function solanaRpcError(message?: string, details?: Record<string, unknown>): AppError {
+  return new AppError("SOLANA_RPC_ERROR", message, details);
+}
+
 export function providerNotConfigured(message?: string): AppError {
   return new AppError("PROVIDER_NOT_CONFIGURED", message);
 }
--- a/apps/sdp-api/src/routes/issuance.test.ts
+++ b/apps/sdp-api/src/routes/issuance.test.ts
@@ -7,6 +7,7 @@ import { findAssociatedTokenPda, TOKEN_2022_PROGRAM_ADDRESS } from "@solana-prog
 import { afterAll, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
 import { getDb } from "@/db";
 import app from "@/index";
+import { AppError } from "@/lib/errors";
 import { hashString } from "@/lib/hash";
 import type { Address } from "@/lib/solana";
 import * as AuthorityResolution from "@/routes/issuance/handlers/authority-resolution";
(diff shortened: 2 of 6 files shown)


multipletwigs:
@greptileai please re-review — the branch has moved past your last reviewed commit (afef80c7): 7e32dd75 switches the throw sites to the errors.ts factory helpers (adds a solanaRpcError factory), and 252701cc adds the direct-submit-branch unit test your review suggested.

eldarik (approved):

eldarik (approved):

PR #587, opened 2026-07-03
fix(API): Detect Postgres unique violations by SQLSTATE, not SQLite text

tomiet:
The allowlist create handler and Clerk webhook org-mapping recovery both detected duplicate-key errors with `err.message.includes("UNIQUE constraint")` SQLite error string. The data layer is Postgres, which raise SQLSTATE 23505 with the message "_duplicate key value violates unique constraint ..._", so neither branch ever matched:

- POST /allowlist with a duplicate entry returned `500` instead of `409`
- Concurrent Clerk org events (or a Svix retry) fail to converge on existing mapping and surface a `500` instead of recovering.

Switch both sites to existing `isPostgresUniqueViolation()` helper (already used in payment-requests and token.service) and add handler tests covering `409` conflict path plus regression guard that SQLite-style kinda message alone is not treated as a conflict


GuiBibeau (dismissed):
lgtm 👍

GuiBibeau (approved):
Approved after addressing the audit-log assertion thread.

GuiBibeau (dismissed):
Approved after branch update and focused test assertion.

tomiet:
Had to rebase onto latest main for branch protection. **Same** diff, new commit hash. Mind re-approving @GuiBibeau?

multipletwigs (dismissed):

multipletwigs:
@tomiet had to fix a few tests

multipletwigs (approved):

multipletwigs (approved):

pashpashkin (approved):

tomiet:
Thanks for the fix and for the review @multipletwigs!

PR #593, opened 2026-07-06
chore: Harden npm and GitHub Actions supply chain

pashpashkin:
## What changed

- GitHub Actions — third-party actions pinned to commit SHAs
- Package-manager config — All pnpm settings now live only in `pnpm-workspace.yaml`, `.npmrc` trimmed
  to `registry` + `audit`
- Deleted unused `bun.lock` and its three dead `Dockerfile.dockerignore` entries

## Why

Mutable action tags are a live supply-chain vector: a retagged third-party
action runs arbitrary code with access to `DOPPLER_TOKEN` / GCP identity.
On the config side, the old `.npmrc` keys worked for pnpm but are unknown
keys to npm (npm's cooldown key is `min-release-age`, in days) — npm 11
warns on every run and npm 12 will hard-error — and pnpm 11 stops reading
non-auth settings from `.npmrc` entirely. This PR moves everything to where
both current and future tooling actually reads it, and pins the registry.

## Result Validation

- `pnpm install --frozen-lockfile` passes on the pinned pnpm 10.16.0
- All four moved settings verified to resolve from `pnpm-workspace.yaml`
  alone, empirically on pnpm 10.16 and pnpm 11 
- All three action SHAs verified against upstream tags via the GitHub API
- `npm config get` runs warning-free

## Notes

Pinning scope is deliberate: **third-party actions → SHA; GitHub-owned
(`actions/*`, `github/*`) stay on version tags** — the same party controls
the action, the runner, and pin resolution, so a SHA adds nothing there.
Dependabot keeps the pins current; CodeQL's workflow scanning (`security-extended`) flags any future
unpinned third-party ref.


pashpashkin:
@greptileai review

GuiBibeau (approved):

PR #594, opened 2026-07-06
fix(sdp-web): bootstrap project cookie in proxy on first login

multipletwigs:
## Problem

On a fresh enter to platform.solana.org (no `sdp_selected_project_id` cookie), the dashboard fails with **"Selected project required"**:

- Server pages (`dashboard/page.tsx`, payments, custody, ...) call `createSdpApiClient()` during SSR, which throws when no project is selected.
- Client SWR fetches to `/api/dashboard/*` hit the same guard and 400.

The cookie *was* being persisted, but only client-side: `DashboardWorkspaceProvider` picks the sandbox project and calls `selectProjectAction` in a mount effect. By then the SSR pass has already thrown and the child components' fetches have already fired — a race the first paint always loses. The cookie is also `httpOnly`, so the client can't self-heal it synchronously.

## Fix

Bootstrap the cookie in the proxy (middleware) — the only layer that runs before any rendering:

- On `/dashboard(.*)`, `/api/dashboard(.*)`, `/api/playground(.*)` requests with no project cookie, resolve `/v1/projects` with the session's Clerk token, pick the default project (`default-sandbox`, else first — same rule as `reconcileProjectCookieAction`), and stamp the cookie on both the forwarded request headers (so the current SSR pass sees it) and the response (so the browser stores it).
- Requests that already carry the cookie are untouched — the extra API call only happens on the cookieless first entry.
- If the org has no projects or the lookup fails, the request proceeds cookieless and the existing route-level "Selected project required" handling surfaces the failure as before.

Supporting refactor in `lib/sdp-api.ts` (middleware can't use Clerk's request-bound `auth()` or `cookies()`):

- Extracted `acquireClerkToken(getToken)` — the `CLERK_JWT_TEMPLATE` token logic, now shared between server contexts and the proxy.
- Extracted `assembleSdpApiClient(request)` and added `createTokenSdpApiClient(token)` — builds an org-scoped client from an explicit bearer token.

The client-side persist effect in `DashboardWorkspaceProvider` and the route-handler 400 guards stay as defense in depth (they still cover the stale-cookie case).

## Testing

- `tsc --noEmit` and `biome check` clean on the changed files.
- Manual: clear the `sdp_selected_project_id` cookie, hard-reload `/dashboard` — page renders with the sandbox project selected, no 400s in the network tab.

### DIFF
--- a/apps/sdp-web/src/lib/sdp-api.ts
+++ b/apps/sdp-web/src/lib/sdp-api.ts
@@ -23,12 +23,15 @@ function getApiBaseUrl(): string {
   return base.replace(/\/$/, "");
 }
 
-async function getClerkToken(): Promise<string> {
-  const { getToken, orgId } = await auth();
-  if (!orgId) {
-    throw new Error("Active Clerk organization required");
-  }
+type ClerkGetToken = (options?: { template?: string }) => Promise<string | null>;
 
+/**
+ * Acquires the sdp-api bearer token from a Clerk `getToken`, honoring
+ * CLERK_JWT_TEMPLATE when configured. Takes `getToken` as a parameter because
+ * server contexts get it from `auth()` while the proxy middleware gets it from
+ * its `clerkMiddleware` callback.
+ */
+export async function acquireClerkToken(getToken: ClerkGetToken): Promise<string> {
   const template = process.env.CLERK_JWT_TEMPLATE;
   if (template) {
     const token = await getToken({ template });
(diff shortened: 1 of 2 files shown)


multipletwigs:
@greptileai On the two flagged items in `resolveDefaultProjectId`:

**Hardcoded `default-sandbox` slug** — this is a platform invariant, not a magic string: sdp-api's project provisioning (`apps/sdp-api/src/services/project.service.ts`) assigns exactly this slug to every org's auto-created sandbox project, and slugs aren't user-editable. It's also the same discriminator `DashboardWorkspaceProvider` and `reconcileProjectCookieAction` already match on, so the proxy is consistent with the rest of the dashboard. Documented in the function's JSDoc in 95e45b41.

**Silent catch** — intentional degradation: if the projects lookup fails, the request proceeds cookieless and the existing route-level "Selected project required" handling surfaces the failure, rather than the proxy 500ing every dashboard request. The failure isn't invisible in production — the downstream SSR throw / API 400 is what gets logged and traced today, same as before this change.

GuiBibeau (approved):

PR #637, opened 2026-07-09
fix(rpc): add per-request transport deadline to RPC client

GuiBibeau:
## Summary

- `createRpc` now builds the RPC client from a wrapped `createDefaultRpcTransport` that attaches a per-request deadline (default 30s, configurable via `RpcClientOptions.requestTimeoutMs`).
- A stalled request (socket open, no response) rejects with a transient-classified `SdpRpcError` ("RPC request timed out after Nms") instead of hanging the caller forever — polling deadlines like `confirmTransaction`'s `timeoutMs` are only checked between polls, so a single hung fetch previously defeated them.
- Callers that pass their own `abortSignal` to `.send()` are left untouched (no double-abort / rewrap).
- Uses `AbortController` + `setTimeout` (cleared on settle) rather than `AbortSignal.timeout` for consistent behavior across workerd/Node and no dangling timers.

## Why

CI evidence from the recurring Surfpool integration flake (10+ occurrences across main and PRs #623/#625/#628): the test's API server received a request and never responded for ~265s — a hung RPC fetch inside the handler. This converts that hang into a fast, retryable error.

## Validation

- `pnpm turbo typecheck` (8/8)
- `biome check` clean
- New tests in `apps/sdp-api/src/services/solana/rpc.test.ts`: stalled-fetch rejects within the deadline as transient-classified `SdpRpcError`; caller-supplied `abortSignal` is not rewrapped. Full file run 20+ times locally — new tests stable on every run.
- 16 tests passing across `rpc.test.ts` + `app.test.ts`.

## Notes

- While stabilizing this, found the pre-existing `confirmTransaction`/`sendAndConfirmTransaction` tests flake locally (~1/10) on unmodified main: the 1s poll sleep occasionally stalls under the vitest workerd pool. Verified against a pristine copy of the file from origin/main. Separate issue, not addressed here.


multipletwigs (dismissed):

keeganthomp (approved):
I liked greptile suggestion regarding `signal` as it relates to the config, but not a hard req :)

pashpashkin (approved):

PR #654, opened 2026-07-10
fix: eliminate unsafe float arithmetic on money amounts

### LINKED ISSUE #615: Batch total renders a floating-point artifact (0.15000000000000002 USDC) on the recipients step and review
**Area:** Payments / dashboard batch · **Type:** bug (display)

The batch total is summed in JS float without rounding, so 0.05 x 3 renders as `0.15000000000000002 USDC` on the Recipients-step footer and the Review header (the Result screen is clean).

**Reproduce:** Pay → Batch ONCHAIN → 3 recipients at 0.05 USDC each → observe the total.
**Impact:** an operator sees a garbage-looking total on a money screen and loses trust in the figure.
**Suggested fix:** round/format to the token decimals (or sum in integer base units) before display.

multipletwigs:
Fixes #615, then sweeps the whole repo for the same bug class.

## The reported bug

The batch-send total was summed in JS floats, so 3 × 0.05 USDC rendered as \`0.15000000000000002\` on the recipients step and review header — and could spuriously trip the exceeds-balance gate when the wallet held exactly the batch total. The total is now summed exactly with the official [\`@solana/fixed-points\`](https://github.com/anza-xyz/kit/tree/main/packages/fixed-points) package (u64 / 9 decimals, matching the amount schema), exposed as an exact decimal string, and regression-tested.

## Audit findings fixed alongside

A scan of sdp-web, sdp-api, and packages for float math on money turned up more of the same class:

**Amounts that could reach a money action**
- \`payment-requests.ts\` reconciliation passed \`Number(row.amount)\` to \`@solana/pay\`'s \`validateTransfer\` — the gate that persists a \`confirmed\` transfer. Now converted with a new round-trip-checked \`toNumberAmount\` (\`@sdp/solana/amount\`), which throws on precision loss instead of silently shifting the expected amount. \`toMosaicAmount\` now reuses it, which also fixes a latent false-failure for sub-1e-6 amounts (scientific-notation \`toString\`).
- The offramp **MAX** button round-tripped the wallet balance through \`Number()\` → \`String()\` before submitting it as the transfer amount. It now passes the exact \`uiAmount\` string through.
- The MoneyGram widget prefill (\`parseFloat\`) and the Mural/BVNK estimate request payloads (\`Number(...)\`) now use \`toNumberAmount\` at the JSON boundary.

**Client-visible estimate fields**
- MoonPay/Coinbase \`fees.total\` were float additions (the classic \`12.300000000000001\` artifact); now \`sumDecimalAmounts\`. Stripe's local fee-summing helper was folded into the same shared helper.
- \`exchangeRate\` for Stripe, Lightspark, Coinbase, and BVNK was float division; now a new BigInt-based \`divideDecimalAmounts\` (9-dp, round-half-away).
- Mural sandbox simulate-payin derived cents via \`Math.round(amount * 100)\`; now \`parseDecimalAmount(String(amount), 2)\`, which throws on >2dp input.

**Cosmetic / docs**
- \`formatLamportsAsSol\` formats from the bigint directly (SOL decimals from \`WELL_KNOWN_TOKENS\`).
- The treasury-fund tutorial taught float dividend math in a mint example; it now demonstrates integer cents/basis-points.
- The \`/^\d+(\.\d{1,9})?$/\` amount regex existed in four places; now exported once as \`ONCHAIN_AMOUNT_PATTERN\`.

## Deliberately not changed

- **UI balance gates** (\`Number(total) > Number(balance)\` in the send wizards): with the totals now exact, a single float parse per side can only misjudge >15-significant-digit boundaries; a true decimal compare would throw on tokens with more than 9 decimals. The server re-validates regardless.
- **Mural webhook \`tokenAmount\`** arrives as a JSON number and is persisted via \`String(...)\` on settlement — exact for anything ≤15 significant digits; fixing beyond that requires parsing raw body text.
- **\`computeTodaysVolume\`** sums mixed tokens/currencies into one figure — that's a unit-semantics question, not a float bug; flagging rather than guessing the intended definition.

## Notes

- \`@solana/fixed-points@7.0.0\` is new in sdp-web (added to \`minimumReleaseAgeExclude\` like the other \`@solana/*\` packages). It's the kit v7 line — self-contained, coexists with the kit 6.5 catalog pin.
- sdp-web now depends on \`@sdp/solana\` (for \`toNumberAmount\`), which required aligning its tsconfig \`target\` to the workspace's ES2022 (bigint literals).
- \`@sdp/payments\` and \`@sdp/solana\` gained \`test\` scripts (node:test + tsx, same as \`@sdp/env-config\`) with suites for the new helpers.

## Verification

- New tests: \`sumBatchAmounts\` (4), \`decimal\` helpers (5), \`amount\` helpers (4) — all passing
- \`@sdp/api\` suite 80/80; typecheck + biome clean across all four packages


multipletwigs:
@greptileai please re-review — resolved the semantic conflict with #642 (amount shim removal) and reworked the BVNK number-amount handling on @solana/fixed-points.

arseniy-nikitochkin (approved):

PR #710, opened 2026-07-14
feat(HOO-716): idempotency for POST /transfers

G1de0n:
Adds `Idempotency-Key` support to `POST /transfers` so retries are safe: a repeated request returns the original transfer and performs at most one on-chain submission.

## Behavior
- Optional `Idempotency-Key` header. With the same payload, a repeat replays the stored transfer; a different payload under the same key returns `409`.
- Keys are scoped to `(organization_id, project_id)`, so they cannot collide across projects.

## Hardening
- Race-safe claim via a partial unique index plus a unique-violation fallback, so concurrent retries still submit exactly once.
- The request fingerprint includes private-transfer options, so reusing a key with different MagicBlock options is rejected rather than silently replayed.
- Sequential retries short-circuit before MagicBlock preparation and policy enforcement, so a normal retry does no duplicate remote calls or policy/audit work.
- Confidential transfers persist their private-transfer metadata so a replay returns a consistent response.
- Fingerprint helper is shared with the existing issuance idempotency path.

## Concurrency
The partial unique index guarantees a single on-chain submission even for simultaneous duplicate requests. The early short-circuit is best-effort: a duplicate that arrives before the first row is inserted may still run policy enforcement and, for confidential transfers, a MagicBlock preparation call before its insert is rejected by the unique constraint. This can produce duplicate audit records and one extra preparation call, but never a second on-chain submission.

## Tests
Repository, route, and unit coverage for replay, `409` on payload mismatch, project isolation, confidential replay, policy enforcement running only once on sequential retry, and failed-transfer replay.


G1de0n:
@greptile

Thanks for the review.

On the failed-transfer key retention: this is intentional and matches the existing issuance idempotency path. The key is claimed at record creation (`status=processing`) before submission, so a concurrent or retried request replays instead of double-submitting — including the case where a transient error surfaces after the transaction was already sent on-chain. Replaying the failed row (`200` with `status=failed`) rather than allowing a same-key retry is what avoids that double-submit risk; a genuinely new attempt uses a new key. Added a characterization test (`replays a failed transfer on retry without submitting again`) to lock this in.

Both P2 inline suggestions are applied: the call sites now reuse the captured `idempotencyKey`, and the standard replay path uses `buildTransferReplayPayload` for consistency.

GuiBibeau (approved):

Issue #673, opened 2026-07-13
Payments transfers ignore the `Idempotency-Key` header: retries double-pay (issuance is idempotent, payments is not)


resourcefulmind:
**Area:** Payments / transfers + batches · **Type:** bug (money-safety)

`POST /v1/payments/transfers` does not honor the `Idempotency-Key` header. Two identical POSTs with the same key produce two separate on-chain transfers. Issuance honors idempotency (`buildIdempotencyMetadata` reads the header and fingerprints the op, e.g. `burn.ts`); the payments transfer and batch handlers have **zero** idempotency handling.

**Reproduce** (clean sequential pair, 2026-07-09)
```
POST /v1/payments/transfers  (Idempotency-Key: K, dest=D, 0.013 USDC)  -> 200 xfr_538473ad slot 475068289
POST /v1/payments/transfers  (SAME Idempotency-Key K, identical body)  -> 200 xfr_195ab44a slot 475068304
GET  /v1/payments/transfers  -> two 0.013 transfers, different ids + slots = money moved twice
```

**Impact:** a retry, double-click, or at-least-once job runner double-pays on the highest-stakes endpoint. Batches share the same gap (code-confirmed).
**Suggested fix:** apply the issuance idempotency path (`buildIdempotencyMetadata` / `Idempotency-Key` fingerprint) to `/transfers` and `/transfer-batches` so a same-key retry returns the original transfer.

multipletwigs:
Closing as fixed: `/transfers` honors `Idempotency-Key` since PR #710 and `/transfer-batches` since PR #751 (migration 0026). Both fingerprint via `src/lib/idempotency.ts` and replay on match; regression tests live in `routes/payments.test.ts` ("replays a transfer when the same Idempotency-Key + body is retried") and `routes/payments/transfer-batches.test.ts`.

PR #751, opened 2026-07-17
fix(payments): honor Idempotency-Key for POST /transfer-batches 

multipletwigs:
Retrying a transfer batch with the same Idempotency-Key now returns the original batch instead of paying every recipient again (#673).

- idempotency_key/idempotency_fingerprint on payment_transfer_batches with a partial unique index scoped to (organization_id, project_id, key).
- Request fingerprint over the resolved source, token, ordered recipients, and options; reusing a key with a different payload returns 409.
- Race-safe claim: pre-check plus unique-violation fallback at batch insert, which lands before any on-chain submission.
- Replay/conflict resolution shared with POST /transfers via resolveIdempotencyReplay in lib/idempotency.
- OpenAPI: both POST endpoints document the header and the 409.


multipletwigs:
@greptileai please re-review — since the last review: batch + recipient inserts now run in a single transaction (concurrent idempotency-replay readers can no longer observe a batch without its recipients), a stored null fingerprint is treated as unclaimed instead of a 409, and the Postman collection was regenerated for the updated endpoint descriptions.

G1de0n (approved):

Issue #674, opened 2026-07-13
Batch transfers are not screened against the compliance providers SDP already integrates


resourcefulmind:
**Correcting my own report first.** The original title claimed batch transfers bypass the wallet destination allowlist. That was wrong when I filed it, and it is still wrong. `transfer-batches/create.ts` maps every recipient into its own policy leg carrying `destination: recipient.destinationAddress`, so policy evaluates each destination individually. Resolved destinations are also pinned into the approved payload, with the stated intent that "a counterparty account whose address changes between approval and replay fails the replay match instead of executing to a destination policy never saw." That half is closed. Please do not let it discredit the half below.

## The surviving gap: SDP has address screening and nothing on the money path calls it

`apps/sdp-api/src/services/compliance/` implements screening across four providers, `chainalysis.ts`, `elliptic.ts`, `range.ts`, `trm.ts`, exposed as `POST /v1/compliance/address-screenings` and described in the OpenAPI spec as running "address screening against configured compliance providers".

That service is imported by exactly one file: `apps/sdp-api/src/routes/compliance/handlers.ts`, its own route handler. No transfer, batch, or payments code path invokes it.

## Failure scenario

An organisation sends a batch payout to 20 recipients. Every destination is checked against the wallet policy, and none is checked against Chainalysis, Elliptic, Range or TRM, even though all four are wired up and reachable. Screening is available only if the caller remembers to call a separate endpoint first and act on the result themselves.

## Why this is not "working as designed"

The compliance providers were integrated deliberately and are production code, not a spike. Nothing in the transfer path opts out of them explicitly; there is simply no call site. If the intended contract is that screening is caller-driven, that should be stated in the docs and the batch guide, because a reasonable integrator will assume a platform that ships four screening providers screens the payments it sends.

## What I am asking for

Either wire screening into the transfer and batch paths behind whatever configuration is appropriate, or document explicitly that screening is opt-in and caller-invoked.

enochakinbode:
gm I'm interested in contributing to this project. I noticed this issue is already assigned, so I won't start working on it. If it becomes available or if there are similar issues that could use help, I'd be happy to take one on. Thanks!

multipletwigs:
@enochakinbode feel free to take this one on if you're interested :)

multipletwigs:
also happy to help out, and gather feedback about local dev

enochakinbode:
@multipletwigs  alright, right on it, i will let you know about my experience.

[...]

PR #777, opened 2026-07-18
test(api): guard batch transfer destination-allowlist enforcement

multipletwigs:
Regression guard for the first half of #674.

Investigation showed the claimed bypass does not exist: `enforceBatchPolicies` (transfer-batches.ts:601-611) runs every recipient through `assertWalletPolicyAllowsTransfer` with the allowlist enforced, before any DB write — a violating recipient 403s the whole batch with zero rows persisted. The `enforceDestinationAllowlist: false` call the issue pointed at is the separate aggregate amount check (no single destination). But nothing tested this, so a regression would ship silently.

Adds two integration tests (real Postgres): a 2-recipient batch with one non-allowlisted destination → 403 FORBIDDEN + zero batch/recipient rows; an all-allowlisted batch still completes. 14/14 file tests green; tsc + biome clean.

The second half of #674 (server-side sanctions screening in the transfer path) remains open — it needs latency/fail-mode/cost decisions.


multipletwigs:
@greptileai re-review please — follow-up commits address the previous findings.

eldarik (approved):

PR #780, opened 2026-07-18
fix(api): unify list-endpoint pagination envelopes and paginate /wallets

multipletwigs:
Fixes #676.

- `GET /v1/counterparties` and `GET /v1/counterparties/accounts` move from hand-rolled `{counterparties/accounts, total, page, pageSize}` shapes to the canonical `paginated()` envelope (`{data, meta: {total, page, pageSize, hasMore, requestId}}`) used by transfers/issuance.
- `GET /v1/wallets` gains `page`/`pageSize` (defaults and max copied from `listTransfersQuerySchema`); previously it returned every wallet with no pagination metadata. Pagination slices the cached wallet summary set in-memory (the cache stores the full filtered set — not redesigned), and balances are only fetched for the returned page.
- **Breaking response shapes** (pre-mainnet): dead envelope types deleted from `@sdp/types`; all sdp-web consumers, OpenAPI schemas, playground configs, tutorials, and generated artifacts updated.
- Not touched: nested `GET /v1/counterparties/{id}/accounts` shares the legacy shape but wasn't named in the issue — follow-up candidate.

## Testing
Multi-page integration tests (real Postgres: seed 3 rows, `pageSize=2`, assert `meta.total`/`hasMore` + page-2 remainder) for `/v1/wallets`, `/v1/counterparties`, and `/v1/counterparties/accounts`. Full sdp-api suite: 101 files / 1048 passed, 0 failed. `tsc --noEmit` clean in sdp-api, sdp-web, sdp-docs.

## Merge order
Conflicts with #778 on `llms-full.txt` + two tutorial MDX files — whichever merges second needs a rebase and a `pnpm generate:ai` re-run (regenerate, don't hand-resolve).

PR #781, opened 2026-07-18
fix(payments): persist and surface counterparty on direct transfers

### LINKED ISSUE #561: The counterparty is not shown on transactions, and for direct payments it is not even recorded
**Area:** Payments / transactions 

The transactions list shows "Unavailable" for the counterparty in two distinct situations:

- **Ramp transfers** carry a `counterpartyId`, but the list does not resolve and display it.
- **Direct payments** do not persist the counterparty at all. Paying a counterparty through the dashboard Pay flow records `counterpartyId: null`, because the send request sends only `source`, `destination`, `token`, `amount`, and `memo`.

**Reproduce**
```bash
# after paying a selected counterparty through the dashboard Pay flow:
curl "$BASE/v1/payments/transfers?limit=1" -H "Authorization: Bearer $KEY"
#   the record has counterpartyId: null
```

**Impact:** the transactions list is the reconciliation surface. A payment that shows no counterparty looks orphaned, and for direct payments the link to the payee is lost permanently.
**Suggested fix:** resolve and render the counterparty for ramp rows, and persist `counterpartyId` on direct transfers.

### LINKED ISSUE #588: Direct transfers cannot record a counterparty: createTransferSchema has no counterpartyId field
**Narrowing this.** Of the four original asks, two have shipped and one is fixed on the surface this issue named. What remains is one write-time defect.

## Shipped, closing these out

- **Counterparty name is shown.** `transactions-results.tsx:299`, served by the join at `payments.repository.postgres.ts:520` (`c.display_name AS counterparty_display_name`).
- **Transaction type is shown.** `transactions-results.tsx:74` via `resolveTransferTypeLabel`, so `transfer` and `offramp` are no longer both rendered as "Outbound".
- **Token symbols resolve in the transactions list**, including devnet mints, via `payments-overview.utils.ts:81-106` and the catalogue at `well-known-tokens.ts:132`. **Correcting my own earlier comment on this issue:** I said the dashboard inherits the API's raw mint. It does not; the list resolves symbols itself. The remaining token inconsistency is API-surface-only, where `mappers.ts:40` returns `row.token` verbatim so ramp rows carry a symbol and direct transfers carry a raw mint.

## The surviving defect, stated precisely

A transfer to an arbitrary address has no counterparty and should show the address. That is correct behaviour and is not what this is about.

The defect is that a transfer sent **to a saved counterparty from the dashboard** also loses the link. `use-onchain-send-wizard.ts` takes `counterpartyId` (`:62`), fetches that counterparty's accounts (`:99`), and derives the destination from the selected account (`:122`, `resolveAccountAddress(selectedAccount)`). It then calls `createTransfer` with `destination` and no counterparty (`:194`). The flow knows exactly who is being paid and drops it at the last step.

The API cannot carry it either: `createTransferSchema` (`routes/payments/schemas.ts:511-519`) has no `counterpartyId` field, and `handlers/transfers.ts:197` hardcodes `counterpartyId: null` inside `createTransferRecord`, whose input type has no such field.

## What it looks like in the product

In the transactions list, Pay and Deposit rows show a counterparty name because those paths persist the id. Transfer rows show a bare address, including transfers that were sent to a saved counterparty through the dashboard send flow. Those payments also cannot be filtered by counterparty, since the filter reads `counterparty_id`.

## This is not a platform limitation

`counterpartyId` is persisted by payment requests (`services/payments/payment-requests.ts:171`), by recurring payments (`services/payments/recurring-payments/collection.ts:1042`), and by single-recipient batches (`transfer-batches/execute.ts:189`). The direct transfer path is the only outlier. The database, repository and read path all support it already.

## Please do not close this by pointing at #561

#561 was closed COMPLETED on 2026-07-28, but the PR that would have finished it, #781 `fix(payments): persist and surface counterparty on direct transfers`, whose body reads "Fixes #588, fixes #561", was **closed unmerged** on 2026-07-21. Its display half shipped separately in #801. The persistence half has never merged.

## Why it matters more than a display bug

This is write-time data loss. Every such transfer sent between filing and the fix has no counterparty recorded, and no later change can backfill what was never written.

multipletwigs:
Fixes #588, fixes #561.

Direct transfers silently dropped the counterparty: the create schema had no `counterpartyId`, the handler never stored it (despite full DB/repository support used by batches/recurring), and the response mapper only exposed it for ramp types. The dashboard sent it nowhere and rendered raw addresses.

## API
- `createTransferSchema` accepts optional `counterpartyId`; threaded create-handler → `createTransferRecord` → repository (no migration needed).
- `mapTransferRow` exposes `counterpartyId` for all transfer types, not just ramps; OpenAPI schemas updated.
- `Idempotency-Key` fingerprint includes `counterpartyId` — **back-compat verified**: requests without it fingerprint byte-identically to pre-change (undefined keys are stripped before hashing), so stored fingerprints still replay.
- POST /transfers now returns **201** (was 200) via the shared `created()` helper, consistent with every other create endpoint — flagging as a minor breaking change.

## Dashboard
- Send flows pass the selected counterparty through `createTransfer`.
- Transactions list resolves `counterpartyId` → name (full counterparty set via the existing paginated `fetchAllCounterparties` helper, gated render until loaded) and adds the type column via the existing-but-unwired `resolveTransferTypeLabel`.

## Testing
- Integration (real Postgres): POST with `counterpartyId` → 201 exposes it → GET by id and filtered list round-trip it; idempotency fingerprint unit tests.
- Full sdp-api workers suite: 101 files / 1047 passed, 0 failed. `tsc --noEmit` clean in both apps; biome clean on all 18 changed files.

## Merge order
Touches `payments-workspace.data.ts` alongside #780 (pagination envelopes) — whichever merges second needs a small rebase.

Implemented by Codex, then independently verified and revised (removed a non-conventional web error-factory module; fixed a first-100-counterparties lookup cliff that could crash the transactions view for large orgs).


multipletwigs:
@greptileai re-review — both P1s addressed: the counterparty name map now fetches with `includeArchived: true` (soft-deleted counterparties resolve to names; the send-flow pickers keep the active-only default), and `PaymentTransferSummary.type` is now required (verified NOT NULL in DDL + unconditional in `mapTransferRow` + non-optional in OpenAPI — the optionality was stale), deleting the throw. Refresh button revalidates the counterparties fetch. One note: `payments-page.data.ts`'s defensive JSON re-parse follows its own pre-existing default-per-field convention (`type ?? \"transfer\"`, like `id ?? \"\"` above it) — converting that whole function to strict parsing is a separate cleanup.

multipletwigs:
Re the two remaining review notes — both are intended per this repo's fail-loudly convention, documenting rather than changing:

1. **`resolveCounterparty` throw**: with `includeArchived: true` the name map now covers the full project-scoped counterparty set (active + archived; the platform never hard-deletes). A counterpartyId missing from that set is referential corruption, and silently rendering a raw ID would mask it in a payments reconciliation surface. The previous crash risk was real only because the map was incomplete — that's what got fixed.
2. **Fetch-failure state**: an explicit error panel + the page's standard Refresh retry (which revalidates the counterparties fetch) is preferred here over partially rendering unlabeled rows. Degrading to name-less rows is exactly the silent-fallback rendering this codebase bans.

Final call is the maintainer's on review — flagging that these were deliberate, not oversights.

Issue #782, opened 2026-07-18
Boolean query params via z.coerce.boolean() parse "false" as true


multipletwigs:
Found while wiring `includeArchived` on `GET /v1/counterparties` (PR #781): `listCounterpartiesQuerySchema` uses `z.coerce.boolean().default(false)`, and `Boolean("false") === true`, so a client sending `?includeArchived=false` gets archived rows included. Any other boolean query param using `z.coerce.boolean()` has the same footgun (grep for it across route schemas).

Fix: use a string-literal enum (`z.enum(["true","false"]).transform(...)`) or `z.stringbool()` (zod 4) for boolean query params. PR #781 sidesteps it client-side by only appending the param when true.

PR #789, opened 2026-07-18
fix: resolve boolean query param parsing and allow cancelling pending_activation recurring payments

DHRUVILPATEL2303:
### Summary
This PR addresses two separate issues in `sdp-api`:
1. **Boolean Query Param Validation**: Previously, route query schemas utilizing `z.coerce.boolean()` parsed `"false"` string query parameters to `true` (since `Boolean("false") === true`). This was causing routes like `GET /v1/counterparties?includeArchived=false` to mistakenly return archived records. A custom `queryBooleanSchema` helper has been added and integrated to properly preprocess string booleans.
2. **`pending_activation` Subscription Cancellation**: Recurring payments stuck in `pending_activation` state were throwing conflict errors when trying to cancel. We now allow these subscriptions to be canceled directly in the database, bypassing on-chain interaction.

### Changes
* Created `queryBooleanSchema` helper in `routes/payments/schemas.ts`.
* Updated query schemas for `counterparties`, `counterparty-accounts`, and `asset-profiles` to utilize `queryBooleanSchema` for `includeArchived`.
* Updated `cancelRecurringPayment` in `services/payments/recurring-payments.ts` to update database state directly to `"canceled"` if `pending_activation`.
* Added integration test verifying direct cancellation of `pending_activation` recurring payments in `routes/payments.test.ts`.

### Verification
* Ran typecheck (`pnpm typecheck`): **Passed**
* Ran module boundary checks (`pnpm check:module-boundaries`): **Passed**
* Integration test case (`cancels pending_activation recurring payments directly without on-chain tx`): **Passed**


multipletwigs:
@DHRUVILPATEL2303 this should be to close https://github.com/solana-foundation/solana-developer-platform/issues/782 right? will get to it in a bit

multipletwigs:
@DHRUVILPATEL2303 I think [`z.stringbool()`](https://zod.dev/codecs?id=stringbool) might be a more idiomatic way to do it

```ts
export const queryBooleanSchema = z.stringbool({ truthy: ["true"], falsy: ["false"] });
```

DHRUVILPATEL2303:
@multipletwigs Done! Refactored the `queryBooleanSchema` in `common-schemas.ts` to use `z.stringbool({ truthy: ["true", "1"], falsy: ["false", "0"] })` to make it more idiomatic in Zod 4, while preserving compatibility for query params using `1` and `0` and adding case-insensitivity.

DHRUVILPATEL2303:
> @DHRUVILPATEL2303 this should be to close #782 right? will get to it in a bit

Yes, that's correct! This is to close #782.

koriyoshi2041:
Could we add focused coverage for the shared query boolean schema before merging? The current new test only exercises pending-activation cancellation, so the original regression is not protected. A small table test covering false and 0 as false, true and 1 as true, plus one rejected value would protect all three includeArchived routes from drifting back to truthy string coercion.

DHRUVILPATEL2303:
> Could we add focused coverage for the shared query boolean schema before merging? The current new test only exercises pending-activation cancellation, so the original regression is not protected. A small table test covering false and 0 as false, true and 1 as true, plus one rejected value would protect all three includeArchived routes from drifting back to truthy string coercion.

@koriyoshi2041 Done! Added focused unit test coverage in `apps/sdp-api/src/routes/common-schemas.test.ts`.

It includes:
- Table-driven unit tests for `queryBooleanSchema` verifying `"false"` & `"0"` parse to `false`, `"true"` & `"1"` parse to `true`, and invalid string values are rejected.
- Schema integration tests for `includeArchived` across all three affected route schemas (`listCounterpartiesQuerySchema`, `listCounterpartyAccountsQuerySchema`, and `listAssetProfilesQuerySchema`) to guarantee none drift back to truthy string coercion.

koriyoshi2041 (approved):
Re-reviewed at `5df4436`. The focused table coverage now protects both direct parsing and all three `includeArchived` route schemas for `false`/`0`, `true`/`1`, and invalid input. This resolves my coverage concern; the remaining Vercel failures are authorization-only.

DHRUVILPATEL2303:
@multipletwigs @koriyoshi2041

multipletwigs:
~~hey @DHRUVILPATEL2303 thanks for helping out, could we fix up the conflict and also limit the boolean to just "true" and "false"? Test cases is good to go.~~ Approved, i've fixed up the conflicts.

multipletwigs (approved):

eldarik (approved):

DHRUVILPATEL2303:
@multipletwigs @eldarik all checks are passed . you can merge it

PR #852, opened 2026-07-22
fix(tests): standardize test runs on TEST_DATABASE_URL

multipletwigs:
- Integration test runs used to resolve `DATABASE_URL` (falling back to the local dev `sdp` database) and wipe it during suite cleanup — running integration tests locally destroyed your dev data.
- Every test entry point (`run-workspace-tests.mjs`, `kora-surfpool/integration.sh`, integration vitest config, `env.node.ts`, `kora:devnet:test`) now reads only `TEST_DATABASE_URL`, defaulting to `sdp_test`; `db:migrate:test` creates and migrates that database automatically.
- The integration vitest config now hard-assigns `DATABASE_URL` from `TEST_DATABASE_URL` instead of `??=`, so an ambient/`.env.local` `DATABASE_URL` can never leak into the suite.
- Removed the derive-`_test`-from-`DATABASE_URL` arm in `migrate-postgres-test.mjs` and a duplicate `db:migrate:test` run in unit mode.
- CI unit + integration-shard jobs now pass `TEST_DATABASE_URL` (and preserve it through Doppler) instead of `DATABASE_URL`; browser e2e keeps `DATABASE_URL` since it runs the live API, not the test suite.
- Verified: full `@sdp/api` suite green locally (117 files / 1212 tests).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,9 +132,9 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CI_HAS_DOPPLER_TOKEN: ${{ secrets.DOPPLER_TOKEN_CI != '' }}
-      DATABASE_URL: postgresql://sdp:sdp@127.0.0.1:5432/sdp
+      TEST_DATABASE_URL: postgresql://sdp:sdp@127.0.0.1:5432/sdp_test
       REDIS_URL: redis://127.0.0.1:6379
-      DOPPLER_PRESERVE_ENV: DATABASE_URL,REDIS_URL
+      DOPPLER_PRESERVE_ENV: TEST_DATABASE_URL,REDIS_URL
     services:
       postgres:
         image: postgres:16-alpine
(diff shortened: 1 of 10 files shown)


multipletwigs:
Not reproducible — unit tests never touch `sdp_test`. `node-global-setup.ts` boots a Testcontainers Postgres, migrates it, and overwrites `process.env.TEST_DATABASE_URL` with the container URI before workers spawn, so the ambient value is shadowed. Two proofs: the CI Unit Tests job is green on a Postgres that has no `sdp_test` at all, and running `signing.service.reuse.test.ts` locally with `TEST_DATABASE_URL` pointed at a nonexistent database still passes. `db:migrate:test` in unit mode would migrate a database nothing connects to; `sdp_test` is integration-only and integration mode migrates it.

multipletwigs:
@greptile-apps let me know what you think of the pr above.

G1de0n (approved):

PR #866, opened 2026-07-22
feat: asset management ui and issuance workspace ui improvements

arseniy-nikitochkin:
## Summary

Phase 6 of the Issuance v2: a per-policy compliance configuration layer (trading hours, approval rules, investor reporting), type-aware asset detail fields with editable symbol/decimals pre-deploy, a per-asset audit/activity feed, and a redesigned issuance token list (grid/list toggle, filters, search). Bundled with a pass of UI polish on the asset-management tabs (compliance controls, operations, wallet identity) and several accessibility fixes.

## Compliance policy configuration

- New **capacity config modal** (`capacity-config-modal.tsx`) for the four configurable compliance capacities — trading-hours schedule (24/7, market hours, or custom days/times/timezone), transfer & redemption approval rules (all / above-amount / new-counterparty, with threshold + approver roster), and investor reporting (cadence, format, recipients). `kyc` and `issueRetireControls` stay plain toggles with no config.
- Capacity state changed shape from `Record<CapacityKey, boolean>` to `Record<CapacityKey, { enabled, config? }>`, separating "policy is on" from "how it's configured." `coerceCapacities()` normalizes legacy bare-boolean drafts (wizard localStorage + DB hydration) so old persisted data doesn't crash the new UI — worth a close look since it's the only thing standing between old data and a render error.
- `advanced-settings-editor.tsx` reworked (600+ line diff): manager-facing vs. technical-mode copy, a deploy-config preview (technical mode only), locked-row treatment for required/deployed settings (padlock + non-interactive row instead of a disabled checkbox), a `trailing` slot so a row's "Configure" action doesn't sit inside the toggle's clickable label, and container-query-responsive grids (`containerResponsive` prop) so the compliance tab's narrower two-column layout doesn't rely on viewport breakpoints.
- Once a token is deployed, the settings editor now hides extensions that were never selected at deploy (`visiblePermanent`) instead of showing dead, permanently-off rows.
- Replaced a custom `role="tablist"`/`aria-selected` button group with a new **`SegmentedControl`** built on native `<input type="radio">` — the old markup implied tab/tabpanel semantics it didn't have; the new one gets correct AT announcements and keyboard roving focus for free.
- New **`TimeField`** component (two themed `Select` dropdowns for hour/minute) replaces native `<input type="time">` in the trading-hours form — the native time picker's selection highlight can't be restyled to match the SDP palette.
- Compliance capacity copy went through several passes to drop invented/second-person wording and keep crypto/token jargon exclusively in technical-mode strings.

## Type-aware asset fields & editable symbol/decimals

- Asset detail fields are now chosen by `(category, type)` instead of category alone — e.g. crypto-backed vs. fiat-backed stablecoins get distinct field sets (collateral ratio/oracle/liquidation threshold vs. reserve/custodian), tokenized securities split into equity/debt/fund-specific terms. Asset-type registry bumped v2 → v3.
- **Symbol and decimals are now editable while a token is an undeployed draft**, both in the create wizard and the Details tab of asset management — locking automatically once deployed (mint address present). Enforced end-to-end: `updateTokenSchema`/`TokenService.updateToken` on the API reject `symbol`/`decimals` after deployment (400), with route + service tests covering both the allowed and rejected paths.
- New shared `issuance-token-fields.tsx` reuses the same field-resolution engine as the create wizard so the token list/grid views render consistent type-aware fields without duplicating logic.

## Audit trail / activity feed

- New `GET /v1/issuance/tokens/:tokenId/audit` endpoint and `AuditService.getForAsset`/`countForAsset`, aggregating both directly-logged rows and child events (transactions, allowlist entries, frozen accounts) that reference the token via `metadata.tokenId`.
- Filtering extended end-to-end (API → BFF route → UI) to **action, status, and actor type** (`user`/`api_key`/`system`, derived from which actor id is set — no stored column). Fully unit-tested, including combined-filter parameter ordering.
- **Bug fix bundled in**: `AuditService.log()` previously resolved the actor/org only from API-key auth context, so dashboard-driven (non-API-key) actions were written with a null `organization_id`/`user_id` and were invisible to org-scoped queries. It now also reads the Clerk/session context. No backfill for historical rows already written with a null org/user id — flag for follow-up if that gap matters.
- New `activity-tab.tsx` (filterable/paginated activity table) and `permissions-tab.tsx` (permissions/extensions section extracted out of the overview tab, including an "external authority" warning banner).
- System-actor label changed from "Automated" to "SDP".

## Issuance token list & workspace

- New `issuance-token-list.tsx`: collapsible list view where rows expand in place (via a reworked `HeightReveal`) into a type-aware field card, alongside the existing grid view.
- New `IssuanceFilterPopover`: icon-only trigger opening status/template/created-date filters plus a sort selector, client-side, matching the existing free-text search. Grid/list toggle simplified from a two-button segmented control to a single icon-only button showing the view it switches *to*.
- `dashboard/issuance/(overview)/page.tsx` now carries fields the API already returned but the page previously dropped (decimals, maxSupply, mint/freeze flags, description, uri, signingWalletId), plus a best-effort `/v1/issuance/asset-profiles` fetch merged by tokenId (soft-fails to core fields if the feature flag is off).
- Toolbar restacks to two rows below `sm` instead of wrapping awkwardly; search input border thinned to 1px to match the filter/toggle buttons.

## Asset-management tab polish

- **Operations tab**: redesigned as a section header + side-by-side action cards with circular icon badges (was a plain list).
- **Compliance tab**: all form inputs (wallet address, reason, allowlist/denylist) thinned to a 1px border to match surrounding buttons/cards; pause control's button right-aligned; allowlist/denylist rebuilt with a search bar + label-filter dropdown on one row and an icon-based empty state; empty-list message now sits inline with the add button.
- Allowlist page size bumped 100 → 500 (both the BFF request and the API's cap) so the new client-side search/filter covers the full list without a dedicated server-search endpoint (~1–1.5 day estimated effort, deferred).
- **Wallet identity card**: address value font reduced to 12px (`text-xs`) for both Wallet ID and Public Key rows.
- **Action tab bar** (`action-pills.tsx`): rebuilt from a pill button group into an underlined tab bar with per-action icons (allowlist/seize/force-burn/freeze/pause).
- `TokenActionCard`/`TokenActionForms`/`OpsActionForms` gained a `variant`/`formVariant` prop (`card` / `flat` / `bare`) so the same action forms can render with legacy card chrome, the new flat asset-profiles chrome, or no chrome at all when an ancestor already supplies the frame (e.g. the compliance tab's single wrapping card) — avoids a card-in-card.

## Accessibility & smaller fixes

- New `useInlineValidationMessage` hook: suppresses the browser's native validation bubble and renders the message as styled inline text instead, while still relying on native constraint validation to block submission (no `noValidate`). Wired into the mint/burn destination fields and the wallet address field, with correct `aria-describedby`/focus handling.
- `WalletProviderMark` now accepts `provider?: KnownCustodyProvider | null` (was required) so it renders a neutral fallback for wallets with no known provider (e.g. raw external addresses) instead of requiring a caller-side branch.
- Dashboard shell's sidebar-collapse breakpoint bumped from `lg` to `xl`, keeping the full sidebar visible on more viewport widths before collapsing to the mobile overlay.
- `dashboard-workspace-context.tsx`: fixed a stale-tab bug where App Router `<Link>` navigation (no popstate) let an explicit deep link like `?tab=playground` get wiped by a stale snapshot.
- `dashboard-workspace-panel.tsx`: added `scrollbar-gutter: stable` so an expanding list row summoning a scrollbar doesn't shift column width.
- Playground now accepts a `?tokenId=` deep link and preselects it in the request fields; new `playground-links.ts` builds these shareable URLs from asset-management admin actions.
- Removed the `SettingSigning` concept (`custodial-only` / `custodial-or-wallet`) entirely from the advanced-settings types and generated support matrix — small breaking change to that shape if anything external consumed it.

## Testing

- New/updated unit tests: `issuance-filter-popover.unit.test.ts`, `time-field.unit.test.tsx`, `advanced-settings-editor.unit.test.tsx`, `draft-mapping.unit.test.ts`, `asset-details-config.unit.test.ts`, `setting-combos.unit.test.ts`, `capabilities.test.ts`.
- New/updated API tests: `issuance.test.ts` (symbol/decimals lock behavior), `audit.service.test.ts` (status/actorType filters, combined-filter parameter order).

image


arseniy-nikitochkin:
@greptileai review pls

PR #873, opened 2026-07-23
feat: support business counterparties on Lightspark on-ramp

multipletwigs:
## Summary

Grid supports both ramp directions for `BUSINESS` customers once KYB is approved — verified live against the Grid sandbox. The gate is verification status, not ramp direction: business customers are created `UNVERIFIED` and get `CUSTOMER_NOT_VERIFIED` on quotes in **both** directions until `POST /verifications` approves them (sandbox approves synchronously; individuals are auto-approved on create).

This removes the hard-coded "business on-ramp unsupported" gate and wires KYB verification into business customer provisioning — which also fixes the existing business off-ramp path, since it never submitted verification and would hit `CUSTOMER_NOT_VERIFIED` against today's Grid API.

## Changes

- `lightsparkCounterpartyRequirements`: business on-ramp now returns `collect` with the businessInfo fields (`businessLegalName`, `businessTaxId`, `businessIncorporatedOn`) until a Grid customer exists, then `ready` — same JIT pattern the off-ramp already used
- `LightsparkRampClient.submitVerification`: `POST /verifications`; production returns the outstanding KYB requirements in `errors`
- `ensureLightsparkCustomer`: submits KYB after business customer creation and persists the customer id **only after approval**, so a pending verification re-runs on the next requirements advance (creation is idempotent by `platformCustomerId`); throws with Grid's outstanding requirements when not approved
- Requirements advance now threads `collectedData` on the lightspark on-ramp arm (schema + handler)
- Declared on-ramp `entityTypes` now include `business`; ramp-support snapshot regenerated

## Live verification (Grid sandbox)

- Unverified BUSINESS customer → on-ramp and off-ramp quotes both fail `CUSTOMER_NOT_VERIFIED`
- After `POST /verifications` (auto-approves in sandbox) → both quote directions succeed with locked rates and payment instructions
- INDIVIDUAL customers are auto-approved on create — existing individual flows unaffected

## Notes / deferred

- Production KYB needs far more than the 3 collected fields (registration number, business type, address, beneficial owners, 4 document types). Sandbox auto-approves, so the non-approved path is unreachable today; when production KYB is built, the throw should graduate to typed `customer_verifying` / `customer_verification_failed` requirements statuses like BVNK/Mural (plus `kybStatus` cached in provider_data)
- Left the 409-then-lookup retry path in `getOrCreateCustomer` as-is; it only costs an extra call on explicit user retry after a failed verification

## Test plan

- `vitest` lightspark provider + counterparties + payments route suites pass (236 tests)
- `tsc` clean on sdp-api and sdp-payments; biome clean

### DIFF
--- a/apps/sdp-api/src/lib/ramps/providers/lightspark/client.test.ts
+++ b/apps/sdp-api/src/lib/ramps/providers/lightspark/client.test.ts
@@ -429,6 +429,25 @@ describe("LightsparkRampClient", () => {
     ]);
   });
 
+  it("treats a replayed verification for an approved customer as approved", async () => {
+    vi.spyOn(globalThis, "fetch").mockResolvedValueOnce(
+      new Response(
+        JSON.stringify({
+          code: "INVALID_INPUT",
+          reason: "Customer KYC status is already APPROVED",
+        }),
+        { status: 400, headers: { "Content-Type": "application/json" } }
+      )
+    );
+
+    const client = new LightsparkRampClient();
+    const verification = await client.submitVerification(LIGHTSPARK_CONTEXT, {
+      customerId: "Customer:cus_123",
+    });
+
+    expect(verification).toEqual({ verificationStatus: "APPROVED", errors: [] });
+  });
+
   it("derives content-addressed payout account keys", async () => {
     const key = await lightsparkPayoutAccountKey("USD", {
       paymentRails: "ACH",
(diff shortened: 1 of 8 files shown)


multipletwigs:
@greptileai Re the `errors` null-guard finding: `errors` is a **required** field on the `Verification` schema in the Grid OpenAPI spec (see `components.schemas.Verification.required` — `[id, customerId, verificationStatus, errors, createdAt]`), with the description "List of issues preventing verification from proceeding. Empty when verificationStatus is APPROVED or IN_PROGRESS." — i.e. it's an empty array in non-error states, never absent. Spec: https://github.com/lightsparkdev/grid-api/blob/main/openapi.yaml

Making the field optional and adding `?? []` would defend against a response that violates the provider's own contract, and this codebase deliberately avoids fallbacks on contractually-present data — if Grid ever omitted the field, we'd rather fail loudly than emit an error message with the outstanding-requirements list silently empty.

Leaving the code as-is.

arseniy-nikitochkin (approved):

PR #884, opened 2026-07-23
feat: HOO-769 Submit stored privy credentials

pashpashkin:
## What changed

- Added `POST /internal/dashboard/custody/provider-credentials` for submitting customer-owned Privy credentials.
- Added Clerk authentication, Project membership checks, `Idempotency-Key` replay, Organization entitlement checks, and the default-off `PRIVY_BYOK_PROVISIONING_ENABLED` gate.
- Added first installation, failed-initial-setup replacement, and post-deactivation reinstall flows using immutable Provider Credential versions and Project-scoped Custody Connections.
- Credentials are stored through `CredentialSecretStore`:
  - GCP Secret Manager for managed deployments.
  - Encrypted database storage for self-hosted deployments.
- Added backend-specific cleanup, safe auditing, orphan-risk alerts, and redacted responses.
- Updated fresh legacy Privy admission to prevent conflicting legacy and stored-credential setup.
- Added a dedicated `CREDENTIAL_FINGERPRINT_PEPPER` for HMAC-based submission fingerprints

## Why

This enables self-service Privy BYOK credential submission without requiring deployment-level `PRIVY_APP_ID` and `PRIVY_APP_SECRET`.

Submission creates pending resources only. Provider validation, wallet creation, and Connection activation remain separate operations owned by the next task.

## Impact

- Adds an internal dashboard API; no public API contract is changed.
- Adds nullable idempotency columns and an Organization-scoped unique index to `provider_credentials`.
- Adds `PRIVY_BYOK_PROVISIONING_ENABLED`, disabled by default.
- Does not call Privy, create wallets, change custody defaults, or migrate existing `custody_configs`.
- Concurrent or failed writes do not leave duplicate domain resources. Ambiguous GCP writes or failed cleanup emit a redacted operational alert.
- Adds `CREDENTIAL_FINGERPRINT_PEPPER`, required when Privy BYOK submissions are enabled. A missing or blank value fails closed before any secret or domain data is written.
- The value must be shared across all API instances and remain stable; changing it causes previously committed `Idempotency-Key` replays to conflict.
- Adds `CREDENTIAL_FINGERPRINT_PEPPER`. The API can start without it, but Privy BYOK credential submission fails closed with `500 Internal Server Error` before any secret or domain write when the value is missing or blank.
- The value must be identical across all API instances sharing the same database and remain stable so committed `Idempotency-Key` requests remain replayable.

### Pre-release

1. Confirm the Provider Credential and Custody Connection schema is deployed.
2. Apply `0034_provider_credential_submission_idempotency.sql`.
3. Deploy all application instances with `PRIVY_BYOK_PROVISIONING_ENABLED=false`.
4. Before enabling Privy BYOK submissions, configure a stable `CREDENTIAL_FINGERPRINT_PEPPER` on every API instance:
   - Generate with `openssl rand -hex 32`.
   - New self-hosted configurations generate it automatically.
   - Existing installations must set it explicitly before enabling Privy BYOK provisioning.
5. For managed deployments:
   - Verify `GCP_SECRET_MANAGER_PROJECT_ID`.
   - Verify the runtime service account has the required Secret Manager access.
6. For self-hosted deployments:
   - Verify `CUSTODY_ENCRYPTION_KEY`.
   - Use the default `encrypted_db` backend or configure it explicitly.
7. Select a canary Organization with the Privy entitlement.

### Post-release

1. Wait until every application instance is running the new version.
2. Enable `PRIVY_BYOK_PROVISIONING_ENABLED` for the target environment.
3. Run a canary submission and verify:
   - `201 Created`.
   - One pending Provider Credential and one pending Custody Connection.
   - No wallet or `custody_scope_defaults` row is created.
   - Repeating the request with the same key returns the committed result without another secret write.
4. Monitor:
   - `provider_credential_orphan_risk`.
   - Unexpected increases in `403`, `409`, or `503` responses.
5. No legacy backfill or post-deployment data migration is required.

## Result validation

- Full API suite: **1278 passed**, **14 skipped**.
- Provider credential route: **35 passed**.
- Credential secret store: **23 passed**.
- API lint: passed.
- API and workspace typechecks: passed.
- Specification and standards reviews: no remaining findings.

## Does it affect current flows?

Only fresh legacy Privy setup admission:

- When `PRIVY_BYOK_PROVISIONING_ENABLED` is enabled for an entitled Organization, fresh legacy Privy creation or reactivation returns `403 Forbidden`.
- A pending, checking, or active Privy Connection blocks fresh legacy setup with `409 Conflict`, even after the flag is disabled.
- When the flag is disabled and no blocking Connection exists, legacy setup follows its previous behavior.
- Existing active legacy behavior is unchanged:
  - repeated `POST /v1/wallets/initialize` returns `409 ALREADY_INITIALIZED`;
  - `POST /v1/wallets/switch` can still select an active Config.
- Existing wallet, signing, and custody-resolution flows are unchanged.
- Other custody providers are unaffected.

## Notes

- Provider validation and Connection-owned wallet creation are deferred to the next tasks.
- Post-acceptance credential rotation is deferred to the next task.
- Legacy data migration is not required for this MVP.
- A shared durable reservation between old legacy writers and the new stored-credential writer is not included. During a mixed-version rollout, an old instance could race with stored setup. Keep the feature flag disabled until all instances run this version; add shared serialization or reservation if mixed-version write safety becomes required.


pashpashkin:
@greptileai review

apps/sdp-api/src/services/provider-credential-submission.service.ts
```diff
@@ -0,0 +1,921 @@
...
+    },
+  });
+}
+
+async function runSubmissionTransaction(submission: StoredSubmission): Promise<TransactionResult> {
+  return submission.db.transaction(async (tx) => {
+    const txStore = new ProviderCredentialStore(tx);
+    if (!(await txStore.lockProject(submission.organizationId, submission.projectId))) {
+      throw new Error("Project disappeared during credential submission");
+    }
```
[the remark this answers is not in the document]
pashpashkin: Why do you think having two different credentials in the same organization, each attached to a different project, is a problem?

apps/sdp-api/src/services/provider-credential-submission.service.ts
```diff
@@ -0,0 +1,921 @@
...
+
+async function computeSubmissionFingerprint(context: SubmissionContext): Promise<string> {
+  let pepper: string;
+  try {
+    pepper = requireEnv({ API_KEY_PEPPER: context.c.env.API_KEY_PEPPER }, "API_KEY_PEPPER");
+  } catch {
+    await auditFailure(context.c, context.audit, context.auditBase, {
+      reason: "missing_fingerprint_pepper",
+    });
+    throw internalError();
```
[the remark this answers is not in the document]
pashpashkin: Credential submission fingerprints now use a dedicated `CREDENTIAL_FINGERPRINT_PEPPER`

pashpashkin:
@greptileai review

apps/sdp-api/src/routes/internal-custody/index.ts
```diff
@@ -0,0 +1,49 @@
...
+const privyCredentialSubmissionSchema = z
+  .object({
+    provider: z.literal("privy"),
+    fields: z
+      .object({
+        credentialLabel: z.string().trim().min(1),
+        scope: z.enum(["organization", "project"]),
+        appId: z.string().trim().min(1),
+        appSecret: z.string().min(1),
+      })
```
GuiBibeau, line 20: Could we keep `walletLabel` in the request/catalog and persist it in `setup_metadata` when creating the connection? HOO-769 calls this out for HOO-770, but this path currently rejects or drops it.

apps/sdp-api/src/services/stores/provider-credential.store.ts
```diff
@@ -0,0 +1,434 @@
...
+       matching_connections AS (
+         SELECT c.id, c.organization_id, c.project_id, c.provider, c.scope,
+                c.provider_credential_id, c.provider_credential_scope_key,
+                c.default_custody_wallet_id, c.status, c.setup_metadata,
+                c.last_check_status, c.last_check_at, c.last_check_failure_code,
+                c.activated_at, c.created_at
+         FROM custody_connections c
+         WHERE c.organization_id = ?
+           AND c.project_id = ?
+           AND c.provider_credential_id IN (SELECT id FROM lineage)
```
GuiBibeau: What should replay mean after replacement: the original result, current connection state, or a conflict? Right now it returns the old credential alongside a connection pointing to the replacement.

pashpashkin: Yeah, this is confusing, and the client would need to handle it explicitly

I wanted to keep the implementation simple and avoid storing the original response, and still return the Credential created by the original request. But returning 409 after the Connection moves to a replacement Credential may be clearer.

apps/sdp-api/src/routes/internal-custody/index.ts
pashpashkin: I was thinking of separating credential setup from wallet creation. This endpoint would only store the credentials and create a pending Connection. walletLabel would be sent later with the explicit `POST /v1/wallets` request, so we don’t store wallet details before the user decides to create one
Does it make sense?

GuiBibeau (dismissed):

packages/sdp-env-config/src/fields.ts
```diff
@@ -536,6 +536,14 @@ export const FIELDS: EnvField[] = [
...
     required: true,
   },
+  {
+    key: "CREDENTIAL_FINGERPRINT_PEPPER",
+    section: "secrets",
+    kind: "secret",
+    label: "Credential fingerprint pepper",
+    required: true,
+    help: "HMAC key for stored-credential request fingerprints. Keep it stable so committed requests remain replayable.",
+  },
```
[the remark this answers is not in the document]
pashpashkin: Good catch. Re-running the configurator is not the normal upgrade path on the working app, but it is plausible. The CLI currently overwrites .env and rotates all generated secrets, not only this pepper. This is a broader self-hosted configuration issue and will be addressed separately from this ticket

apps/sdp-api/src/services/stores/provider-credential.store.ts
pashpashkin: I changed the contract so the idempotency key identifies the credential submission itself, not the mutable Connection state.  
A replay now returns only the original providerCredential in its current persisted state with 201. The Connection is still created or reassigned atomically, but it remains internal and is no longer returned or resolved during replay.  
After replacement, the original idempotency key returns the Credential created by the original request in its current persisted state, while the new key returns the replacement Credential

arseniy-nikitochkin (dismissed):

resourcefulmind (commented):
Reviewed the full diff at `e59e2744`, all 27 files. Confirmed CI green on that exact commit (29 check runs, zero non-success) rather than on the PR rollup, since the last two merges landed after the bot stopped reviewing.

The credential handling is careful and I found no correctness, data-integrity, or secret-exposure defect. Fail-closed on a missing pepper before any write, secret-before-transaction with compensation on every rollback path, unlocked preflight re-classified under lock, organization-scoped replay with a fingerprint check. The concurrency and compensation tests are genuinely good, particularly the lost COMMIT reconciliation and the cross-project idempotency race.

One thing I would like confirmed before this goes into a release, plus three notes.

**Behavior change on a public route that is not flag-gated.** `POST /v1/wallets/initialize` with `provider: "privy"` and an active exact-project config used to return 409 CONFLICT, via `SigningError("ALREADY_INITIALIZED")`. It now returns 201 with the existing config. The new reuse branch checks only `existing?.status === "active"`, so this is not behind `PRIVY_BYOK_PROVISIONING_ENABLED` and reaches every org with an active legacy Privy config as soon as this deploys.

That makes two lines in the description inaccurate: "no public API contract is changed" (this route is mounted under `/v1`) and "Existing active legacy Privy Config reuse remains unchanged" (the reuse behavior is new). It also means the rollback story for this specific change is a revert, not a flag flip, which is worth stating explicitly in the release notes.

I checked `apps/sdp-web` and found nothing branching on CONFLICT for this call, so the in-house impact looks small. Could you confirm the 409 to 201 change is intentional? If yes, I will carry it in the release notes as a behavior change. If it is intentional but not wanted yet, it could sit behind the same flag.

Notes, none blocking:

- GCP secret containers are created per credential id, but compensation only calls `destroyVersion`. A compensated write leaves an empty container behind permanently, with no reaping path. Bounded by conflict frequency, so a follow-up ticket rather than a change here.
- The `runtime_env` rejection in `createSubmissionSecretStore` is the one branch in the new service without a test.
- I re-verified all three Greptile P1s against this head. Two are resolved (the organization-roots one was retracted and the schema confirms it, the pepper one is fixed by the dedicated variable). The configurator rotation one is live, and I agree with deferring it: `collectFromEnv` preserves a provided value, so it only bites on a re-run without the current env, and it is the same class of issue as `API_KEY_PEPPER` rotation, which is worse and already present.

GuiBibeau (dismissed):

resourcefulmind (approved):
Re-reviewed at `569633a1`, approving. Removing the reuse branch instead of flag-gating it is the better fix: initialize is back to 409 for an active exact-project config, the audit log is out of the conditional, and both description lines I flagged are accurate now, so rollback is a flag flip again rather than a revert.

I did stress the new early return in `assertFreshPrivyLegacySetupAllowed`, since bailing out of a security gate is where this kind of fix usually goes wrong. It holds, details in a comment on the line. The conflict test is a real regression test too, it runs flag-on and entitled, which is exactly the case that would have 403'd before.

CI green on `569633a1` specifically. No Greptile reviews available as at review time, so I re-checked its three findings by hand again, none are in the changed files.

PR #885, opened 2026-07-23
fix(policies): stabilize search input and polish controls layout

tobySolutions:
## Summary

- **Search stability (fix):** the wallet/API-key search box could revert to a stale query. It mirrored the URL back into the input on every `state.query` change, so a slow RSC round-trip for a previous query overwrote text the user had typed since — the field "stuck" on old text when clearing and retyping. It now only adopts the URL on genuine back/forward navigation, tracking the query it last pushed itself (debounced typing **and** Clear-filters paths).
- **Layout:** the table and summary are now independent cards with `align-items: start`, so a short table no longer stretches into a dead gap beside the taller summary. The summary is a sticky side panel.
- **Table:** comfortable, non-colliding column widths with a clean horizontal-scroll fallback on narrow viewports; the design-system scroll shadow now fades in the card surface (`--table-bg`) instead of painting a dark band over the rows; "Configure controls" no longer wraps and the toolbar collapses to two columns on tablet.

No API, data, or behavior changes — presentational + one client-state fix.

## Validation

- `pnpm typecheck` — 11/11 workspaces
- `pnpm lint` (biome) — 0 errors (43 pre-existing warnings, none added)
- `pnpm --filter sdp-web test` — 55 files, 311 passed
- Manual: search clear → immediate retype (no revert); table at desktop / tablet / narrow widths; light + dark.

## UI

Visual-only changes to `/dashboard/policies`:
- Table and summary split into two cards (removes the empty gap under a short table); summary is now sticky.
- Table columns rebalanced with breathing room; horizontal scroll on narrow screens uses a soft surface-colored fade instead of a dark shadow bar.
- Toolbar wraps gracefully on tablet; "Configure controls" no longer breaks onto two lines.

Verified locally across breakpoints and themes.

### Before
Screenshot 2026-07-23 at 14 42 49

### After 

Screenshot 2026-07-23 at 14 45 33


GuiBibeau (approved):

resourcefulmind (approved):
Clean fix, the `lastPushedQueryRef` approach solves the search-revert at the root and the two-card layout reads well. 
One optional follow-up: worth a regression test for the search-revert itself, only the layout assertion changed. Approving 🙌

PR #915, opened 2026-07-24
chore(api): remove dead sessions KV store

G1de0n:
## Summary

Follow-up cleanup requested in the #896 review: the `sessions` KV store is now dead code.

After #896, `SessionService.getSession` resolves membership and permissions authoritatively from Postgres on every request and no longer reads, writes, or deletes the `session:<id>` KV cache. The `sessions` store was therefore write-only and, after the cache removal, entirely unused — yet `kv-redis.ts` still provisioned it.

This removes the dead store:
- drop `sessions` from the `KVStoreSet` interface (`runtime/kv.ts`)
- drop `sessions` from `STORE_PREFIXES` and `createKVStoreSet` (`runtime/kv-redis.ts`)
- drop `kv.sessions` from the `clearKVStores` test helper
- update the prefix-isolation test to assert three stores instead of four

No behavior change: nothing read or wrote this store.

## Validation

- `pnpm --filter @sdp/api typecheck` — passes (whole project; confirms no remaining consumer)
- `pnpm --filter @sdp/api lint` (biome) — clean
- `pnpm --filter @sdp/api test` — 132 files, 1407 passed, 14 skipped (Postgres via testcontainers)


resourcefulmind (approved):
Exactly the cleanup that was needed, nothing more. Verified there are zero remaining references to the sessions store anywhere in the API source, so this is pure deletion. Closes the loop on #896. 👍

PR #916, opened 2026-07-24
feat: free-form key-value rampsMemo on ramp quote transfers

multipletwigs:
## Summary

Adds an optional `rampsMemo` field — free-form key-value pairs for offchain reconciliation — to on/offramp quote creation. The pairs are stored on the `payment_transfers` row the quote handler creates and returned on all transfer reads.

### API
- Migration 0036: `payment_transfers.ramps_memo JSONB NOT NULL DEFAULT '{}'` with a `jsonb_typeof = 'object'` check (same pattern as `provider_data`).
- `rampsMemo` accepted on `POST /v1/payments/ramps/onramp/quote` and `offramp/quote` — up to 20 pairs, keys ≤64 chars, values ≤256 chars — threaded through `persistRampQuoteTransfer` and the BVNK pending-transfer path into `createTransfer`.
- Returned by the transfer list/detail mappers; documented in the openapi transfer schema and onramp quote request. (The offramp quote request has never been registered in the openapi spec — pre-existing gap, not addressed here.)
- The existing string `memo` column is untouched; `rampsMemo` is a separate structured field.

### Shared types
- `PaymentOnrampQuoteRequest` / `PaymentOfframpQuoteRequest` and `rampsMemo` on `PaymentTransferSummary`.
- `RAMPS_MEMO_LIMITS` in `@sdp/types` is the single source for the 20/64/256 limits — imported by the API zod schema, the dashboard validation, and the dialog input constraints.

### Dashboard
- Both ramp wizards get an optional **Add memo** row on the details step that opens a modal patterned on the bulk-import dialog: editable key/value grid, paste-anywhere parsing of tab- or comma-separated rows, add/remove rows, inline validation, en+fr strings.

## Test plan
- `payments.test.ts`: the memo persists and is returned via list + detail for a MoonPay onramp quote and a BVNK offramp channel quote; 21-key payloads are rejected with 400. Full file passes (93 tests).
- openapi spec tests (7) and touched web unit tests (15) pass.
- Typecheck clean for `@sdp/types`, `@sdp/api`, `sdp-web`; biome clean on touched files.


multipletwigs:
@greptileai fixed the P1 — rampsMemo now carried through normalizePaymentTransfer. Please re-review.

resourcefulmind (approved):
Clean feature, walked the whole thing. Limits are single-sourced and enforced on both ends, round-trip holds across every persist path, en/fr parity complete. Greptile's P1 is stale, `normalizePaymentTransfer` sets `rampsMemo` on the current head `(payments-page.data.ts:143/170)` and nothing reads it on the frontend anyway. Approving. 

Wanted to also suggest the below:
- The 64/256 length limits have no negative test (only the 21-key 400), so a future loosening ships green; and pasting a comma value into a single memo cell gets reparsed into rows (recoverable, but worth gating row-parsing on multi-cell paste). 🙌

PR #901, opened 2026-07-23
fix(security): isolate playground API keys by project

GuiBibeau:
## Summary
- require an authenticated dashboard user and active organization in the playground execute route
- verify supplied API-key material by exact hash against the session organization and selected project before forwarding
- fail closed for foreign, inactive, expired, or unresolvable keys while preserving Clerk-session execution
- keep the verification helper internal and exclude it from public API discovery surfaces

## Security validation
- unauthenticated callers are rejected before an API client is created
- a key outside the selected tenant boundary is rejected without executing the requested operation
- a same-project key is forwarded only after exact ownership verification
- verification failures stop execution and return a generic error
- API key display prefixes are not trusted for ownership and full secrets are not logged or stored

## Local validation
- `pnpm --filter sdp-web test:unit` — 67 files / 385 tests passed
- `pnpm --filter @sdp/api test` — 126 files / 1,310 passed / 14 skipped
- `pnpm --filter sdp-web typecheck` — passed
- `pnpm --filter @sdp/api typecheck` — passed
- `pnpm --filter sdp-web build` — passed
- `pnpm --filter @sdp/api build` — passed
- `pnpm --filter @sdp/api lint` — passed
- `pnpm check:module-boundaries` — passed
- targeted Biome check — passed
- `pnpm --filter sdp-web lint` — blocked before linting by the current ESLint 10 / eslint-plugin-react compatibility error; CI uses changed-file Biome checks instead

## Screenshots
Not applicable: this changes only server-side authorization and has no visible UI changes.

Finding: `finding_e76e2d4a7a226385`.

PR #922, opened 2026-07-24
fix(playground): give the un-run Response tab a real empty state

tobySolutions:
## What

Before a request runs, the API playground **Response** tab rendered a placeholder JSON object (`{ "message": "Run request to inspect..." }`) as a code block, which looked unfinished next to the **Example** tab sample.

## Fix

Show a proper empty state (braces icon + prompt) in the Response tab until a request is executed. Once there is a result or error, the live JSON renders as before.

Change is in the shared `ApiPlaygroundShell`, so it applies to every playground: payments, custody wallets, payment requests, counterparty, and issuance.

## Verification

- typecheck / biome / unit (386) all pass
- Verified all 6 playgrounds route through the shared shell; nothing else renders the Response tab.

### UPDATES:
Screenshot 2026-07-24 at 14 46 28


resourcefulmind (approved):
Nice fix. The empty state only renders when there is genuinely no result and no error, the copy reuses an existing translated key, and the execute path from #901 is untouched. Approving. 

Two tiny follow-ups if you're in that file again: the old placeholder branch in `buildResponseBody` is dead code now, and during a re-run the panel shows "Run request..." while a request is in flight, a small loading state would read better there.

arseniy-nikitochkin (approved):

PR #939, opened 2026-07-25
feat(types): broaden the well-known token catalogue

tobySolutions:
**Stack** — merge top-down:
1. **#939 ← this PR**
2. #943

## What

Adds the assets orgs most commonly hold alongside the existing stablecoins:

| Group | Tokens |
| --- | --- |
| Stablecoins / FX | EURC, USDS, USDY |
| Liquid staking | JitoSOL, mSOL, bSOL, INF |
| Wrapped BTC | cbBTC, WBTC |

Also adds `name` and `category` to support grouped, labelled token pickers.

## Every mint was checked against chain

Address, decimals and owning token program verified on mainnet and devnet before being pinned. Devnet entries are declared only where the address resolves to a real mint on devnet with matching decimals. Two cases where it does not:

- **USDS** has an account at its address on devnet, but it is a different mint with 9 decimals rather than 6. Declaring it would have misscaled every devnet amount by three orders of magnitude. Devnet omitted.
- **JUP, BONK, RAY, ORCA and WBTC** resolve to System Program accounts on devnet rather than mints. Mainnet-only.

An existence check alone would have shipped both of those wrong.

## isUsdStable

This flag prices a token at $1 without a feed (`helius-das.service.ts:137,145`), so it stays reserved for tokens that actually track a dollar:

- **EURC** is euro-denominated → `false`
- **USDY** accrues yield above par → `false`

Both are commonly called stablecoins, neither is USD-stable in the sense this flag means.

## On pinning rather than fetching

Solana's production-readiness guidance is to hardcode or allowlist mint addresses rather than resolve them from a registry at runtime, because spoofed tokens share the name and symbol of the real asset and only the mint distinguishes them. This is also a policy allowlist, so a third party changing our list at runtime is not a property we want. `jup-ag/token-list` is archived, and no registry covers devnet at all.

## Verification

`pnpm --filter @sdp/api test` 1417 passed, `pnpm --filter sdp-web test:unit` passing, lint and module boundaries clean.


packages/sdp-types/src/well-known-tokens.ts
```diff
@@ -29,22 +41,40 @@ export interface WellKnownToken {
...
+ * Every mint here is pinned deliberately. Solana's production-readiness guidance
+ * is to hardcode or allowlist mint addresses rather than resolve them from a
+ * registry at runtime, because spoofed tokens share the name and symbol of the
+ * real asset and only the mint distinguishes them. Devnet and mainnet are
+ * separate deployments, so a token appears on a cluster only when it has a
+ * verified mint there.
+ *
+ * Every entry — address, decimals and owning token program — is checked against
+ * chain by the well-known token conformance test. Add tokens by running that
+ * test, never by copying an address from a chat or a search result.
```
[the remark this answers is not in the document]
tobySolutions: Correct — the comment pointed at something that didn't exist. Verification was a throwaway script, not a committed test.

Superseded since: `pnpm check:well-known-mints` now does it for real, and the devnet USDS line here is wrong — that account has zero supply, so it's an unrelated mint at the same address rather than USDS at a different scale. See the thread below.

multipletwigs:
@tobySolutions could we use svgs for the images instead?

packages/sdp-types/src/well-known-tokens.ts
```diff
@@ -29,22 +41,47 @@ export interface WellKnownToken {
...
+ * real asset and only the mint distinguishes them. Devnet and mainnet are
+ * separate deployments, so a token appears on a cluster only when it has a
+ * verified mint there.
+ *
+ * Every entry here — address, decimals and owning token program — was verified
+ * against mainnet and devnet before being added, which is how USDG was found to
+ * be a Token-2022 mint and how devnet USDS was rejected for having different
+ * decimals than its mainnet counterpart.
+ *
+ * There is no automated conformance check yet. Until there is, verify a new
```
multipletwigs: What is a conformance check?

packages/sdp-types/src/well-known-tokens.ts
```diff
@@ -76,16 +117,139 @@ export const WELL_KNOWN_TOKENS = {
...
   },
+  USDS: {
+    symbol: "USDS",
+    name: "USDS",
+    decimals: 6,
+    isUsdStable: true,
+    tokenProgram: "spl-token",
+    category: "stablecoin",
+    mints: {
+      // Devnet deliberately omitted: an account exists at this address on devnet
```
multipletwigs: Interesting, then in this case perhaps moving decimals into this mints key would be better

packages/sdp-types/src/well-known-tokens.ts
```diff
@@ -76,16 +117,139 @@ export const WELL_KNOWN_TOKENS = {
...
+      // but it is a different mint with 9 decimals, which would misscale amounts.
+      // biome-ignore lint/security/noSecrets: Mainnet USDS mint address constant, not a secret.
+      "mainnet-beta": "USDSwr9ApdHk5bvJKMjzff41FfuX8bSxdKcR81vTwcA",
+    },
+  },
+  USDY: {
+    symbol: "USDY",
+    name: "Ondo US Dollar Yield",
+    // Yield-bearing: redeems above 1 USD and drifts upward, so it must not be
```
multipletwigs, line 185: makes sense, so it's not a stable, might need price discovery with jupiter, could u add into a follow up where we swap over from helius? our price discovery is not very good rn

multipletwigs:
wait @tobySolutions what changes is this for the policy engine? could u add screenshots on how it would look like?

packages/sdp-types/src/well-known-tokens.ts
tobySolutions: Nothing, I made the term up and it pointed at something that didn't exist.

Built it instead: `pnpm check:well-known-mints` checks every entry against chain (is it a mint, decimals, owning program). 22/22 pass. Not in CI since it needs public RPC.

packages/sdp-types/src/well-known-tokens.ts
tobySolutions: Agreed, done in a7a875f4. `mints` is now `{ address, decimals }` per cluster.

Also correcting myself: the devnet account has zero supply, so it's an unrelated mint at the same address, not USDS at a different scale. Stays out for that reason.

packages/sdp-types/src/well-known-tokens.ts
tobySolutions: Yeah. `resolveKnownUsdPrice` (helius-das.service.ts:131) hardcodes 1 for anything isUsdStable and null otherwise, so Helius carries everything else.

USDY is isUsdStable:false on purpose, so it never gets the shortcut. Same for EURC and the staked-SOL entries.

tobySolutions:
SVGs: swapped 6 of 9 in a7a875f4 (SOL, USDC, PYUSD, mSOL, EURC from @web3icons/core, MIT; bSOL from SolBlaze). 153KB → 98KB. SOL was a black square in a round chip and EURC had no logo at all, so those two were real bugs.

The other 3 can't be. Jito and Paxos both ship PNG only, and every WBTC vector I found is drawn in flat white so it renders invisible on our chip.

Policy engine: it's the asset picker in wallet policy authoring. Before, it only offered assets the wallet already held, so a new wallet got an empty list. Now the catalogue seeds a "Common tokens" group with logos and names, filtered to the active cluster. Screenshot below is sandbox, which is why it's the devnet set.

tobySolutions:
> wait @tobySolutions what changes is this for the policy engine? could u add screenshots on how it would look like?

Screenshot 2026-07-27 at 15 33 02


Screenshot 2026-07-27 at 15 33 06
 
 Those other ones are local tokens I seeded in the db for test so that's why they do not have logos chief.

packages/sdp-types/src/well-known-tokens.ts
tobySolutions: i will pick it up in a follow-up PR and thrash it out there chief. Thanks for surfacing.

packages/sdp-types/src/well-known-tokens.ts
```diff
@@ -16,85 +16,292 @@ export const SPL_TOKEN_PROGRAMS = {
...
+ * separate deployments, so a token appears on a cluster only when it has a
+ * verified mint there.
+ *
+ * Every entry here — address, decimals and owning token program — was verified
+ * against mainnet and devnet before being added, which is how USDG was found to
+ * be a Token-2022 mint and how the devnet account at the USDS address was found
+ * to be an unrelated 9-decimal mint rather than USDS.
+ *
+ * After adding or changing an entry, check it against chain with
+ * `pnpm --filter @sdp/types mints:verify`. That confirms, for every cluster a
```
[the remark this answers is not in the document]
tobySolutions: Fixed in 12440e43 — it's `pnpm check:well-known-mints` from the root now, which is the command that actually exists.

scripts/check-well-known-mints.mjs
```diff
@@ -0,0 +1,106 @@
...
+// for that token.
+//
+// Deliberately not wired into CI: it needs public RPC, and a rate-limited
+// endpoint would make the pipeline flaky for no safety gain. Run it by hand
+// whenever you add or change a catalogue entry.
+
+import {
+  SPL_TOKEN_PROGRAMS,
+  WELL_KNOWN_TOKENS,
+} from "../packages/sdp-types/src/well-known-tokens.ts";
```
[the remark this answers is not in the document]
tobySolutions: Good catch, this was real. engines.node is >=22.0.0 and CI runs Node 22, where importing the TS source under bare node fails outright — I'd only run it on Node 24.

Fixed in 04de01e6: it now goes through the tsx @sdp/api already depends on, so transpiling no longer depends on the runtime. Adding tsx at the root would need a new direct dependency and this workspace's minimumReleaseAge cooldown won't resolve one. Verified: 22 verified, 0 failed.

resourcefulmind (approved):
Nice catalogue expansion. I verified every new mint against live RPC: correct addresses, decimals, and token programs, no duplicates. The bigger change (moving decimals per-mint and turning `mints[cluster]` into `{ address, decimals }`) is fully propagated to consumers and typechecks clean. 
Also confirmed the verifier fix from your latest push, running it through `tsx` means it now works on the whole supported Node range instead of crashing on the `.ts` import below 22.18, and the JSDoc command is corrected. 

Since that verifier is our only guard for mint correctness and it's not in CI, good to have it working for the next edit. The earlier red Token Flows check was a Doppler-install flake, now clear. Approving.

tobySolutions:
> Nice catalogue expansion. I verified every new mint against live RPC: correct addresses, decimals, and token programs, no duplicates. The bigger change (moving decimals per-mint and turning `mints[cluster]` into `{ address, decimals }`) is fully propagated to consumers and typechecks clean. Also confirmed the verifier fix from your latest push, running it through `tsx` means it now works on the whole supported Node range instead of crashing on the `.ts` import below 22.18, and the JSDoc command is corrected.
> 
> Since that verifier is our only guard for mint correctness and it's not in CI, good to have it working for the next edit. The earlier red Token Flows check was a Doppler-install flake, now clear. Approving.

Appreciate the feedback sir. Thank you very much for the promptness!

PR #941, opened 2026-07-25
feat(shell): give sidebar sub-nav items their own icons

tobySolutions:
**Stack** — merge top-down:
1. **#941 ← this PR**
2. #945
3. #946

## What

Top-level sidebar items carry an icon; their children did not. The payments group expands into six unmarked text rows, which gets harder to scan as sub-items are added.

`SubNavItem` gains an optional `icon`, and the six payments children are given one:

| Child | Icon |
| --- | --- |
| Transactions | `ReceiptTextIcon` |
| Counterparty | `UsersIcon` |
| Pay | `ArrowUpRightIcon` |
| Deposit | `ArrowDownLeftIcon` |
| Requests | `FileTextIcon` |
| Recurring | `RepeatIcon` |

## Notes

The field is optional, so any nav group that has not been assigned icons renders exactly as before. Both the link and the disabled variants render it, so a locked sub-item stays aligned with its siblings.

Icons are `aria-hidden` — the adjacent label already names the destination, so announcing them would just be noise.

## Verification

`pnpm lint`, `pnpm check:module-boundaries` and `pnpm --filter sdp-web test:unit` (392 passed) all clean. No new typecheck errors.

PR #942, opened 2026-07-25
feat(members): build the members page on the existing /v1/members API

tobySolutions:
> Stacked on #941. Review that first; this diff narrows once it merges.

## Context — this was not a greenfield stub

The page rendered a "coming soon" card, but the backing work was already done and simply never wired up:

- **`/v1/members` is implemented and mounted** (`app.ts:37`) — `GET /`, `POST /invite`, `POST /accept`, `DELETE /:memberId`
- **`app/members/actions.ts` already had `listMembers` and `inviteMember`** written against those endpoints, and **never imported by anything**

So this connects existing pieces rather than adding a surface.

## Two defects in the unreachable action

Both would have fired the first time anyone used it:

1. It threw `t("Shared.validation.emailRequired")` — **that key does not exist**. The validation path would have rendered a raw key or crashed.
2. It revalidated `/members`, which only redirects. The page lives at `/dashboard/members`, so the list would not have refreshed after an invite.

## Behaviour

- Lists active members: name, email, role, join date
- Invite form posts to the existing endpoint; `inviteMember` goes through Clerk server-side, so this surfaces SDP's own view of membership rather than duplicating Clerk's
- Failures return instead of throwing, so the reason renders inline rather than tripping an error boundary and discarding the typed-in address
- A failing member list no longer takes the invite form down with it — an admin whose list is erroring can still need to add someone

**Remove is deliberately not exposed.** It is `org:admin` and destructive; it belongs with a confirmation flow, not a first pass.

## Worth a decision at review

Gui's steer was *"I think clerk handles all this, we probably should clean that up"* — which predates knowing there is a complete `/v1/members` API plus written-but-unwired actions behind the stub. If the intent is still that Clerk owns membership and SDP should not present its own view, this should be closed and the stub deleted instead. Easy either way; that is why it is a small, self-contained PR.

## Implementation note

The design-system `Select` is controlled and renders no native form control, so the chosen role is mirrored into a hidden input to reach the action's `FormData`.

## Verification

`pnpm lint` 0, `pnpm check:module-boundaries` 0, `pnpm --filter sdp-web test:unit` 392 passed, no new typecheck errors. English and French copy added together.


tobySolutions:
Superseded by #945 (API) and #946 (UI). Same change split so the API surface and the dashboard work review separately — the combined branch had grown to 24 files across both, including an auth-path guard that deserves its own attention rather than being buried in a large diff.

Verified lossless: the two branches diff clean against the original head, so nothing was dropped or added in the split.

PR #946, opened 2026-07-25
feat(members): members management on the settings page

tobySolutions:
**Stack** — merge top-down:
1. #941
2. #945
3. **#946 ← this PR**

## Where it lives

Members render as a card on `/dashboard/settings`, **not** a standalone route.

The sidebar was already scrolling before this work, and members is organization configuration rather than a product module — grouping it with issuance and payments would misfile it, and adding a top-level entry made the overflow worse. `/members` redirects to settings so existing links still land.

## What it shows

Members and **pending invitations in one list**, because an invited person holds no membership row until they accept and otherwise appears nowhere.

| Row type | Actions |
| --- | --- |
| Pending invitation | Copy invite link · Revoke |
| Member | Remove (confirmation dialog) |

**Copy invite link** matters because email delivery cannot be confirmed from the product — it is the only way to unblock someone whose mail never arrived. Hidden when Clerk could not be reached and no link exists, since a copy that silently yields nothing is worse than no button.

**Remove** is disabled with the reason shown when it would remove you or the last admin. The API enforces both; the menu only reflects it, so the rule is visible before someone commits rather than as an error after.

## Display

Where a member has no name the **email becomes the identity line** rather than sitting under an "Unnamed" placeholder that says nothing and reads as an error.

An email that is not an address is not rendered as one. A misconfigured JWT template stored a literal `{{…}}` placeholder as a user's email, and printing it verbatim looked like a rendering fault rather than the data problem it was. (#945 stops new ones; this stops existing ones looking broken.)

## Other

- Results report through **sonner**, matching the RPC settings form directly above. Success resets the invite form; failure keeps the address so it can be corrected
- Paging uses `membersPage` in the URL, so a page survives refresh and can be linked. Page one stays out of the query string. Paging does not scroll to top — the table sits below the fold
- Reuses the existing `ArrowPagination` rather than adding another control
- Fills in the **settings skeleton**, which drew only the organization card and so jumped from one card to two once members loaded

## Screenshots

Worth attaching before merge — this is almost entirely visual.

## Verification

`pnpm --filter sdp-web test:unit` **391 passed**, `typecheck` 0, `lint` 0, `check:module-boundaries` 0 — on this branch alone.

Invite → accept → list flow confirmed end to end against a real Clerk invitation.


tobySolutions:
Screenshot 2026-07-25 at 19 57 52

resourcefulmind (dismissed):
Clean members UI. The remove/last-admin gating mirrors #945's server guards, and copy-link just writes the accept URL to the clipboard. 
The red Surfpool/Access check is a flake (#945 passes that same shard, and this PR adds no access code), so I re-ran it. 
One non-blocking note: pending invites aren't paginated, so they repeat past 25 members. Approving 🙌

tobySolutions:
> Clean members UI. The remove/last-admin gating mirrors #945's server guards, and copy-link just writes the accept URL to the clipboard. The red Surfpool/Access check is a flake (#945 passes that same shard, and this PR adds no access code), so I re-ran it. One non-blocking note: pending invites aren't paginated, so they repeat past 25 members. Approving 🙌

Thank you chief! Looking into this as well and thanks for the approval sir.

tobySolutions:
> Clean members UI. The remove/last-admin gating mirrors #945's server guards, and copy-link just writes the accept URL to the clipboard. The red Surfpool/Access check is a flake (#945 passes that same shard, and this PR adds no access code), so I re-ran it. One non-blocking note: pending invites aren't paginated, so they repeat past 25 members. Approving 🙌

Thanks for the feedback on this chief.

tobySolutions:
one quick note @resourcefulmind sir.

#945 makes the API reject a Clerk token when the email claim isn't a real address. Our dev JWT template had a bad shortcode so Clerk was sending the literal {{user...}} string instead of the email.

Can you check the sdp-api template on staging and prod before you promote? Clerk dashboard > JWT Templates > sdp-api > Claims. It should be "email": "{{user.primary_email_address}}". If it has .email_address on the end, that's the broken one, just edit and save.

resourcefulmind (approved):

resourcefulmind:
> one quick note @resourcefulmind sir.
> 
> #945 makes the API reject a Clerk token when the email claim isn't a real address. Our dev JWT template had a bad shortcode so Clerk was sending the literal {{user...}} string instead of the email.
> 
> Can you check the sdp-api template on staging and prod before you promote? Clerk dashboard > JWT Templates > sdp-api > Claims. It should be "email": "{{user.primary_email_address}}". If it has .email_address on the end, that's the broken one, just edit and save.

Thanks for the heads-up. Confirmed the code side: #945's guard rejects a token whose email claim isn't a real address, so a bad template shortcode would surface as an auth failure rather than getting silently stored. I'm treating the sdp-api template check as a pre-promote gate on 0.49.0 and coordinating it with the infra owner before the release goes out. Will confirm here once it's sorted. 🙏

PR #951, opened 2026-07-27
feat(web): ramp memo as optional wizard step with view-summary modal

multipletwigs:
### Memo becomes its own optional step
- New `MEMO` step in both ramp wizards (onramp: Deposit → Memo → Provider; offramp: Wallet → Withdraw → Memo → Complete), with the provider-requirements step still inserting itself after it when collection is needed.
- The quote/onboarding advance now fires on leaving the memo step, so memo fields always ride the quote payload.
- The step edits key-value rows inline — paste-splitting, per-row validation, and a live JSON preview of the exact `rampsMemo` payload — replacing the `MemoField` trigger row + `MemoDialog` modal, which are deleted.
- Memo rows live in `useRampWizard` (`memoRows`), gating Next on row validity; an empty memo is valid since the step is optional.

### View summary modal
- `PaymentsWizardFrame` accepts an optional `summary` node and renders a **View summary** button in the step header (from the second step onward) that opens it in a modal.
- New `WizardSummaryList` shows the labeled selections made so far: counterparty, method, wallet, amount, receive asset / payout currency, provider, and memo field count.
- Each wizard hook exposes its own `summaryDetails`; shared builders (`preStepSummaryDetails`, `optionalDetail`, `memoSummaryDetails`) live in `ramps/wizard-summary.ts`. Wired across the pre-steps, onramp, offramp, onchain send, and onchain receive flows.

CleanShot 2026-07-27 at 12 46 13@2x

Including View Json button as well in the row
CleanShot 2026-07-27 at 12 46 33@2x

Counterparty Transaction Dialog
CleanShot 2026-07-27 at 12 58 12@2x

## Notes
- Amounts in the summary use the existing `formatTokenAmount` for locale grouping.
- All new copy added to both `en` and `fr` catalogs; unused memo dialog keys removed.


multipletwigs:
@greptileai Both findings addressed in c2cebb04:

- **Existing quote survives edits (P1)**: the onboarding-ready quote poller in `use-ramp-wizard.ts` is now additionally gated on `isLastStep`, so a quote can only be created while the user sits on the transaction stage. Combined with the existing back-becomes-cancel behavior once a quote exists, backward edits can no longer produce or keep a stale quote.
- **Summary reformats partial amounts (P2)**: added `summaryAmount()` in `wizard-summary.ts` — amounts only pass through `formatTokenAmount` when they're a complete decimal; partial input like `12.` is preserved as typed. Used by the onramp, offramp, and onchain-send summaries.

Please re-review.

multipletwigs:
@greptileai Fixed in the latest commit: `formatTokenAmount`'s exact-decimal branch now derives grouping and the decimal separator from `Intl.NumberFormat` for the active locale (whole part formatted as a BigInt so every input digit is preserved), instead of hardcoding commas/dots. French summaries render `1 234 567,89`. Added unit tests covering en/fr grouping, high-precision digit preservation, negative fractions, and non-numeric passthrough. Please re-review.

multipletwigs:
@greptileai The branch has settled — please review the current head (d71072a3). Since your last score: locale-aware separators in `formatTokenAmount` (with unit tests), memo keys normalized to dashed form at the single conversion point (with unit tests), the memo summary row simplified to a solid View JSON drill-in button, memo editor inputs matched to the ramps input-height tokens, and the redundant powered-by header badge removed in favor of the summary's provider row.

eldarik (approved):
lgtm

PR #961, opened 2026-07-27
fix(web): move the colour theme control into settings

tobySolutions:
The colour theme control was a light/dark switch pinned to the bottom of the sidebar, plus three more copies in the top bars and org onboarding.

Two problems with that:

- It lived in navigation chrome rather than in settings, where people look for preferences.
- It only ever wrote `light` or `dark` to storage. `defaultTheme="system"` was already set, but one click on the switch made system unreachable with no way back.

This replaces all four mounts with a single System / Light / Dark control in an Appearance card on `/dashboard/settings`. `theme-toggle.tsx` is deleted.

Notes:

- `useTheme` now returns the stored preference and a `hydrated` flag alongside the resolved theme. The three existing consumers only read the resolved theme and are unchanged.
- The card renders a placeholder until hydration instead of painting a selection, which would otherwise flash the wrong option for system dark users.
- The settings skeleton gains a matching third card.
- Not permission gated, since it is a per device preference rather than org state. Say the word if it should live somewhere else.
- The icon is hidden below 360px so "System" does not truncate.

Verified at 320, 390, 768 and 1440 in light and dark with no horizontal overflow. Theme e2e reworked for the new control, 11/11 passing. All pre-PR gates green.

Screenshot 2026-07-27 at 23 38 17
Screenshot 2026-07-27 at 23 38 28
Screenshot 2026-07-28 at 00 01 30
Screenshot 2026-07-28 at 00 01 35


resourcefulmind (approved):
Very clean refactor, and it fixes a real bug (system was unreachable once you clicked the old switch). The new control makes all three reachable and gates the selection on hydration so there's no wrong-option flash. 
I checked the backward-compat: the three existing useTheme consumers only read the resolved theme, which is unchanged, and the delete of `theme-toggle.tsx` leaves no dangling refs. en/fr parity is there and the e2e is reworked. 

On your question: leaving it ungated is right, it's a per-device localStorage preference, not org state, so it shouldn't be behind an org permission. 

Approving 🙌

tobySolutions:
> Very clean refactor, and it fixes a real bug (system was unreachable once you clicked the old switch). The new control makes all three reachable and gates the selection on hydration so there's no wrong-option flash. I checked the backward-compat: the three existing useTheme consumers only read the resolved theme, which is unchanged, and the delete of `theme-toggle.tsx` leaves no dangling refs. en/fr parity is there and the e2e is reworked.
> 
> On your question: leaving it ungated is right, it's a per-device localStorage preference, not org state, so it shouldn't be behind an org permission.
> 
> Approving 🙌

Thank you very much for this feedback chief! Very much appreciate the prompt feedback and thank for the confirmation on my gated-ness question.

PR #965, opened 2026-07-28
feat(web): ff for AlphaLedger integration for tokenization 

multipletwigs:
Groundwork for the AlphaLedger tokenization-engine integration:

- Adds the `alphaledger-tokenization-engine` dashboard feature flag (default off, "Mosaic only") to gate offering AlphaLedger as an issuance provider option.
- Small polish on the issuance wizard classification step: page-level heading, larger section labels, refined selection-card selected state, and the sub-type reveal no longer re-animates on every classification swap.


tobySolutions (approved):
Heads up, making the classification title an h1 gives the page two h1s. The create route already renders one in the top bar, the centered "New asset" title.

Keeping it as h2 should do it. text-2xl and tracking-tight are what make it bigger, so it still looks the same.

PR #969, opened 2026-07-28
fix(web): break sign-in loop for sessions without an active organization

### LINKED ISSUE #807: Dashboard onboarding loop when signed-in Clerk user has no active organization
# Dashboard sign-in loop when Clerk session has no active organization

## Summary

After a user signs in or signs up with Clerk, the app redirects them to `/dashboard`. The dashboard server layout requires both a Clerk `userId` and an active Clerk `orgId` before it will render.

This creates a loop for signed-in users who do not yet have an active Clerk organization selected: they are redirected back to `/sign-in` before the dashboard can show the existing organization selection UI.

## Current behavior

1. User signs in or signs up with Clerk.
2. Clerk redirects to `/dashboard`.
3. `apps/sdp-web/src/app/dashboard/layout.tsx` reads Clerk auth state:
   ```ts
   const { orgRole, orgId, userId } = await getSdpAuth();
   ```
4. The layout redirects if either `userId` or `orgId` is missing:
   ```ts
   if (!userId || !orgId) {
     redirect(await getAuthEntryPath());
   }
   ```
5. For a signed-in user with no active Clerk organization, `userId` exists but `orgId` is `null`.
6. The user is redirected back to `/sign-in`.
7. Signing in again redirects to `/dashboard`, and the loop repeats.

## Expected behavior

A signed-in user should not be sent back to sign-in just because `orgId` is missing.

Instead, `/dashboard` should render an organization setup or selection state where the user can create/select an active Clerk organization.

## Relevant code

Sign-in/sign-up both redirect to `/dashboard`:

```tsx
// apps/sdp-web/src/components/clerk-client-provider.tsx
<ClerkProvider
  signInUrl="/sign-in"
  signUpUrl="/sign-up"
  signInFallbackRedirectUrl="/dashboard"
  signUpFallbackRedirectUrl="/dashboard"
>
```

The dashboard layout currently blocks rendering when `orgId` is missing:

```tsx
// apps/sdp-web/src/app/dashboard/layout.tsx
const { orgRole, orgId, userId } = await getSdpAuth();

if (!userId || !orgId) {
  redirect(await getAuthEntryPath());
}
```

The dashboard shell already appears to contain a fallback UI for the signed-in/no-org state:

```tsx
// apps/sdp-web/src/components/dashboard-shell.tsx
if (!orgId) {
  return (
    <main>
      ...
      <OrganizationSwitcher hidePersonal />
      ...
    </main>
  );
}
```

However, this fallback is unreachable because the server layout redirects before `DashboardShell` can render.

## Proposed solution

Change the dashboard layout so it only redirects when the user is not signed in:

```ts
if (!userId) {
  redirect(await getAuthEntryPath());
}
```

This allows the signed-in/no-org state to render an organization selection or onboarding UI.

## Open question
I am happy to submit a PR, but I want to confirm whether this behavior is temporarily intended.

multipletwigs:
## Summary

Fixes #807.

A signed-in Clerk session with no active organization (fresh sign-up, org deleted, membership revoked) looped forever between `/dashboard` and `/sign-in`:

1. `dashboard/layout.tsx` treated `!orgId` the same as `!userId` and redirected to sign-in.
2. Clerk's sign-in page sees an authenticated user and immediately redirects back to `/dashboard`.

The org-selection fallback that already existed in `DashboardShell` was unreachable because the server layout redirected before the shell could render.

## Changes

- `dashboard/layout.tsx` now redirects to sign-in only when `userId` is missing. When the session has no active org it early-returns `SelectOrganizationPanel` instead — before the org-scoped project/onboarding fetches, which would only fail without an org.
- Extracted the shell's `!orgId` fallback into `components/select-organization-panel.tsx`; the shell renders the same component for mid-session org loss (`useAuth().orgId` clearing client-side).

## Why the rest of the flow already tolerates this

- `proxy.ts` only enforces signed-in (`auth.protect()`) and skips project-cookie bootstrap when there's no org.
- Every dashboard page already guards `!orgId` with `redirect("/dashboard")` (or `return null` on the home page), so the layout guard was the single remaining spot that turned the no-org state into a loop.

## Testing

- `pnpm run typecheck` (sdp-web) ✅
- `biome check` on changed files ✅
- `pnpm run test:unit` — 447 passed ✅


resourcefulmind (approved):
Clean fix. Redirecting only on `!userId` and early-returning the org panel before the org-scoped fetches breaks the loop without running calls that would just fail with no org, and pulling the fallback into a shared `SelectOrganizationPanel` for both the server layout and the client shell reads well. 
I verified the normal org case still renders the full shell and the i18n strings moved cleanly.  
Approving 🙌

PR #975, opened 2026-07-28
fix(api): fail invites with an unresolved inviter instead of writing a bad FK

tobySolutions:
`invitations.invited_by` is a NOT NULL foreign key onto `users`, but the invite handler bound `inviterUserId || "system"`. That sentinel has no user row, so it could only ever violate the constraint.

It is unreachable in practice: the Clerk middleware creates the local `users` row if it is missing and returns a non-null id, and `api_keys.created_by` is itself a NOT NULL FK onto `users`. So this is dead-code removal plus a guard, not a fix for a live failure. Worth doing because the sentinel reads as a supported fallback and previously sent a reviewer down the wrong path.

Behaviour on the working path is unchanged. If the branch is ever reached it now returns UNAUTHORIZED instead of an opaque constraint violation.

The comment on `getAnyOrgAdminClerkUserId` is reworded: legacy rows may still carry "system", so the read path keeps tolerating it, but nothing writes it anymore.

Two other items from the same ticket didn't survive investigation: the invite form is already role-gated at `settings/page.tsx:176` (withdrawn), and replacing the hand-rolled web member types is a cross-package change, so it's split out.

### DIFF
--- a/apps/sdp-api/src/routes/members/handlers.ts
+++ b/apps/sdp-api/src/routes/members/handlers.ts
@@ -112,7 +112,7 @@ async function getClerkUserId(db: DatabaseClient, userId: string): Promise<strin
  *
  * Clerk requires a `requesting_user_id` to revoke an invitation, and the more
  * specific sources can all come up empty: an API key carries no Clerk user of
- * its own, `invited_by` may be the `"system"` sentinel with no identity row,
+ * its own, legacy rows may carry a `"system"` sentinel with no identity row,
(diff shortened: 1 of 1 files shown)


tobySolutions:
Closing. On investigation this changes nothing reachable: both sources of `inviterUserId` are NOT NULL FKs onto `users` and the Clerk middleware creates the row when missing, so the `"system"` sentinel was never actually written. Not worth review time. Reasoning kept on PRO-1566.

Issue #910, opened 2026-07-24
Heal legacy Clerk identity data that surfaces template literals in asset activity


GuiBibeau:

resourcefulmind:
Closing this. Fixed in three parts: #945 stops bad values being saved, #977 shipped a migration in v0.49.1 that repaired the rows already stored, and #976 guarded the asset activity display in v0.50.0.

PR #976, opened 2026-07-28
fix(api): stop unsubstituted Clerk placeholders rendering as the actor

tobySolutions:
A display safety net, not the root-cause fix. See #977 for that.

`audit.service.ts` built the actor label as `user_name || user_email || "Team member"` with no check that either value was fit to show, so a user row holding an unsubstituted Clerk placeholder printed it verbatim in asset activity. Treats a value containing `{{...}}` as absent. Not an email-shape check, since `name` is free-form and the fault is the placeholder wherever it landed.

Keeps two failures apart rather than collapsing them: a user with nothing recorded stays "Team member", while an identity that was recorded but is unusable reads "Unknown user". Labelling both the same makes corruption look ordinary and removes the only signal an operator has, which is how this went unnoticed until a placeholder happened to be visible.

Worth having alongside #977 because it is defence at the render boundary: any future identity corruption stops at the screen rather than reaching a customer. It does not recover the identity and deliberately does not try.

Reproduced locally by seeding a user with the placeholder as their email plus an audit row attributing a deploy to them, then removed the seed data.


tobySolutions:
Relates to #910. Display half only — stops an unusable identity reaching the screen. #977 is the root-cause fix.

GuiBibeau (approved):

GuiBibeau (approved):

PR #977, opened 2026-07-28
fix(api): ignore and repair Clerk identity emails holding a template placeholder

tobySolutions:
`resolveClerkUser` resolved a user's email with `COALESCE(aui.email, u.email)`. A plain COALESCE prefers whichever copy is non-null, including one still holding an unsubstituted Clerk JWT template placeholder. The lookup short-circuits on the Clerk id and never revisits the stored value, so an affected row stays wrong indefinitely.

That email is matched literally against `invitations.email` in `ensureMembership`, so an affected user never matches their own pending invitation: they get provisioned with their Clerk role instead of the invited one, and the invitation stays pending. Silent, with no visual symptom, which is why it went unnoticed.

Reads both copies and takes the first that is actually an address, reusing the existing `isPlausibleEmail` rather than adding a second validator. Migration 0040 repairs each copy from the other, in both directions, only where the other is plausible. Rows where both are corrupt are left alone since recovering those needs a Clerk lookup, and the `users` update is guarded against its UNIQUE constraint so it cannot fail a deploy.

Verified two ways: the new test fails with the fix reverted and passes with it applied, and the migration healed a genuinely corrupted row in a local database where `auth_user_identities.email` was the placeholder while `users.email` was correct.

Worth a careful look since this is the auth path. There is no page that renders the corrupted value, so there is no UI check to go with it.


tobySolutions:
Relates to #910. This is the data half: migration 0040 repairs affected rows on deploy, and the resolver stops preferring a corrupted copy so it cannot recur.

GuiBibeau (dismissed):

resourcefulmind (dismissed):
Approving. The migration holds up where these usually don't: it's idempotent, the two statements can't feed each other (one needs `users.email` good, the other needs it corrupt), and the UNIQUE guard is sufficient rather than hopeful, since `users.email` is NOT NULL UNIQUE and the placeholder is byte-identical, so no two rows can ever collide on it. Leaving both-corrupt rows alone is right, and reusing `isPlausibleEmail` beats a second validator.

I want to bring this to your attention though:

`resolveExistingClerkContext` still has `COALESCE(aui.email, u.email, ?)` at line 78, and that's the one on the hot path: `buildClerkContext` returns early at line 382 for any established user with an org, so `ensureClerkUser` only runs on provisioning. No live bug, since the only reader is `getClerkAuth`, which is dead code repo-wide. Still worth closing, because it's wrong via the exact pattern this PR removes and it'll go live looking already-fixed. The cases surviving 0040 are both-copies-corrupt and `aui.email` NULL with `u.email` corrupt; `ensureClerkUser` already handles the first by falling through to the token email, so it's the same one-liner at 382.

The migration has no `provider = 'clerk'` filter although `resolveClerkUser` does, so it repairs every provider's identity rows. A user can hold several (unique is on `(provider, provider_user_id)`), and `UPDATE users ... FROM auth_user_identities` with several matches lets Postgres pick one arbitrarily. `AND aui.provider = 'clerk'` closes both.

All of the above is hand-checked against `1bf05da1`.

tobySolutions:
> Approving. The migration holds up where these usually don't: it's idempotent, the two statements can't feed each other (one needs `users.email` good, the other needs it corrupt), and the UNIQUE guard is sufficient rather than hopeful, since `users.email` is NOT NULL UNIQUE and the placeholder is byte-identical, so no two rows can ever collide on it. Leaving both-corrupt rows alone is right, and reusing `isPlausibleEmail` beats a second validator.
> 
> I want to bring this to your attention though:
> 
> `resolveExistingClerkContext` still has `COALESCE(aui.email, u.email, ?)` at line 78, and that's the one on the hot path: `buildClerkContext` returns early at line 382 for any established user with an org, so `ensureClerkUser` only runs on provisioning. No live bug, since the only reader is `getClerkAuth`, which is dead code repo-wide. Still worth closing, because it's wrong via the exact pattern this PR removes and it'll go live looking already-fixed. The cases surviving 0040 are both-copies-corrupt and `aui.email` NULL with `u.email` corrupt; `ensureClerkUser` already handles the first by falling through to the token email, so it's the same one-liner at 382.
> 
> The migration has no `provider = 'clerk'` filter although `resolveClerkUser` does, so it repairs every provider's identity rows. A user can hold several (unique is on `(provider, provider_user_id)`), and `UPDATE users ... FROM auth_user_identities` with several matches lets Postgres pick one arbitrarily. `AND aui.provider = 'clerk'` closes both.
> 
> All of the above is hand-checked against `1bf05da1`.

Got it sir. I made some updates to this effect. Would love another review.

resourcefulmind (approved):
Really clean fix and a great regression test, reading both copies and picking the first plausible one is the right call, and the placeholder-then-invited-role test proves it fails without the change. Went through migration 0040 carefully too: it's idempotent, the two updates hit disjoint row sets, and both-corrupt rows are left alone. On the second statement's unique-constraint safety, since the placeholder is byte-identical and `users.email` is unique, the repair can only ever match a single `users` row, so it can't self-collide on the constraint, the guard holds. Approving.

PR #982, opened 2026-07-29
feat(web): wallet policy authoring rework + allow/blocklist storage fix

multipletwigs:
Reworks the wallet policy authoring flow (PRO-1569) and fixes the allow/blocklist storage bug.

## Bugs fixed

- Toggling Allow list / Block list reused the same address list — the two lists are now stored and serialized independently and can coexist in one policy.
- Selecting a restriction category but leaving it empty could activate a no-op allow-by-default policy while the UI reported active controls — activation is now guarded by what the built payload actually contains.
- A screened-clean destination was silently dropped if you navigated away before the screening animation finished — clean addresses now commit the moment the result arrives.
- Post-screening/bulk/flagged commits landed in whichever list was active when the request resolved — async commits now pin the allow list they were screened for.
- SOL (and any well-known holding) showed its raw mint address instead of the registry symbol/name in the asset picker.
- Focusing the destination field rendered an empty dropdown shell when there was nothing to show.
- Compliance screening ran (and 403'd) for orgs without an enabled compliance provider — the policy page now resolves compliance availability server-side and the editor commits allowlist adds directly when screening isn't enabled.

## Destination payload, before → after

Before — one list, reinterpreted by the toggle (switching modes re-emitted the same addresses as the other rule kind):

```json
{
  "destinationAllowlist": [],
  "rules": [
    { "id": "blocklist-destinations", "kind": "destination", "blocklist": ["addr1", "addr2"], "action": "deny" }
  ]
}
```

After — allow and block lists are independent and both serialize:

```json
{
  "destinationAllowlist": ["addr1"],
  "rules": [
    { "id": "allowlist-destinations", "kind": "destination", "allowlist": ["addr1"], "action": "allow" },
    { "id": "blocklist-destinations", "kind": "destination", "blocklist": ["addr2"], "action": "deny" }
  ]
}
```

Round-trips back into the editor with each list intact (unit-tested).

## UI rework

- Policy flow split from one 1.9k-line file into per-capability modules (`intent-step`, `limits-assets-step`, `asset-editor`, `destination-editor` + `use-destination-editor` hook, `operation-editor`, `review-step`, `policy-summary-rail`, shared constants/`FormSection`).
- `PaymentsWizardFrame` → shared `WizardFrame` (`src/components/`) with `aside`/`toolbarActions` slots; the policy flow now uses the same shell as the payments wizards, so step spacing is identical by construction. Wizard footers dropped their one-off translucent backgrounds (they are flex siblings of the scroll region — nothing ever scrolls beneath them).
- Destinations: textarea replaced with a counterparty-search combobox with compliance screening (single add animates, clean auto-commits, risky requires hold-to-add; blocklist adds skip screening) and comma/newline bulk paste with batched screening and a per-address review queue. Orgs without a compliance provider (`ComplianceNotEnabledError`, HTTP 403) add directly.
- Operation families list what they actually gate, from the API's `enforceWalletOperationPolicy` call sites (payments = single/batch/recurring sends, ramps = quotes, issuance = mint/authority, raw signing = signer checks; `transfer`/`program`/`provider_admin` are reserved — no endpoint emits them yet).
- Empty capability sections are valid; only real input errors block (bad decimal, daily < per-transaction, invalid mint, over-long operation type).
- Review step lists actual values instead of counts; summary rail uses accordions with per-asset mint addresses and Issued/Custom badges so a spoofed symbol can't impersonate a canonical token.
- Filled field style is now the default for `Input`/`Select`/`Combobox` (the copy-pasted payments class strings are deleted); form sections render as cards.
- Perf pass per Vercel React guidelines: map-indexed lookups for accounts/assets, memoized destination parsing.

AFTER VS BEFORE
CleanShot 2026-07-29 at 10 44 53@2x

AFTER
CleanShot 2026-07-29 at 10 51 43@2x

## Notes

- Old localStorage policy drafts (pre-split `destinationText` shape) fail validation and are discarded — accepted, devnet-only drafts.
- `validatePolicyState` no longer emits `*_required` errors; `PolicyValidationErrors` gained `review: "no_restrictions"`.
- 52 policy unit tests + i18n suite green; `tsc` clean except the pre-existing `.next/dev/types` artifact; `check:i18n:strict` failure (`src/flags.ts`) pre-exists on main.


multipletwigs:
@greptile-apps are u there

arseniy-nikitochkin (approved):

PR #998, opened 2026-07-30
fix(web): point custody explorer links at the active cluster

tobySolutions:
The wallet Recent activity table built its explorer href from a devnet-pinned helper, so signatures on a production project linked to the wrong network. It now resolves the cluster from the active project.

The signer check and faucet links stay on devnet deliberately — both operations only ever run there — but they now go through the shared `explorerTxUrl` helper instead of a local copy of the URL builder, which retires the last duplicate. Added a unit test that fails if the activity link is pinned to a cluster again.

Finishes the custody half of the explorer-link migration started in #917.

---
**#1011 is stacked on this branch** and targets it rather than `main`, because both edit `wallet-actions-menu.tsx` in overlapping hunks. Merging this one first is what unblocks it — GitHub will retarget #1011 to `main` automatically. If this lands as a squash merge, #1011 may briefly show a conflict that clears once `main` is merged back into it.


multipletwigs (approved):

multipletwigs:
fyi though production isn't actually unlocked anytime soon yet 👍

tobySolutions:
> fyi though production isn't actually unlocked anytime soon yet 👍

Got it chief, no problem at all. This helps environment adaptation.

PR #1001, opened 2026-07-30
fix(web): show a new payment request's details without reopening it

tobySolutions:
Creating a payment link closed the modal, so the link was only reachable by waiting for the table to repopulate and then opening the new row again.

The create response already carries the full request (`created(c, mapPaymentRequest(row))`), so the details view now opens straight from it while the table refreshes behind. A response without a `publicToken` still means the request was created, so that case falls back to closing rather than reporting a failure that did not happen.

Not verified by click-through: the e2e organization has no custody wallet, so the wallet select has no options and the form cannot submit there. Attaching an existing wallet is not possible either — custody configs are encrypted with the owning `organizationId`. Worth a manual pass by someone with a funded sandbox wallet.


resourcefulmind (approved):
Approving. `created()` gives `{ data, meta }` and `proxyToSdpApi` streams it without `parseSdpApiResponse`, which would have unwrapped `data` and quietly broken this, so `res.data.data` really is the created request. No race in the modal swap either, the state calls batch into one commit and `router.refresh()` doesn't remount the workspace so `selected` survives.

Three non-blockers:
- The `dashboardFetch<{ data: PaymentRequest }>` cast is an assertion rather than validation, and `dashboardFetch` can return `data: null` on an empty 2xx which that type doesn't allow. Optional chaining covers it, the type's just stricter than reality.
- No regression test, and there's nowhere obvious to put one, vitest is pinned to `node` here with no interaction harness. If the envelope ever changes this silently reverts to the old behaviour with nothing failing.
- The details modal doesn't take focus when it opens, so it lands on `<body>`. True of every modal in the app, but it stings here since the payment link is the point.

Red Transactions shard wasn't yours, it passed on the merge-base and failed twice on main earlier. Re-ran it, green.

tobySolutions:
> Approving. `created()` gives `{ data, meta }` and `proxyToSdpApi` streams it without `parseSdpApiResponse`, which would have unwrapped `data` and quietly broken this, so `res.data.data` really is the created request. No race in the modal swap either, the state calls batch into one commit and `router.refresh()` doesn't remount the workspace so `selected` survives.
> 
> Three non-blockers:
> 
> * The `dashboardFetch<{ data: PaymentRequest }>` cast is an assertion rather than validation, and `dashboardFetch` can return `data: null` on an empty 2xx which that type doesn't allow. Optional chaining covers it, the type's just stricter than reality.
> * No regression test, and there's nowhere obvious to put one, vitest is pinned to `node` here with no interaction harness. If the envelope ever changes this silently reverts to the old behaviour with nothing failing.
> * The details modal doesn't take focus when it opens, so it lands on `<body>`. True of every modal in the app, but it stings here since the payment link is the point.
> 
> Red Transactions shard wasn't yours, it passed on the merge-base and failed twice on main earlier. Re-ran it, green.

Thank you very much for the detailed and thorough reviews and feedback. I will look into those fixes and sort them out as well in due time.

PR #1006, opened 2026-07-30
feat(sdp-web): standardize header and tabbed headers title

multipletwigs:
**Standardize header management**
1. if there's tabs in the header, align left
2. else align center

**Approval Page Improvement**
CleanShot 2026-07-30 at 22 01 11@2x


tobySolutions (approved):
Couple of small things on the auto refresh, non blocking.

`setLoadError` fires on background failures too, and the guard above is `hasLoadError && requests.length === 0`. So on an empty inbox a single failed poll briefly swaps in "Unable to load" with no toast, since silent hides it. Probably want silent to skip setLoadError as well.

Each tick is also two fetches plus the sdp:approval-requests-updated dispatch, which the shell answers with a third for the badge count. Might be worth a visibilityState check so it isn't polling backgrounded tabs.

Tab consolidation looks good.

multipletwigs:
thanks! will check out the policy thing in a bit, im reworking a lot of the ui

PR #1008, opened 2026-07-30
feat(web): replace the mobile drawer with a bottom bar and More sheet

tobySolutions:
Reaching any destination on a phone meant opening the slide-over first. The four places people move between most — Home, Wallets, Payments, Issuance — now sit in a thumb-reachable bar, with everything else behind **More**.

More opens a bottom sheet rather than the old sidebar: the workspace switcher first, then Manage and Workspace as two-column tiles. The payments sub-actions are deliberately absent, because the Payments tab opens a page that already carries that sub-navigation.

The top bar's drawer trigger is hidden below `xl`. Two entry points to the same destinations is worse than one, and the drawer's two unique offerings — the workspace switcher and the Settings/API Docs links — moved into the sheet.

**This changes two existing e2e specs, which is the part worth reviewing:**
- `navigation-loading.e2e.spec.ts` drove mobile navigation through the retired trigger; it now clicks the bottom bar's Wallets tab. The behaviour under test — pre-commit navigation feedback on mobile — is unchanged.
- `payments-command-center.e2e.spec.ts` checked payments sub-nav persistence by opening that drawer at 390px. That menu belongs to the persistent sidebar and only exists at `xl` and above, so the check moved to a desktop viewport. Same assertion, different viewport.

Tab links go through `DashboardNavigationLink` so they keep the same pre-commit feedback, More has its own accessible name so exactly one control is named for opening navigation, and the bar unmounts while the sheet is open rather than hiding under it. Checked at 390px: no horizontal overflow, correct active tab. Navigation loading contract passes 8/8.


multipletwigs (approved):
definitely appreciate ss's on these but it's ok

tobySolutions:
Screenshot 2026-07-30 at 19 04 07
Screenshot 2026-07-30 at 19 04 22

PR #1013, opened 2026-07-30
feat(sdp-web): standardize policy layout

multipletwigs:
Standardize all table based with pagination layout
CleanShot 2026-07-31 at 01 03 21@2x
CleanShot 2026-07-31 at 01 03 41@2x
CleanShot 2026-07-31 at 01 05 48@2x

CleanShot 2026-07-31 at 12 53 07@2x

CleanShot 2026-07-31 at 12 30 13@2x


tobySolutions:
Clean refactor. Few things:

  - summary panel is gone, those counts show nowhere else now. needs a sign off?
  - approvals footer shows rows-per-page + pagers on an empty list, policies and transactions hide theirs
  - pending count + reload float mid footer, used to sit left with the range
  - approvals table shows from lg but is still min-w-[1118px] with shadows off, scrolls with no hint
  - pageSize is local state there, URL state in policies and transactions
  - api-keys: delete the last key and both that CTA and the empty state one show till reload
  - cold load of ?tab=api_keys&page=2, does it hold? urlTab is null on the server snapshot

  Screenshot with rows?

multipletwigs:
> summary panel is gone, those counts show nowhere else now. needs a sign off?

Yeah, it's gone now, i think not really needed atm, can slot it back in again if it's needed

> approvals footer shows rows-per-page + pagers on an empty list, policies and transactions hide theirs

sg can remove

> pending count + reload float mid footer, used to sit left with the range

mmhm, i do agree it looks kinda off, will adjust to the left 

> approvals table shows from lg but is still min-w-[1118px] with shadows off, scrolls with no hint
> pageSize is local state there, URL state in policies and transactions
> api-keys: delete the last key and both that CTA and the empty state one show till reload

checking and adjusting these 

cold load of ?tab=api_keys&page=2, does it hold? urlTab is null on the server snapshot

GuiBibeau (approved):

PR #1034, opened 2026-07-31
fix(web): add French home copy for the rebuilt dashboard

tobySolutions:
French for the balance card copy #1004 adds in English.

Token and Wallet stay in English. `jeton` and `portefeuille` aren't what French practitioners say, and the locale guidance lists both as forbidden. Casing follows the rest of the catalog, which runs about 190:11 for Token and 357:22 for Wallet.

Split out from #1004 because #1022 made it a CI failure for one PR to change English and French together. Order doesn't matter, the keys just sit unused until #1004 lands. Three now-unused keys in the same block are left for a follow-up.


GuiBibeau:
this will be caught by the translation bot that has the language context

tobySolutions:
> this will be caught by the translation bot that has the language context

Understood sir. Thank you very much!

PR #1058, opened 2026-08-03
refactor(policy): extract the wallet/policy engine into @sdp/policy

multipletwigs:
## What

Pure file refactoring of the wallet/policy code — **no change to existing logic**. The rule engine moves out of `apps/sdp-api` into a new **`packages/sdp-policy`** so it is unit-testable and auditable in isolation, ahead of the PRO-1597 implementation work (dry-run, approval lifecycle, amount_window all build on this engine).

Evaluator semantics are preserved verbatim: the evaluation suite ports 1:1 to the package, and the pre-existing 650-line enforcement suite runs against the new adapter unchanged.

## Known inefficiencies (deliberately not addressed here — next PR)

- `rules: row.rules as unknown as …` cast when reading stored revisions — the read boundary is unvalidated; fix is scoped in PRO-1610 (type tightening) together with the rest of the policy-foundation cast/fallback cleanup.
- `CreateWalletOperationInput` exists as structural twins in `@sdp/policy/ports` and the repository layer; the repo should import the port type.
- `matchedRules` is spread-copied at the store boundary to satisfy the repository's loose `Record<string, unknown>[]` typing; `PolicyEvaluation.matchedRules` stays loose for persisted reads.
- Enforcement is still compensation-based (mark-failed on mid-flow errors) rather than transactional — that rewrite is PRO-1612 and now lands as a port-level change.
- Both policy systems (legacy `payment_wallet_policies` + control profiles) still run side by side; the fold is PRO-1617.

## Testing

- `@sdp/policy`: 57 tests (per-rule + fold + enforce with fake store)
- `apps/sdp-api`: full suite 1870 passed / 14 skipped
- `apps/sdp-web`: policy unit suites 52 passed; tsc + biome clean across all three


multipletwigs:
Triage of the 5 Hacktron findings — none are introduced by this PR (it moves files, so the scanner audited the whole policy surface as new code). Each will be addressed independently of this refactor:

1. **Early return skips `assertApiKeyWalletPolicyTarget` (HIGH)** — pre-existing; semantics are identical on `main`. Wallet scoping for API keys is enforced at the route/auth layer (covered by `wallet-scope-route-coverage.test.ts`); the defense-in-depth gap on the no-bindings path is folded into the PRO-1607/PRO-1612 hardening work.
2. **No role/self-approval checks on approval lifecycle (HIGH)** — pre-existing and already tracked as PRO-1607 (approval authorization: group membership, no self-approval). The PoC overstates: a readonly key fails the `wallets:write` middleware.
3. **Client-controlled `providerReference` (HIGH)** — pre-existing; nothing currently reads that reference back from a provider (provider approval sync is unwired). Will be handled in the provider-sync design when it's built.
4. **Actor spoofing via `raw_payload.actor` (HIGH)** — false positive: `createWalletOperationRawPayload` spreads the client payload first, then unconditionally overwrites `actor` with the server-derived actor from the authenticated context, which every enforcement call site supplies. A client-smuggled `actor` cannot survive the write.
5. **Compensation leaves the operation in `created` (MEDIUM)** — false positive: `failApprovalRequest` passes `operationStatus: "failed"`, and `updateApprovalRequestStatus` transitions the wallet-operation row in the same transaction. Fault-injection tests cover this path.

GuiBibeau (approved):
agreed we dix hackthron in follow up as these were already present

PR #1081, opened 2026-08-04
feat(private-channels): enable wallet-to-wallet transfers

TarekkMA:
- Allow transfers between any verified wallets belonging to members of the same private channel, including transfers between a user’s own wallets.
- Reject only transfers where the sender and recipient wallet addresses are identical.
- Return wallet-level recipient records from the API with ownership metadata and self-owned wallets listed first.
- Display custody wallet names for the user’s own recipient wallets instead of `You · <address>`.
- Exclude the selected source wallet from recipient options and clear stale selections when the source changes.
- Move the recipient wallet field to the end of the transfer form.
- Refresh the sender’s channel balance when starting a new transfer after confirmation.
- Update shared types, OpenAPI documentation, translations, repository queries, and regression tests.


apps/sdp-web/src/app/dashboard/payments/private-channels/transfer/transfer-form.tsx
```diff
@@ -51,18 +51,35 @@ function walletLabel(wallet: CustodyWalletSummary): string {
...
-    }));
-  });
+  const ownWalletLabels = new Map(
+    sourceWallets.map((wallet) => [wallet.publicKey, walletLabel(wallet)])
+  );
+
+  return recipients
+    .filter((recipient) => recipient.pubkey !== sourcePubkey)
+    .map((recipient) => {
+      const ownWalletLabel = recipient.isSelf ? ownWalletLabels.get(recipient.pubkey) : undefined;
```
nbaztec: The transfers should be done wallet to wallet, as such we should keep the same wallet names instead of owners. A wallet can be verified by multiple "owners"

nbaztec (approved):

GuiBibeau (approved):

PR #1110, opened 2026-08-04
feat(earn): rebuild the deposit flow around one strategy, no curator step

keeganthomp:
## What

Rebuilds the Earn deposit flow. The old wizard asked for a **curator** before showing anything worth buying, then made the reader hand-author per-token allocation weights on a 0.1 grid. That's replaced by:

**funding wallet → profile → filtered catalogue → one strategy → review**, then an optional API-integration hand-off and the live program.

Picking one strategy sends `pct: 100` for that strategy's stablecoin lane. Net **−1400 lines** in the module.

## Everything on screen is a field Ground actually publishes

Ground reports **no** risk tier, rating, or score anywhere — its own docs say so, and `riskMetadata.riskTier` is written only by the local dev seed. So the profiles are transparent **filters** over observable facts (redemption speed, backing, pool size), and the copy says exactly that rather than implying a rating. Profile cards and the dashboard hero show live catalogue figures, not static marketing copy.

The filter vocabulary mirrors Ground's `POST /v2/wallets/strategy/optimize` constraints so a profile could later be handed to that endpoint. It is **not** wired here: optimize returns a *blend*, which contradicts picking one strategy, and it has no SDP client/route/proxy today.

Run against the live sandbox (15 synced USDC rows), the profiles separate cleanly:

| profile | strategies | top APY |
|---|---|---|
| liquidity first | 7 | 5.9% |
| balanced | 13 | 5.9% |
| yield first | 15 | 8.2% |

Two cosmetic notes from that data: the first two tie on top APY because the best instant source beats every delayed one (count differentiates), and `instant + RWA` yields zero results because all four RWA sources settle in 1–2 days — that lands on the empty state.

## Three things with no first-class data source, handled honestly

- **"Import Fireblocks" can't mean adopting a customer's vault.** The API takes only `{provider, walletLabel}`, uses platform credentials, and rejects existing-object selectors. The step offers connect-or-select, gates on custody entitlement (Fireblocks is not entitled by default), and doesn't promise a return trip that has no plumbing.
- **The API-integration screen** keys off "the org has active API keys", resolved server-side — SDP persists no organization type. Snippets only use routes that exist; V1 has no partner deposit-signing handshake to generate.
- **The funding wallet is session state, deliberately.** It shapes the funding instructions; no API moves funds from it, so nothing implies a transfer. See the revert note below.

## Two real bugs fixed along the way

- **A double-submitted confirm fired two provider mutations.** `PUT /program` now takes a caller-minted `requestId`, forwarded on both branches. Verified against real Ground: same key + same payload replays (200, unchanged); same key + **changed** payload returns **409**. That second half is why the client keys the id to the selected strategy — reuse across a changed allocation would wedge the confirm permanently.
- **The catalogue silently truncated at 100 strategies.** The dashboard requested one page and the list endpoint has no provider filter, so past 100 active rows strategies just stopped appearing. Now paged, stopping on a short page *or* the reported total, with a hard cap.

## One thing built and then reverted

I added a `funding_wallet_id` column (migration 0050) to persist step 1's choice, and **took it back out** after asking what it actually bought:

- Nothing functional depended on it — no money moved, nothing gated, Ground never saw it. Its only consumer was preselecting the wallet on a return visit.
- The provenance argument is *worse* than what exists: Ground already reports `fromAddress` on every detected deposit — the real sending address, on-chain. The column stored stated intent before any money moved, so it could disagree with the truth.
- The payoff that would justify it — defaulting the withdraw modal's destination — isn't built.

Both CLAUDE.mds record that it was tried and reverted, and what should trigger bringing it back, so the next person doesn't rebuild it.

Also reverted: a `fund` alternative I'd added to the RWA classifier, justified by a claim that turned out false. The live catalogue emits `market | liquidity | loan | reserve | rwa | treasury`; JAAA reports `rwa` and JTRSY/USTB report `treasury`, all of which already matched. The observed values are now recorded so nobody extends it on a guess.

## Verification

- Typecheck + biome clean across `sdp-web`, `@sdp/api`, `@sdp/earn`
- **798** sdp-web tests, **43** API earn tests, **80** `@sdp/earn` tests
- i18n: 0 dangling keys, 0 unused; `messages/en/` only, so the Translation Catalog Policy passes
- Live against the Ground sandbox: catalogue reads, program read (real devnet deposit address, blended APY), a real single-strategy switch (**USDT lane preserved** — confirmed independently at Ground), the requestId replay/conflict pair above, and every rejection path refusing before any provider call
- `main` merged in (0.54.0), no conflicts

**Not verified:** nobody has driven the logged-in UI. Routes compile and serve; component behaviour is covered by render tests against real catalogue shapes. Worth a human pass before this leaves draft.

## Follow-ups deliberately not in scope

- Default the withdraw modal's destination — the trigger to reintroduce a persisted funding wallet.
- `POST /v1/wallets` accepts an untrimmed label, which is why the wallet display-name fallback had to be centralised (a whitespace label used to render an empty cell).


multipletwigs (approved):
🐑

PR #1117, opened 2026-08-05
fix(web): align wallet card skeleton button margin and update policy e2e flow

multipletwigs:
- Wallet card skeleton: Manage-button pulse wrapped in the real card's `mt-auto pt-3` so it keeps its gap above
- wallets e2e: policy page now only asserts the Revision history trigger; the Policy audit link is asserted and followed from the wallet detail panel (follow-up to #1115, pushed after it merged)


resourcefulmind (approved):
Approving. Checked the skeleton wrapper against the real card and `mt-auto pt-3` matches
`wallets-overview.tsx:212` exactly. The e2e change keeps the same assertions, just moves the audit-link check to where #1115 put the link.

PR #1120, opened 2026-08-05
feat(web): show every custody provider and what can be done about it

tobySolutions:
The choose-provider step only rendered providers already enabled for the organization. A hosted org normally saw a single card, and an org with nothing enabled hit a dead end that named no providers at all — so the rest of the catalog was invisible.

The shared catalog already carried a display-status vocabulary and a category per provider, and nothing in the UI read either. This resolves a status per provider from its connection, its entitlement and its catalog setup mode, then renders the ones that can actually be selected as cards grouped by category. Everything else — gated and not-yet-supported alike — is named in a compact group so the organization can see what exists without being offered an action that does not work.

Deliberately not included: the request-access route for gated providers, which is separate work with its own design, and the self-serve credential path, which needs the provider details step before an uninstalled provider can lead anywhere.

Scope note: this is the presentation half. The setup-status read described in the ticket's additional requirements does not exist on the API yet, so status is currently derived on the dashboard from the configs and provider-access responses.


GuiBibeau (dismissed):

tobySolutions:
reverted the trim. The version you approved is back, unchanged. I'd pushed a follow-up that pulled the fireblocks request-access button, which dismissed your approval before I realised you'd already looked at it. sorry for the churn. needs a re-approve when you get a sec.

GuiBibeau (approved):

PR #1127, opened 2026-08-05
fix(web): let workspace list cards grow so paginated tables scroll

multipletwigs:
- `DashboardWorkspaceCard` was `flex-1` + `overflow-hidden` inside the scrolling overview panel; `overflow-hidden` zeroes a flex item's automatic min-height, so the card clamped to the panel height, clipped rows past the fold, and made the paginated footer look sticky with no way to scroll (transactions, approvals, policies, policy audit)
- change the card to `shrink-0 grow` so it fills the panel when short but grows past it when rows overflow, letting the panel scroll
- add a `clamp` prop restoring the old `flex: 1 1 0%` geometry for pages that intentionally clamp the card and scroll rows in an internal overflow region; opt in the recurring payments workspace and its skeleton
- verified live on transactions/policies (panel scrolls, footer reachable) and recurring (internal region still scrolls)


resourcefulmind (approved):
Approving. I checked all eight DashboardWorkspaceCard consumers rather than the three you verified.
Approvals, policy audit and the wallet skeletons all inherit the new default, and none of them has an internal scroll region inside its card, so growing is safe for every one. Recurring really was the only page needing clamp, and its skeleton matches it.

PR #1130, opened 2026-08-05
test(web): track the recurring loader's grow classes

tobySolutions:
Main's unit suite is red right now: #1127 moved the recurring list loader from flex-1 to grow with an explicit shrink basis, and this loading-state test still expects the old classes. One line, the expectation now matches the rendered markup. Verified failing on origin/main and passing with the change.


resourcefulmind (approved):
Approving. Tracks #1127's change to DashboardWorkspaceCard, and pins a class combination
that PR explicitly called load-bearing: shrink-0 grow rather than flex-1, because
overflow-hidden zeroes a flex item's automatic min-height and a shrinkable card clamps to
the panel and clips rows. Correct assertion to have.

PR #1156, opened 2026-08-07
feat(policy): store commit messages on wallet policy revisions

multipletwigs:
**Changes**

- Support for commit messages between policy versions
- Support for Diff UI on wallet policies

CleanShot 2026-08-07 at 18 28 26@2x


multipletwigs:
Re: Hacktron finding (rule kind mutation bypass) — addressed in 63f84092:

- rule identity in the change summary is now `kind:id` — a rule resubmitted under the same id with a different kind renders as a full removal plus a full addition, never as an item diff
- added a unit test pinning the exact scenario (`asset` → `always` under the same id)
- `PUT /v1/payments/wallets/:walletId/policies` now rejects payloads with duplicate rule ids (400, `Duplicate rule id: <id>`) via a superRefine on the request schema, with an API test
- note on severity: the drawer only compares locally-authored wizard state against the active policy, so the deceptive rendering was not reachable through any current call site — fixed as hardening since the summarizer may later compare arbitrary revisions

tobySolutions (approved):

tobySolutions:
One question though @multipletwigs, does the diff show who makes what changes and all that?

multipletwigs:
@tobySolutions yeah it shows in revision history

multipletwigs:
@tobySolutions yeah it shows in revision history

tobySolutions:
> @tobySolutions yeah it shows in revision history

Oky awesome sir 🫡

PR #1163, opened 2026-08-07
fix(web): hold unrouted gated custody providers at not configured

tobySolutions:
DFNS, IBM Haven, Anchorage and Utila landed on a Request access state with no button, because only Fireblocks has a request route in the catalog. The page said access was requestable and offered no way to request it. Request access now only shows when the catalog entry actually carries a route; the rest hold at Not configured until HOO-775 wires per-provider routes. The detail page still explains that access is arranged with the SDP team, and compliance entitlements are untouched.


resourcefulmind (approved):
Approve. Moving the four unrouted providers to not_configured is the right call: request
access is a promise the page has to keep, and only Fireblocks carries a URL to keep it with.
Gating on storedCredentialSetup.mode rather than availability is the correct discriminator,
the regression tests cover both filter chips in both directions, and nothing under packages/
is touched so there is no shared-type risk.

tobySolutions:
> Approve. Moving the four unrouted providers to not_configured is the right call: request access is a promise the page has to keep, and only Fireblocks carries a URL to keep it with. Gating on storedCredentialSetup.mode rather than availability is the correct discriminator, the regression tests cover both filter chips in both directions, and nothing under packages/ is touched so there is no shared-type risk.

Awesome! Thank you very much. Great feedback and support.

PR #1207, opened 2026-08-10
feat: HOO-1018 DeepSec: enforce organization access settings and invitation binding

arseniy-nikitochkin:
## Summary

Fixes the three confirmed findings in organization access state from the DeepSec audit: an allowed-IP setting that was never enforced, lost updates on the settings column, and invitation acceptance that was not bound to the caller.

Part of **HOO-976 — Epic: Security Audits & Mainnet Readiness**.

Two commits, one per area:

| Commit | Scope |
|---|---|
| `31585397` | Allowed-IP enforcement + canonicalization, atomic settings patch, migration 0055 |
| `a11cd79d` | Invitation acceptance bound to the invited identity |

## 1. Allowed IPs were accepted, stored, documented — and never read

`settings.allowedIpAddresses` was writable through `PATCH /v1/organizations/{orgId}`, returned by the API, and described in the OpenAPI spec as restricting access. No authentication path read it. Only `api_keys.allowed_ips` was ever enforced, so the organization-wide restriction was inert for every caller.

**Enforcement** now runs in `lib/organization-ip-allowlist.ts`, called from all three auth doors:

- `middleware/auth.ts` — API keys, **in addition to** the key's own `allowedIps` (the two intersect; a key cannot widen the organization's restriction)
- `middleware/clerk-auth.ts` — dashboard JWTs
- `middleware/session-auth.ts` — dashboard session cookies

The check happens *before* the auth context is set, so a request from a disallowed origin never reaches a handler with organization identity attached. It reads Postgres per request and is deliberately uncached: a KV-cached API key would otherwise keep its previous origin authorized for up to an hour after an operator enabled the restriction. Cost is one primary-key read per authenticated request.

**Validation and canonicalization** at write time (`canonicalizeIpAllowlistEntry` in `lib/ip-allowlist.ts`). Entries were previously `z.array(z.string())` — `"office wifi"` was a valid value. Now each entry must be a real address or CIDR range, and is rewritten to the single form that means what it enforces:

| Submitted | Stored |
|---|---|
| `203.0.113.5/24` | `203.0.113.0/24` |
| `2001:0DB8:0000::0042` | `2001:db8::42` |
| `::ffff:203.0.113.5/120` | `203.0.113.0/24` |
| `203.0.113.0/24`, `203.0.113.42/24` | `203.0.113.0/24` |

`node:net`'s `BlockList` already tolerates all these variants at match time, so this is not a matching fix — it is about what an operator is shown when they review the restriction later. `203.0.113.5/24` reads as one host and authorizes 256; storing the network address makes the reviewed list and the granted access agree. The list is also capped at 100 entries.

**Self-lockout guard.** The restriction covers every authenticated path, including the endpoint that edits the restriction and the whole dashboard. A list excluding the caller's own origin would therefore be unrecoverable through the API — it would take direct database access to undo. `PATCH` now refuses such a list:

```
400  The allowed IP list must include the address this request came from
     (198.51.100.42), or it would lock the organization out.
```

A request with no resolvable client IP is refused for the same reason: enforcement fails closed without one, so a non-empty list would deny everything. The guard reuses the same `getClientIp` + `isClientIpAllowed` pair as the gate, so it cannot drift from what is actually enforced.

## 2. Concurrent settings updates silently dropped each other

`updateOrganization` read `settings`, merged in JS, and wrote the whole JSON column back with no synchronization. Two concurrent patches both merged onto the state they read, and whichever committed second overwrote the other — so an unrelated edit landing at the same moment could revert a just-installed IP allowlist.

The whole read-merge-write now runs inside a transaction holding `SELECT … FOR UPDATE` on the organization row, matching the locking pattern already used in `routes/members/handlers.ts`. Concurrent patches compose instead of racing. Provider-availability validation stays outside the transaction so no external check is performed while the row is locked.

## 3. Invitation acceptance was not bound to the caller

`POST /v1/members/accept` looked up the invitation by token hash and created a membership for whatever address the invitation named — regardless of who was authenticated. The token alone was the credential. Invitations travel by email and the acceptance link is forwardable, so any holder of a leaked token could enrol the invitee into an organization they never joined and leave them holding a token that now reads as already accepted. An API key could spend an invitation that was never its to accept.

Acceptance now requires the caller to *be* the invitee:

- API keys are refused — an API key is the organization acting on its own behalf and has no identity to be the invited person
- the caller's verified email must match `invitations.email` (case-insensitive; resolved from the Clerk context or the stored identity, both copies consulted because either can hold an unsubstituted JWT-template placeholder)
- the membership is granted to the **caller** in the organization the **invitation** names — never the organization the caller is currently scoped to, so a token cannot be redirected into another tenant
- expiry is re-tested inside the claiming `UPDATE`, so the deadline is enforced by the same statement that consumes the row
- an existing removed membership is reinstated rather than duplicated (there is no unique index on `(organization_id, user_id)`)

Single-use was already correct via the conditional `pending → accepted` claim; it is now covered by tests, including two simultaneous acceptances of one token.

## Migration 0055 — required reading

Values persisted before enforcement were **never validated and never decided an access outcome**. Applying them retroactively is how an organization loses its API and its dashboard at the same moment, with no route left to undo it.

`0055_retire_unenforced_organization_ip_allowlists.sql` renames the key on every existing row:

```
allowedIpAddresses  →  legacyAllowedIpAddresses
```

Nothing is deleted, other settings on the row survive, and enforcement ignores the renamed key. So **no organization's access changes on deploy**, whatever is currently stored. Re-applying one is a single `PATCH`, which validates the entries, stores them canonically, and refuses a list that would lock the sender out. The key is documented as read-only in the OpenAPI schema.

After this runs, the validated `PATCH` is the only writer of the enforced key — the one other writer of this column (`clerk-organization-provisioning.service.ts`) only ever writes `providerOverrides`.

Unparseable settings blobs are left untouched by the migration and read as *no restriction* at request time (logged at error level), matching how every other reader of that column already treats them. Denying over an unreadable row would lock an organization out over data nobody can repair through the API, and inducing that state requires database write access — at which point the allowlist could simply be cleared. A malformed entry inside *valid* JSON still fails closed: there a real restriction exists and cannot be evaluated.

## Acceptance criteria

- [x] Allowed IP entries are canonicalized, validated, and enforced on every intended auth path
- [x] Organization settings use an atomic patch and never lose concurrent security changes
- [x] Invitation acceptance is single-use, expires, and is bound to the invited identity and organization
- [x] IPv4/IPv6 bypass, stale-write, replay, and wrong-user acceptance tests pass

## Test evidence

**20 files / 217 tests pass** across `src/lib/ip-allowlist.test.ts`, `src/middleware`, `src/routes/organizations*`, `src/routes/members`, `src/db/migrations`. `tsc --noEmit` clean on `@sdp/api` and `@sdp/types`; biome clean; OpenAPI + docs artifacts regenerated with no diff to committed output.

New coverage:

- `routes/organizations/access-settings.test.ts` — IPv4-only list vs IPv6 client and the reverse, IPv4-mapped clients, missing client IP, key-vs-organization intersection, session path, canonical storage, dedup, entry cap, self-lockout guard, unreadable settings, retired legacy key, and two concurrency races (allowlist vs unrelated settings edit, allowlist vs rename)
- `routes/members/invitation-acceptance.test.ts` — wrong user, API key caller, cross-organization grant, replay, two simultaneous acceptances, expired, revoked, unknown token, reinstatement, name backfill
- `db/migrations/retire-unenforced-organization-ip-allowlists.migration.test.ts` — valid, invalid-entry, non-array, no-allowlist, unparseable and null rows, plus re-run safety
- `lib/ip-allowlist.test.ts` — canonicalization across host bits, RFC 5952 elision, IPv4-mapped folding, sub-`/96` prefixes, dedup

> **Note on the full suite:** the complete `sdp-api` run (~2,300 tests) did not finish locally — the shared test Postgres container dropped partway through a 35-minute sequential run and the remaining 48 files cascaded with `ECONNREFUSED` in `clearTestDatabase` teardown. Zero assertion failures; a sampled failed suite passes 83/83 in isolation. CI is the authority here.

## Reviewer notes — two deliberate decisions

1. **`POST /v1/members/accept` no longer creates user accounts.** The old handler minted a `users` row when the invited email was unknown. That is incompatible with binding acceptance to the caller — a caller who does not exist cannot be the invitee. Nothing in the repo calls this endpoint (`sdp-web` has no reference; real invites flow through Clerk's link into `clerk-auth.ts`), but it is a public documented endpoint, so the behavior change is called out here and in the OpenAPI description. `security` for the operation is now `sessionCookie` rather than `apiKeyAuth`, and `401`/`403` are documented.
2. **The approved-operation replay path is not gated.** `tryApprovedOperationReplayAuth` in `unifiedAuthMiddleware` returns before the API-key path, so a server-driven replay of an already-approved wallet operation does not re-check the organization allowlist. The original request that produced the approval did pass the check, and the replay is initiated by the server rather than by a caller at an arbitrary origin. Flagged as a scope decision rather than left to be discovered.

Also worth a look in review: the lockout guard has **no override flag**, so an allowlist cannot be configured for a network the configurer is not currently on. That is the safe default, and a `force`-style opt-out is a small follow-up if a real workflow needs remote configuration.


multipletwigs:
59796180 tightens the allowlist implementation (net −64 lines, no behavior change):

- `parseIpRange` had two sources of truth: `node:net`'s `isIP` gated validity while separate numeric parsers fed canonicalization, leaving unreachable null arms where the two could drift apart. The numeric parsers are now the single entry grammar; `isIP` remains only for the client probe in `isClientIpAllowed`.
- The zod schema now validates, canonicalizes, and dedupes in one `.transform` pass, so handlers receive the canonical list directly — `resolveSettingsPatch` and its unreachable error branch are deleted, along with the list-level `canonicalizeIpAllowlist` (dedupe lives in the schema; collapse behavior is still covered by the route tests).
- Dropped a dead `?? ""` fallback and two `as` casts.

Correctness gates run before and after the refactor:
- 469,293-check differential fuzz against ipaddr.js (validity, canonical network address, RFC 5952 formatting, v4-mapped collapse, and membership probes at range endpoints) — full agreement, harness sensitivity proven by mutation testing.
- 216,055 parity checks old-vs-new (fuzzed addresses plus character-mutated garbage) — accept-set and canonical output are byte-identical to the reviewed implementation.

multipletwigs (dismissed):

multipletwigs (approved):

PR #1208, opened 2026-08-10
fix(api): HOO-1015 Deepsec: preserve controls during partial wallet-policy updates

arseniy-nikitochkin:
## Summary

Sub-issue of HOO-976 · Security Audits & Mainnet Readiness.

This branch originally fixed the DeepSec finding where a partial wallet-policy patch reset omitted control-profile fields to permissive defaults. **That finding has since been closed structurally by #1232** ("drop legacy `payment_wallet_policies` and cut over to control profiles"), which removed `destinationAllowlist`/`maxTransferAmount`/`maxDailyAmount` and made `rules` and `defaultAction` **required** — there are no omittable control fields left to reset. The patch-semantics work is therefore dropped in the merge, along with its substrate.

What #1232 did **not** address is the ticket's concurrency criteria, and that is what this PR now delivers. On `main`, `PUT /v1/payments/wallets/:walletId/policies` has no precondition of any kind: two editors load the same policy, both PUT a full replacement, and the second silently discards the first's controls.

### 1. Stale-write precondition → 409
New optional `expectedRevisionId` on the request. The update applies only if it matches the wallet's active control-profile revision id (`null` asserts "no profile is active"); otherwise **409 CONFLICT** and nothing is written. Omitting it preserves today's unconditional-overwrite behaviour, so existing API clients are unaffected.

Because every update now activates a revision, the active revision id versions the whole policy — no separate version token or new storage is needed.

### 2. Per-wallet serialization
The update runs in one transaction that first locks the **wallet row** (`SELECT … FOR UPDATE`). `main` locks only the profile row, but a wallet with no profile yet has no row to lock — so two concurrent first-time updates could each `INSERT` a profile for the same wallet. The wallet row is what exists in both cases.

### 3. Transaction-consistent response
The control-profile summary is now read on the update transaction's own connection. `main` reads it after commit, so a concurrent update landing in that window could make the response report the *other* request's revision — which the dashboard then stores and uses as the base for its next save.

### 4. Dashboard
Both authoring actions (activate, disable) pass `expectedRevisionId` from their server-read policy, so a save built on a stale base surfaces as a visible 409 instead of clobbering another session.

## Tests

`apps/sdp-api/src/routes/payments.wallet-policy-update.test.ts` (9 tests):
- **Preconditions**: matching id applies; stale id → 409 with state unchanged; `null` with an active profile → 409; `null` with no profile → applies; unknown id with no profile → 409 and no profile created; omitted → unconditional overwrite.
- **Races**: concurrent updates serialize into ordered revisions 2 and 3 of a **single** profile, and each response reports the revision *its own* transaction activated (a post-commit read would let both echo the final one); two concurrent **first** updates create exactly one profile with two revisions; a save based on "no profile existed" is rejected once another session created one.


multipletwigs (approved):

multipletwigs (dismissed):

resourcefulmind (dismissed):
Approving. The original DeepSec finding is closed by #1232 rather than here: rules and defaultAction
are required now, so no partial body can drop a control. What is left is the concurrency work, and the
wallet-row lock plus the revision precondition hold up.

G1de0n (dismissed):

GuiBibeau:
@arseniy-nikitochkin can we get this merged soon? 2 weeks since opened

GuiBibeau (approved):

PR #1235, opened 2026-08-11
fix(api): allow tenant callers to mutate counterparty provider data

multipletwigs:
- Fixes the Sentry `TenantScopeViolationError` on `POST /v1/counterparties/:id/requirements` (`CounterpartiesRepository.mutateProviderData is system-only`)
- #1032 deny-listed `mutateProviderData` and `upsertBvnkCustomerProviderData` as system-only, but the tenant requirements flow persists provider state through them (bvnk customer/wallet/rule state, lightspark Grid customerId, mural org), breaking every requirements advance
- Removed the two methods from the system-only list; they remain tenant-bound — the proxy still rejects mismatched org/project claims and the SQL is scoped by organization_id + project_id
- Webhook-only lookups (`findActive*`, `patchMuralOrganizationById`) stay system-only
- Added a regression test: a tenant-bound repository can call both methods within its own scope
- Added full-app integration tests for the requirements advance: lightspark persists the Grid customer pointer, bvnk onramp persists the pending customer reference — both drive the tenant-bound repo factory and fail if the methods are re-deny-listed (mutation-verified)

### DIFF
--- a/apps/sdp-api/src/db/repositories/repository-factory.ts
+++ b/apps/sdp-api/src/db/repositories/repository-factory.ts
@@ -130,8 +130,6 @@ export function createCounterpartiesRepository(
       "findActiveCounterpartyById",
       "findActiveCounterpartyByBvnkCustomerReference",
       "findCounterpartyByMuralOrganizationId",
-      "mutateProviderData",
-      "upsertBvnkCustomerProviderData",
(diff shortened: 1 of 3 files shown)


tobySolutions (approved):
LGTM

PR #1243, opened 2026-08-11
feat(earn): single-vault V1 — one allocation entry per token group (PRO-1667)

keeganthomp:
## Summary

Earn V1 sells deposit and withdraw into a **single vault per deposit token** — no user-set weights, no rebalancing surface. This PR makes that the enforced contract instead of a wizard convention, so partners can no longer author portfolios the dashboard can't manage and the mainnet audit reviews only what V1 ships.

Closes [PRO-1667](https://linear.app/solana-fndn/issue/PRO-1667). Unblocks PRO-1670 (program multiplicity) and PRO-1632 (partner reference).

## What changed

**API** — `allocationGroupSchema` caps each token group at **one** entry (`.max(1)` with a clear message); the existing sum-to-100 rule then pins it to `pct: 100`. That schema bound is the only behavioral change: no handler, wire-shape, or provider-contract edits. The update branch still re-targets via `updatePortfolioStrategy` (switching vaults is a full re-target at 100%).

**Web** — the workspace holdings list no longer renders per-position share percents (`formatShare` + the `programShare` copy key removed). The cash-lane deploys/parked status, busy badges, withdraw gating, and the flat deployed-first list all stay. Deposit flow untouched — `singleStrategyAllocation` already authored the only shape the API now accepts.

**Docs** — the PUT contract docs, `packages/sdp-earn` README/CLAUDE.md, the earn data-flow doc, the pluggability playbook, and both earn module CLAUDE.mds now state the single-vault V1 contract and the dormant re-enable path. ADR 0002 gains a dated addendum and loses its stale curator-preset-weights sentence.

## Deliberately kept (dormant, for post-V1)

Weighted wire shape (arrays of `{yieldSourceId, pct}`), `weightBps` types, `updatePortfolioStrategy` contract + Ground client, the grid/sum/duplicate refinements, and the snapshot's provider-reported allocations/`pct` (read side can genuinely be multi-entry). On the API side, re-enabling weights is relaxing the one bound in `schemas.ts`; shipping weights as a product also needs the dashboard's weight authoring and share display back (see the ADR 0002 addendum).

Docs are phrased **per program** on purpose: PRO-1670 adds concurrent strategies as separate single-vault programs ("multiplicity lives between programs") on top of this cap.

## Tests

- New: two entries in one group → 400 with the cap message, provider never called (weights deliberately sum to 100 so the cap is the only violation).
- New: one usdc + one usdt entry → 201 (the cap is per token group, not per body).
- Rewritten: the sum-to-100 case now sends a lone entry at 60 (two entries would trip the cap first), keeping the dormant sum rule pinned.
- Workspace fixture reshaped to the V1 snapshot (one vault + cash-heavy, freshly funded) — the deployed-first sort test now has teeth since value ordering alone would put cash first; share-percent assertion flipped to a negative.

`@sdp/api` earn suites 40/40, sdp-web earn suites 85/85, both typechecks clean, `@sdp/api` biome clean. (Repo-root `biome check .` failures in `.github/scripts/*.mjs` and the sdp-web ESLint 10 × eslint-plugin-react crash both pre-date this branch.)


cutbow7 (dismissed):
One P2 finding; runtime behavior otherwise looks correct.

- **[P2] Weight re-enablement is not “validation-only.”** `docs/decisions/0002-earn-provider-pluggability.md:304`, `schemas.ts:53`, and `packages/sdp-earn/README.md:69` say weighted portfolios can return by only relaxing `.max(1)`. But the dashboard has no weight editor and this branch removes share visualization. Relaxing the cap alone would recreate the exact API/dashboard mismatch this PR fixes. Scope the claim to server/provider plumbing and document the required dashboard work.

keeganthomp:
@cutbow7 Fair catch on the P2 — addressed in 115bd09c. All five sites that claimed re-enablement is "relaxing the cap, nothing else" (the `schemas.ts` comment, the route CLAUDE.md, the sdp-earn README row, the ADR 0002 addendum, and the data-flow doc) now scope the validation-only statement to the API/provider surface and name the other half explicitly: the dashboard ships no weight authoring (editor removed before V1) and no share display (removed here), so weights returning as a product carries that dashboard work with it — relaxing the cap alone would recreate exactly the API/dashboard mismatch this PR removes.

cutbow7 (approved):

PR #1246, opened 2026-08-12
fix(api): HOO-1004 isolate custody wallets and configuration by tenant (HOO-1004)

arseniy-nikitochkin:
## Summary

Fixes the four confirmed DeepSec custody isolation findings (HOO-1004): Coinbase account-name collisions, cross-project wallet reuse, non-atomic scoped configuration, and wallet-scoped keys mutating wallets outside their bindings.

### 1. Coinbase CDP account-name collisions
CDP account names were derived as `sdp-<env>-<orgId>` truncated to 36 chars, and a `409 already_exists` was silently resolved by returning the existing account's address — so two projects in one org (or two orgs sharing a name prefix) could be handed the **same on-chain key**.

- Account names are now a fixed-length digest of the full tenant identity: **org + project + network + wallet slot** (plus the deployment namespace). No truncation of tenant identifiers ever decides which account a tenant receives.
- The `409 → GET by-name → reuse` path is now gated behind an explicit `reuseExisting` opt-in. The connect flow uses it with a deterministic root slot (retries after partial failure reclaim only that scope's own account); additional wallet creates use a random per-request seed and **fail closed** on collision.

### 2. Cross-project wallet reuse
- The wallet-create lifecycle now forwards `projectId` into Coinbase provisioning (it previously dropped it).
- Removed the unscoped `getWalletByPublicKey` store helper (unused; returned any tenant's wallet).

### 3. Non-atomic scoped configuration (migration `0056_custody_config_scope_integrity.sql`)
- Per-scope config uniqueness is re-created as `UNIQUE NULLS NOT DISTINCT (organization_id, project_id, provider)` — org-level rows (`project_id IS NULL`) were previously exempt under Postgres NULL semantics, allowing duplicate scope rows under concurrency.
- Legacy org-level duplicates are deduplicated in the migration: the most recently updated active row survives; scope defaults, signing requests, and wallets are repointed to it.
- `custody_configs.default_wallet_id` is now pinned to a wallet owned by the same config via a deferrable composite FK (`(id, default_wallet_id) → custody_wallets(custody_config_id, wallet_id)`), with a data repair for dangling pointers.
- The scoped config upsert is a single `INSERT … ON CONFLICT` statement; all ten provider initializers persist the encrypted payload, root wallet record, and default pointer in **one transaction** (`saveProviderConfig`) — the previous stub-config-then-real-config torn window is gone.
- Default-wallet re-point and delete-successor election are single conditional statements (no read-then-write races).

### 4. Wallet-scoped keys mutating unbound wallets
- `DELETE /v1/wallets` and `POST /v1/wallets/default-wallet` now enforce the key's wallet bindings with `wallets:write`; unbound targets are indistinguishable from unknown ones (no existence disclosure).
- `POST /v1/wallets`, `POST /v1/wallets/initialize`, and `POST /v1/wallets/switch` reject wallet-scoped keys outright — their effect (new unbound wallet / re-pointing the scope's default signer) is by definition outside any binding.
- `wallet-scope-route-coverage.test.ts` no longer allowlists these routes as non-wallet-scoped.

## Migration notes

- `0056` runs transactionally via the standard runner. The dedup step only touches pathological duplicate org-level rows; wallets are moved to the surviving config rather than deleted (a duplicate's unmovable same-`wallet_id` copy cascades away with its config row).
- Existing Coinbase accounts keep working — the DB stores the provisioned address; the new naming only affects provisioning-time identity. A tenant re-connecting after full config deletion provisions a fresh account instead of reclaiming the legacy-named one.

## Tests

- **Migration/constraints**: new `custody-config-scope-integrity.test.ts` (uniqueness incl. org-level, ownership FK, cascade behavior) and `custody-config-scope-integrity.migration.test.ts` (shadow-table run of the dedup: survivor election, reference repointing, wallet moves, dangling-default repair).
- **Concurrency**: parallel org-level and project-level upserts resolve to a single row; atomic persist + rollback of `saveProviderConfig`.
- **Collision/cross-tenant**: Coinbase name distinctness across org/project/network/namespace/seed, determinism for retries, no truncation-derived identity, reuse gated on opt-in, fail-closed otherwise.
- **Authorization**: 7 new wallet-scoped key cases (bound vs unbound delete/re-default, create/initialize/switch rejection).
- `tsc --noEmit` and `biome check` clean. (`packages/sdp-custody` standalone `tsc` has ~14 pre-existing failures on `main`, unrelated to this change.)

## Out of scope (follow-up candidates)

- A wallet-bound key with `api-keys:write` can mint a new key bound to any wallet in its org/project (`routes/api-keys/handlers.ts`).
- Earn provider wallets are intentionally org-wide across projects (`earn_provider_wallets` unique key deliberately omits `project_id`).

Closes HOO-1004.


multipletwigs (dismissed):

G1de0n (approved):
LGTM

PR #1266, opened 2026-08-12
feat(earn): ledger every money movement (PRO-1669)

keeganthomp:
Closes [PRO-1669](https://linear.app/solana-fndn/issue/PRO-1669). Builds on #1239 (PRO-1628) and #1256 (PRO-1670).

## Why

SDP records withdrawals but not deposits, so it **cannot answer "what money moved on this program" without calling the provider**. History that lives only in the provider disappears on an outage, on credential removal, or if a provider is ever offboarded — the exact reason #1239 gave withdrawals a no-provider-gate audit list. Deposits are the money-in half of the same audit trail, and PRO-1672 (balance/yield history, serialized behind this) needs net movements per period.

ADR 0002 previously forbade this on one objection: *"observing customer-initiated transfers from chain is indexer-shaped work (a stated non-goal)."* That conflated the **record** with the **observer**. SDP can't *initiate* a customer's SPL transfer; it can perfectly well *observe* one — and the provider's deposits API is an observation we already make on every dashboard poll and then throw away.

## The shape

**One table.** `earn_program_withdrawals` → `earn_program_movements` + `direction` (migration `0057`).

**Observation is source-agnostic, and that is the point.** `observed_via` is open TEXT with **no CHECK**, so the three observers are writers rather than migrations:

| | observer | status |
|---|---|---|
| now | provider-API poll (`cron/earn-deposit-sweep.ts` + the live read's side effect) | this PR |
| next | provider webhooks (Ground HMAC) | PRO-1631 |
| eventually | **SDP indexer reading Solana directly** | desired end state, no ticket |

No transition logic branches on the source; per-source difference lives only in the adapter that *builds* an observation. Two source-grep tests pin that and provider-neutrality.

**Two appliers, one table** — initiated-with-intent and observed-upsert are genuinely different machines; merging them would put `if (direction === …)` inside transition logic.

## For the reviewer: every #1239 invariant, and where it went

This touches a money table that landed days ago, so please check these off rather than rediscover them:

| #1239 invariant | Now |
|---|---|
| withdrawal can never have NULL `idempotency_fingerprint` | `withdrawal_intent_complete` CHECK **+** non-nullable fields on the withdrawal arm of the discriminated row union (the generic reads the ROW type) |
| terminal never regresses | unchanged — terminal statuses appear as no KEY in both matrices, plus the SQL `status = ANY(?)` CAS guard |
| CAS + org scope in the same WHERE | unchanged, `direction` added to the same WHERE |
| global `(provider, provider_reference)` unique | **recreated with `direction`** — only Ground *happens* to prefix `dep_`/`wd_`; a provider with one id space would otherwise collide a deposit against a withdrawal |
| ledger list takes no provider gate | unchanged, and `/movements` inherits it |
| BOLA guard on `wallet_id` | unchanged |
| money as TEXT decimal strings | unchanged |

Two invariants are **new**, expressible only now: a deposit carries no intent columns, and `requested` is withdrawal-only.

`amount_requested_usd` also relaxed to nullable: the two amount columns mean *intent* and *settled*, and writing a deposit's observed amount into a column named "requested" would be the kind of fiction the ADR's forensics rule rejects. Deposits carry their amount in `amount_paid_usd`, so `COALESCE(paid, requested)` reads correctly for both directions with no direction branch.

## Two bugs this PR's own review caught

An adversarial self-review (7 reviewers, findings verified by independent refuters) surfaced two real defects, both now fixed with tests:

1. **The applier collapsed two deposits sharing one transaction signature.** A batching payer landing two transfers to one funding address is reported as two deposits sharing one `txHash`; the signature fallback adopted the first's row, so the second either overwrote it or — if the first had settled — hit the terminal early-return and was **never inserted**. I had a test for this case, but it called `insertProgramDeposit` directly: it proved the *schema* permits two rows while the *applier* could never produce them. The rule is now "adopt unless **provably** a different movement" — reject only when a discriminator *both* sides carry disagrees, which also keeps the indexer↔poller handoff working in both directions.
2. **Period bounds compared lexicographically.** `occurred_at` stores fixed-width `.mmmZ`, and `'…00.000Z' < '…00Z'` is TRUE in Postgres, so an idiomatic `occurredTo=2026-09-01T00:00:00Z` booked the boundary movement into **August and lost it from September**. Both read paths now normalize through one shared helper with the write path. Mutation-tested: removing the fix fails the assertion.

## Verification

- **Full suite green: 223 files, 2680 tests.**
- Migration applied **from scratch** and as an **upgrade over real rows** (a settled withdrawal and a ref-less `requested` one — the crash-window shape): both survive with their ids, `direction`/`observed_via` backfilled, `occurred_at = created_at`.
- All new constraints exercised against live Postgres, including the two that must **permit**: two deposits sharing a signature, and one provider reference across both directions.
- The whole ledger suite runs against a **non-Ground** provider id (the ADR 0002 pluggability proof).
- Exit safety pinned: with `GROUND_SANDBOX_API_KEY` removed, the live reads 503 while `/movements` still serves full history.

## Deliberately out of scope

`/deposits` stays a **live** provider read (the only surface that answers "did my transfer land" in seconds) and gains a best-effort ledger write — one attempt, since the feed re-offers a missed row on the next poll. `/withdrawals` is kept and soft-deprecated in favour of `/movements`. No position or balance ledger — PRO-1672's balance-history question is *framed* in the ADR addendum (snapshots as a reconciliation artifact only, never a serving read) but deliberately left for that ticket to decide. No withdrawal sweep yet; the sweep here observes deposits only, and the route CLAUDE.md now says so.


keeganthomp (commented):
Self-authored PR, so GitHub only permits a COMMENT review. I found three blocking correctness/operational issues; on another author's PR this would be REQUEST_CHANGES. The broad test coverage and provider-neutral shape are solid, but the migration rollout, sweep progress guarantee, and accounting invariant need correction before merge.

apps/sdp-api/src/db/migrations/postgres/0057_earn_program_movements.sql
```diff
@@ -0,0 +1,310 @@
...
+-- All three write THIS row through the same applier, and no transition matrix
+-- branches on the source. `observed_via` is open TEXT with NO CHECK CONSTRAINT --
+-- deliberately unlike `status` -- because a check would make adding the third
+-- observer a migration, which is precisely the coupling this design removes.
+-- Allowed values live in the code registry (EARN_MOVEMENT_OBSERVATION_SOURCES),
+-- the same drift rule `provider` follows (ADR 0001/0002, 0055's header).
+-- The interim poller is expected to become a backstop and then be deleted;
+-- nothing downstream may assume the poller is the writer.
+
+ALTER TABLE earn_program_withdrawals RENAME TO earn_program_movements;
```
keeganthomp, line 51: [P1] Keep the previous API revision compatible with this migration

Our deploy workflows execute migrations before updating the Cloud Run service, and the release runbook explicitly requires schema compatibility across the rollback window. Renaming the live table here removes `earn_program_withdrawals` while the previous revision is still serving and still queries that name, so withdrawal create/detail/list begin failing as soon as 0057 commits; if the new rollout fails, the documented image rollback remains unusable. Ship this as an expand/contract change (for example, retain the old table name or a fully writable compatibility surface for one release, then rename after the rollback window) and test the previous image against the migrated schema.

apps/sdp-api/src/cron/earn-deposit-sweep.ts
```diff
@@ -0,0 +1,462 @@
...
+        );
+      }
+    }
+
+    if (page.nextCursor === null) {
+      // A complete walk: start from the head again next pass.
+      await cache.delete(cursorKey);
+      return;
+    }
+    cursor = page.nextCursor;
```
keeganthomp, line 507: [P1] Persist forward progress before fetching the next page

When page 1 succeeds and page 2 throws, `cursor` has advanced in memory at this line but `cache.put` is reached only after all 20 pages. The wallet-level catch swallows the error, so the next tick has no persisted cursor and starts at page 1 again; a deterministic failure on page 2 permanently prevents every later deposit from entering the audit ledger. Persist each successfully completed page's `nextCursor` (then delete it on a complete walk), or save the latest cursor in the failure path; the current "keeps a cursor minted during this pass" test should assert the put, not merely the absence of a delete.

apps/sdp-api/src/db/repositories/earn.repository.postgres.ts
```diff
@@ -471,17 +557,297 @@ export function createPostgresEarnRepository(db: AppDb): EarnRepository {
...
+      // "money is TEXT decimal strings" storage rule — it preserves exact provider
+      // strings on disk while summing them exactly here, and it throws loudly on
+      // malformed text, which is the correct outcome when the only writers are our
+      // own mappers.
+      const page = await db
+        .prepare(
+          `SELECT direction,
+                  token,
+                  COUNT(*)::int AS movement_count,
+                  COALESCE(SUM(COALESCE(amount_paid_usd, amount_requested_usd)::numeric), 0)::text
```
keeganthomp: [P1] Sum the wallet outflow, not only the payout amount

This is documented as the movement half of `delta balance - net movements = earnings`, but the expression omits `fee_usd` even though Ground's withdrawal preview reduces `totalUsdAfterWithdrawal` by the requested amount plus the fee (the provider fixture goes from 1000 to 949.9 for 50 + 0.1). Every such fee is therefore misclassified as negative earnings. The fallback also books the full request for a `partially_completed` row whose optional `amountPaidUsd` is absent, overstating that outflow. Define the settled wallet-outflow invariant first (including fees and the required handling for missing partial paid amounts), then use that same definition here and in the movement projection.

apps/sdp-api/src/db/migrations/postgres/0057_earn_program_movements.sql
```diff
@@ -0,0 +1,310 @@
...
+    -- while neither direction carries a figure it never had.
+    ADD CONSTRAINT earn_program_movements_deposit_amount_present
+        CHECK (direction <> 'deposit' OR amount_paid_usd IS NOT NULL),
+
+    -- The mirror, so the shape is pinned from both sides: a deposit may not carry
+    -- intent columns at all. Without it, a bug could write a deposit with a
+    -- request_id and collide with a withdrawal on the intent unique.
+    ADD CONSTRAINT earn_program_movements_deposit_has_no_intent
+        CHECK (direction <> 'deposit'
+               OR (request_id IS NULL AND idempotency_fingerprint IS NULL)),
```
keeganthomp: [P2] Enforce every deposit no-intent column

The advertised invariant is that a deposit may not carry intent or actor columns, but this CHECK covers only `request_id` and `idempotency_fingerprint`. A deposit can still be written with `project_id`, `amount_requested_usd`, `destination_address`, `created_by`, or `initiated_by_key_id`; `mapProgramDepositRow` then overwrites several of those raw values with null and hides the corruption from callers. Extend the constraint to every direction-only intent/actor field and add a raw insert/update rejection test; the current repository test only proves that this writer happens to omit them.

apps/sdp-api/src/routes/earn/handlers/program.ts
```diff
@@ -910,3 +969,75 @@ export const listEarnProgramWithdrawals = async (c: AppContext) => {
...
+    status: row.status,
+    amountUsd: row.amount_paid_usd ?? row.amount_requested_usd ?? "0",
+    ...(row.amount_requested_usd !== null && { amountRequestedUsd: row.amount_requested_usd }),
+    feeUsd: row.fee_usd ?? undefined,
+    token: row.token,
+    counterpartyAddress: counterpartyAddress ?? undefined,
+    transactionSignature: row.transaction_signature ?? undefined,
+    providerReference: row.provider_reference ?? undefined,
+    failureReason: row.failure_reason ?? undefined,
+    observedVia: row.observed_via,
```
keeganthomp, line 1000: [P2] Give observedVia one meaning across both directions

The wire contract says this is the mechanism that reported the row's current state, and deposit updates overwrite it on every observation. Withdrawal observations never update `observed_via`, though, so a provider response can advance `requested` to `processing` or `completed` while `/movements` still returns `sdp_intent`. That makes the field mean latest observer for deposits but original initiator for withdrawals. Either pass an observation source through both withdrawal appliers and their CAS update, or make the field immutable discovery provenance for both directions and document it that way.

apps/sdp-api/src/db/migrations/postgres/0057_earn_program_movements.sql
keeganthomp: Fixed in 8a6baf5f — you're right, and the runbook is explicit about it ("avoid deleting or renaming data that the previous release reads").

I'd previously talked myself out of this on the grounds that everything behind the rename is gated on `isEarnEnabled` (both flags default off), so the old revision would never issue SQL against the table. That's not good enough: whether it touches the table is a *runtime* fact this migration can't assert, PRO-1655 exists precisely to run Earn work in a deployed environment, and the runbook doesn't offer a flag exemption.

The expand half now ships here:

- a **writable compatibility view** under the old name, presenting exactly 0055's column list over `direction = 'withdrawal'` rows. It's a simple single-table view (no aggregate/DISTINCT/GROUP BY), so PostgreSQL makes it auto-updatable — the old revision's SELECT/INSERT/UPDATE all keep working with no trigger machinery to get wrong — plus `WITH CHECK OPTION` so a write through it can't create a row it could never read back.
- `direction`, `occurred_at` and `observed_via` keep their **DEFAULTS** rather than dropping them, because that revision's INSERT doesn't mention those columns. The defaults are exactly right for an initiated movement: `withdrawal`, `sdp_intent`, and now (intent *is* when the movement began).

Contract half is a later migration once the rollback window closes: drop the view, drop the three defaults. Nothing in the new revision reads the view.

Tested as you asked — the previous revision's own INSERT (0055 shape), replay SELECT and guarded CAS UPDATE all driven through the view against a migrated database, with a deposit confirmed invisible to it. Pinned as a repository test, so a future change that breaks the compatibility surface fails rather than silently shipping.

apps/sdp-api/src/cron/earn-deposit-sweep.ts
keeganthomp: Fixed in 8a6baf5f. This was the worse half of the bug I'd already been chasing in that function, and I missed it: I added cursor *invalidation* on failure but left the `put` outside the loop, so page 1's progress was never durable. A deterministic page-2 failure meant every later tick restarted at page 1 and died in the same place — no deposit past page 1 could ever reach the ledger, silently, behind a healthy check-in.

Now persisted after **every** completed page, still deleted on a complete walk, and the two failure modes stay distinct: a cursor carried over from a previous pass is discarded when the provider rejects it, while one minted during this pass is kept because it represents real progress.

You were also right about the test asserting the wrong thing — absence-of-delete would have passed with no `put` at all. It now asserts the `put`, and there's a second test walking three pages that pins the exact sequence of persisted cursors.

apps/sdp-api/src/db/repositories/earn.repository.postgres.ts
keeganthomp: Fixed in 8a6baf5f. Both halves were real.

The settled wallet-flow definition is now stated once, in that query's comment, and it's what the aggregate implements:

- `total_usd` sums `amount_paid_usd` **alone** — no fallback to the requested figure. For a `partially_completed` row the request is precisely the amount that did *not* move, so the old fallback overstated exactly the rows most likely to be wrong.
- `total_fee_usd` is summed **separately** and is real wallet outflow: a withdrawal reduces the wallet by paid + fee, so omitting fees booked every one of them as negative earnings.
- a settled row with no paid amount is **unknown**, not zero and not the request. Those are excluded from both sums and surfaced as `unknownAmountCount`, so a caller can tell a period is incomplete instead of receiving a confidently wrong total.

On the projection: I deliberately did *not* fold the fee into the DTO's `amountUsd`, because `feeUsd` is already a separate wire field and adding it in would double-count for any consumer that reads both. The DTO's doc now points at this aggregate as the authoritative definition for period math.

Covered by a test that nets both directions and both tokens with a fee on each settled withdrawal, so dropping the fee term fails.

apps/sdp-api/src/db/migrations/postgres/0057_earn_program_movements.sql
keeganthomp: Fixed in 8a6baf5f. The CHECK now covers all seven: `request_id`, `idempotency_fingerprint`, `project_id`, `amount_requested_usd`, `destination_address`, `created_by`, `initiated_by_key_id`.

Your point about `mapProgramDepositRow` is the part that makes this more than tidiness — it hard-codes those fields to null on read, so a corrupt row wouldn't just exist, it would be actively *hidden* from every caller and from any forensic read that went through the repository.

Added a raw-SQL rejection test that loops the seven columns and asserts each one is rejected, deliberately bypassing the writer — as you noted, the previous test only proved that this particular writer happens to omit them.

apps/sdp-api/src/routes/earn/handlers/program.ts
keeganthomp: Fixed in 8a6baf5f. Took your first option: both withdrawal appliers now stamp an observation source, so `observed_via` has one meaning for both directions — the mechanism that reported the row's **current** state. `sdp_intent` now survives only on a row nothing has observed yet, i.e. the crash-window row whose provider call never came back.

Chose that over immutable discovery provenance because the field's job is to tell an operator whether a newly-added observer is actually driving rows during the poll → webhook → indexer migration, and discovery is already preserved in `provider_data.discoveredVia`.

One thing worth flagging from writing the test: the create path applies the provider's acceptance response immediately, so a freshly created withdrawal already reads `provider_poll`, not `sdp_intent`. That's correct — the provider did report that state — but it surprised me, so both the wire-type doc and the route CLAUDE.md now say so explicitly.

apps/sdp-api/src/cron/earn-deposit-sweep.ts
```diff
@@ -0,0 +1,465 @@
...
+    // The scan reached the end of the collection, so the next pass starts over at
+    // the head. Clearing here (rather than only on an empty page) is what keeps a
+    // platform with fewer programs than the cap from carrying a stale cursor.
+    await cache.delete(scanKey);
+  } else {
+    const last = wallets[wallets.length - 1];
+    if (last) {
+      await cache.put(scanKey, encodeScanCursor(last));
+    }
+  }
```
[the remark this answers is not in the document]
keeganthomp: Done in b37f4ded — you're right, and I'd flagged this risk in the design then talked myself out of it.

The checkpoint now advances only AFTER the batch is processed. Checkpointing on fetch looked equivalent and wasn't: an interrupted pass (deadline, the job's 120s execution cap, or a kill) left the cursor pointing past wallets nothing had swept, and those programs were skipped until the environment scan wrapped — stale audit history for several cadence windows, with a healthy check-in the whole time.

Advancing after the loop inverts the failure: an interrupted pass re-walks the batch. That costs nothing, because every observation is idempotent and a re-observed terminal row is one indexed SELECT and zero writes — so replay is strictly better than a silent skip.

Pinned two ways: an ordering assertion (every observation precedes the checkpoint write) and a deadline test that cuts a 200-wallet batch short and asserts no checkpoint was written. Mutation-checked — moving the write back before the loop fails it.

apps/sdp-api/src/cron/earn-deposit-sweep.ts
```diff
@@ -0,0 +1,465 @@
...
+      // the cursor is only deleted on a complete walk, so every later pass would
+      // replay the identical failing request while the sweep still reports a
+      // healthy check-in. Drop it so the failure costs one pass, not all of them.
+      if (resuming && skippableCode(error) === undefined) {
+        await cache.delete(cursorKey);
+        getLogger().warn(
+          { walletId: wallet.id, provider: wallet.provider },
+          "sweepEarnDeposits: discarded a rejected resume cursor; next pass restarts from the head"
+        );
+      }
```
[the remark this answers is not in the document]
keeganthomp: Done in b37f4ded — correct, and this was a flaw in a fix I made one round earlier. A human reviewer asked me to stop a rejected cursor wedging a wallet forever; I over-applied it to every non-skippable error.

Narrowed to a genuine rejection. The provider fetch layer classifies 409 → CONFLICT, 429 → RATE_LIMITED, 5xx → PROVIDER_UNAVAILABLE, leaving BAD_REQUEST as the only class meaning 'the provider read this request and refused it' — the one that is actually evidence about the cursor we sent. A raw transport error isn't an SdpEarnError at all and is transient by definition.

So RATE_LIMITED, PROVIDER_UNAVAILABLE and transport failures now KEEP the cursor, which is what matters for exactly the case you named: a wallet with more history than the per-pass cap would otherwise restart at the feed head on every blip and never reach the end. Covered by a table-driven test over those three failure kinds, alongside the existing BAD_REQUEST case that still discards.

apps/sdp-api/src/cron/earn-deposit-sweep.ts
```diff
@@ -0,0 +1,500 @@
...
+    // Reached the end of the collection: next pass starts over at the head.
+    // Clearing here (rather than only on an empty page) is what keeps a platform
+    // with fewer programs than the cap from carrying a stale cursor.
+    await cache.delete(scanKey);
+    return;
+  }
+  const last = wallets[wallets.length - 1];
+  if (last) {
+    await cache.put(scanKey, encodeScanCursor(last));
+  }
```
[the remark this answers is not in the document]
keeganthomp: Done in 095f52cf — real gap, though the fix isn't the obvious one.

One correction on severity first: it's not indefinite. Below MAX_WALLETS_PER_RUN (200) programs the scan comes back under the cap, clears its cursor, and every pass starts at the head — so a failed wallet is retried on the next tick. Above the cap the wait is bounded by programs/200 passes. Real, but proportional.

The obvious fix — hold the checkpoint back at the first failure — is worse than the bug. One persistently broken wallet mid-batch would then starve every wallet after it, forever, converting one stale wallet into total stall for the tail of the scan.

So the scan still always advances, and each failed wallet gets ONE immediate second attempt before the checkpoint moves. That recovers what actually causes these failures — a blip, a 429, a brief 5xx — seconds later instead of cadence windows later, and per-page cursor persistence means the retry RESUMES the walk rather than restarting it. A wallet that fails twice is a genuine failure: logged, counted, and left for the wrap.

Two details worth knowing: `failures` now counts only wallets that failed BOTH attempts (with `walletsRetried` alongside), so it's the number an operator should act on; and a steady-state skip — no credentials, no portfolio capability — is deliberately NOT retried, or an un-credentialed environment would double its log volume on exactly the pass someone is reading.

Three tests: transient failure recovered in-pass, double failure counted once with the healthy wallet untouched, and a steady-state skip attempted only once. The bound and the reasoning are now in the module header.

apps/sdp-api/src/cron/earn-deposit-sweep.ts
```diff
@@ -0,0 +1,573 @@
...
+ * Whether the provider rejected the CURSOR, as opposed to failing the request.
+ *
+ * `classifyProviderStatus` maps 409 to CONFLICT, 429 to RATE_LIMITED and 5xx to
+ * PROVIDER_UNAVAILABLE, leaving BAD_REQUEST as "the provider read this request and
+ * refused it" — the only class that is evidence about the cursor we sent. A raw
+ * transport error is not an SdpEarnError at all and is transient by definition.
+ */
+function isRejectedCursor(error: unknown): boolean {
+  return error instanceof SdpEarnError && error.code === "BAD_REQUEST";
+}
```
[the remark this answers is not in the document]
keeganthomp: Done in 22b8dbe3 — correct, and this is the second time this check has been too broad. Worth naming the root cause rather than just the symptom: I was inferring an HTTP fact from our own normalized error code, and `classifyProviderStatus` deliberately throws that fact away — it maps every non-409/429 4xx onto BAD_REQUEST, so a malformed cursor, an expired key, a revoked scope and a deleted wallet are indistinguishable at that layer.

Fixed by using the status itself. The provider fetch layer attaches `providerStatus` on every non-ok response and documents it as its own fact that a per-provider normalizer may not overwrite, so it is the reliable signal. Only 400 and 422 — 'read your request, rejected its input' — now discard the cursor.

One judgement call worth flagging: an error carrying NO provider status (constructed locally, or a raw transport failure that isn't an SdpEarnError at all) is treated as *not* the cursor's fault. The asymmetry is deliberate — keeping a possibly-bad cursor costs one wedged pass that its own TTL clears, whereas discarding a good one silently loses observed history, and silent loss is the failure this ledger exists to prevent.

Five tests, each mutation-checked against the old code-based check: 400 still discards; 401, 403, 404 and a status-less error all keep the cursor. Reverting to `code === 'BAD_REQUEST'` fails four of them.

PR #1281, opened 2026-08-13
fix(cron): log every underlying cause of a failed reconciliation tick

G1de0n:
The reconciliation tick runs its tasks concurrently and collects their failures,
so when more than one fails the job rejects with an `AggregateError`. The handler
logs only the top-level error, and the standard serializer does not walk the
aggregate's list of causes.

The result is a log line saying `pending-transfers tick had multiple failures`
with no indication of which task failed or why — the individual causes are
dropped. Dev hit this three times in the last day, each time leaving nothing
actionable behind.

`describeCronFailure` now unwraps the aggregate and records each cause alongside
the original error. A single failure keeps its existing shape, and a rejected
non-error value is still described rather than dropped.

## Verification

- `job` suite: 14 passed, including three cases covering aggregate, single and
  non-error rejections
- typecheck and lint clean


resourcefulmind (approved):
I checked this only changes the log line and not the behaviour: runCronJob is untouched and process.exit(1) is still there, so nothing that used to fail the job now passes quietly instead.
I Approve

PR #1290, opened 2026-08-13
feat(helius): HMAC-signed router and the timeout ladder

snowmead:
## What

Puts all eight flow endpoints behind **per-caller** HMAC verification, runs the boundary validation from #1289 inside the extractor so no handler can skip it, and gives each route a bounded budget. Handlers parse and validate, then return `501 NOT_IMPLEMENTED` with the real error body — **routed but unimplemented, so a `404` in a test means a routing mistake rather than unfinished work.** Also adds the two endpoints that were specified but missing: the merge opt-in and the nullifier-status read.

## Security invariant: one secret per caller, never "either"

The gateway serves two callers with different privileges, so the route table is also the caller table:

| Caller | Routes |
| --- | --- |
| `sdp-api` | `register`, `merging`, `shield`, `assemble`, `nullifiers/status` |
| `rings-key-auth` | `sync`, `plan`, `prove` |
| unauthenticated | `health` |

Two middleware entry points, `verify_sdp_api` and `verify_key_auth`, over one shared `verify(secret, …)`. **A key-bearing route signed with the sdp-api secret is `401`**, and vice versa — `each_route_rejects_the_other_callers_secret` asserts that for all eight, in both directions. The key-authority design states that this service holds no route *to* the key authority; making the reverse true costs one config entry.

Scoped honestly: in production Cloud Run IAM already separates the two service accounts and rejects unauthenticated callers before our code runs, so this is defence in depth there. It is the **only** such control in docker-compose, self-hosted deployments and integration tests — the same gap that makes HMAC worth having at all.

Everything else about the scheme is byte-identical: HMAC-SHA256 over decimal-seconds timestamp concatenated with the raw body, lowercase hex, 300 s window, constant-time compare. It has to keep matching `createHmacSignature` in `packages/sdp-payments/src/fee-payment/kora.adapter.ts`, and now the key authority's Rust equivalent.

**Explicitly not doing a nested or origin signature.** The key authority mutates the body, so `sdp-api`'s signature over the body *it* composed cannot be re-derived here from the final bytes — JSON re-serialization is not byte-stable and this service cannot know what was added. The property such a scheme would reach for is proof that `sdp-api` authored the action, and the key authority already provides it: `sdp-api` presents a single-use intent token naming the exact operation, and the key authority independently verifies the operation exists, is approved, and is in a provable state *before* unwrapping. This gateway sits downstream of that decision with no authority to refuse on it, so re-verifying the token here would add a dependency without adding a control.

## Validation is in the extractor's bound, not in the handlers

`ValidatedJson<T>` now requires `T: DeserializeOwned + Validate`, and calls `validate()` on the success arm. Every request type implements `Validate`, so the bound is total — a new request type cannot be routed without deciding what its validation is, and the compiler enforces it rather than a review checklist.

Both failure kinds share one error path deliberately. It already logs the serde detail and returns only an `x-request-id` pointer, precisely because these bodies carry viewing and nullifier keys and serde would otherwise quote the offending value straight back.

`a_validation_error_never_echoes_the_request_body` **moved off register onto `/v1/wallets/sync`**. Register no longer carries secrets, so asserting the no-echo property there would prove nothing; it belongs on a key-bearing route.

## Two endpoints that were missing

**`POST /v1/wallets/merging`** — a live defect in the contract, independent of the key authority. `UserRecord.merging_enabled` defaults to **false** (`RegisterData` has no such field, so registration never sets it) and the shielded-pool program rejects `merge_transact` while it is. Merge is documented as the recovery path for `UNSUPPORTED_SHAPE` — the case where a wallet's notes are too fragmented for one transfer — so **on every freshly registered wallet that recovery path fails on chain.** Flipping the flag needs `set_merging_enabled`, which requires the owner to sign, so it is an `sdp-api` plus custody flow that needs a gateway endpoint only because SDP does not encode zolana instructions. The merge *transaction* still carries no custody signature; this is a one-time prerequisite, recorded as a prerequisite rather than as a weakening of that claim.

**`POST /v1/nullifiers/status`** — the reconciliation probe, keyless and pure. Its one rule, in the module docs: **an indexer error must never be reported as `seen: false`.** A false negative tells SDP the operation did not land, so it retries, re-syncs, selects different notes, produces disjoint nullifiers, and the recipient is paid twice. A false positive only stalls one operation into manual reconciliation. That asymmetry is also why `signature` is nullable while `seen` is not: a spend the indexer knows about but cannot attribute still correctly forbids the retry.

## The timeout ladder

Unchanged in substance, now covering both new routes. `merging` takes `RPC_TIMEOUT` (35 s, one registry read, same shape as register); `nullifiers/status` borrows `SYNC_TIMEOUT` (60 s, indexer-bound) rather than introducing a constant the ladder would have to order — a dedicated budget can come with the handler, once there is something to measure.

**Nothing upstream bounds a prove call.** The SDK's `PROVE_REQUEST_TIMEOUT_SECS = 600` reads like a ceiling but is a timeout on one HTTP request: the submit retries three times, transfer and merge proofs are queued, and the poll ceiling counts only time spent *sleeping* between polls — 1200/3 is 400 polls of up to 600 s each. So every budget here is one this service chooses, and they must increase outward because only the innermost bound that fires can answer with a structured `ErrorBody`. A `const` block asserts that ordering at build time, including against the agave client's own 30 s per-request timeout.

- **Every handler still returns `501`.** The wire contract is exercised on the way in — each one parses and validates — but nothing behind them is built.
- **660 s derives from an unmeasured 600 s placeholder.** Replace with p99 of a cold transfer prove and a cold merge prove against the devnet prover, then re-derive the rows above it.
- **Layer 3 is unreachable in production until sdp-infra sets `--timeout=700`.** Cloud Run's 300 s default truncates every prove first. The `const` block pins this crate against the recorded intent; it cannot check whether the flag was applied.
- **The `/plan` merge check is not here.** `/plan` with `kind: Merge` should read `merging_enabled` from the record it already fetches and fail fast, because discovering it after a prove call is expensive and the proof cannot land. That needs the handler.
- **Layer timeouts are a backstop, never the plan.** The async Photon client is built with no HTTP timeout at all, so `/sync` and `/plan` will need their own `tokio::time::timeout`.
- **The handler will have to classify the SDK's own timeout as `PROVER_TIMEOUT`.** It returns a generic `ClientError::ProverServer`, whose natural mapping is `PROVER_FAILED` with `retryable: false` — which would tell SDP never to retry a prove that merely timed out.
- **14 of 19 `GatewayError` variants are unconstructed**, hence the module-level `allow(dead_code)` in #1289. They are the contract SDP codes against, so they land with it rather than trickling in per handler.

## Testing

```
cargo fmt --check
cargo test --locked          # 22 unit + 16 contract tests, all passing
```

Contract tests go through the real router via `oneshot`, so middleware ordering is part of what is asserted. New in this PR: all eight routes reachable and `501`/`400` (never `404`), every route rejecting the other caller's secret, the swapped 31/32 key widths rejected as `INVALID_REQUEST` with the field named and no base64 echoed, out-of-order viewing-key indices rejected, a wrong-width `viewingPubkey` rejected with the expected width named, an empty nullifier list rejected, and the merge opt-in and nullifier-status routes reachable and keyless. `SIGNED_ENDPOINTS` is now `(path, caller)` pairs and exhaustive, so a new route cannot be added without deciding who may sign it.

### DIFF
--- a/apps/sdp-helius-gateway/README.md
+++ b/apps/sdp-helius-gateway/README.md
@@ -38,27 +38,30 @@ Two consequences shape this crate:
 
 ## Status
 
-`GET /health` is the only route wired up so far. It answers unauthenticated, because a liveness probe
-carries no shared secret and requiring one would make an HMAC misconfiguration look like a dead
-container instead of failing requests with a clear cause.
+`/health` is implemented. All eight flow endpoints are routed and return
+`501 NOT_IMPLEMENTED` with the real error body, so a `404` in a test means a routing mistake rather than
+unfinished work. Each one parses and validates its request first, which means the contract is exercised
+before the flow behind it exists.
 
-The full contract is already here as types in [`src/wire/`](src/wire/), ahead of the handlers that will
-consume it. That ordering is deliberate: the contract is the thing SDP builds against and the thing
-worth reviewing first, and `wire/mod.rs` carries a module-wide `allow(dead_code)` for exactly as long
-as nothing reads it.
+`/health` answers unauthenticated, because a liveness probe carries no shared secret and requiring one
+would make an HMAC misconfiguration look like a dead container instead of failing requests with a clear
+cause.
 
-Eight endpoints are specified and land behind this one, five of which never see a secret:
+| Endpoint | Caller | Key material | Status |
+| --- | --- | --- | --- |
+| `POST /v1/wallets/register` | `sdp-api` | no | 501 |
+| `POST /v1/wallets/merging` | `sdp-api` | no | 501 |
+| `POST /v1/operations/shield` | `sdp-api` | no | 501 |
+| `POST /v1/transactions/assemble` | `sdp-api` | no | 501 |
+| `POST /v1/nullifiers/status` | `sdp-api` | no | 501 |
+| `POST /v1/wallets/sync` | `rings-key-auth` | yes | 501 |
+| `POST /v1/operations/plan` | `rings-key-auth` | yes | 501 |
+| `POST /v1/operations/prove` | `rings-key-auth` | yes | 501 |
+| `GET /health` | none | no | **200** |
 
-| Endpoint | Caller | Key material |
-| --- | --- | --- |
-| `POST /v1/wallets/register` | `sdp-api` | no |
-| `POST /v1/wallets/merging` | `sdp-api` | no |
-| `POST /v1/operations/shield` | `sdp-api` | no |
-| `POST /v1/transactions/assemble` | `sdp-api` | no |
-| `POST /v1/nullifiers/status` | `sdp-api` | no |
-| `POST /v1/wallets/sync` | `rings-key-auth` | yes |
-| `POST /v1/operations/plan` | `rings-key-auth` | yes |
-| `POST /v1/operations/prove` | `rings-key-auth` | yes |
+Five of the eight never see a secret. The caller column is enforced, not documentation: each route is
+verified against its own caller's HMAC secret, so a key-bearing route signed with the sdp-api secret is
+`401`.
 
 Registration is keyless because a `ShieldedAddress` is built entirely from public halves — the owner's
 Ed25519 pubkey, the 32-byte nullifier pubkey and the 33-byte compressed P256 viewing pubkey — so no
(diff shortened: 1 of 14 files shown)


nbaztec (dismissed):

TarekkMA (dismissed):

snowmead:

PR #1291, opened 2026-08-13
feat(helius): WalletAuthority that holds no signing key

snowmead:
## What

Implements `WalletAuthority` for a `RequestAuthority` that holds **no Ed25519 signing key at all**, and proves it compiles and runs that way. This is the security argument for running Rings as a sidecar, so it belongs in code the compiler and a test check rather than in a design document that asserts it.

Nothing constructs this yet — the module carries `allow(dead_code)` until the flow handlers do. It is here as a standalone layer because the question it answers ("does a prove-only custodian need spend authority?") gates the whole architecture, and it is cheaper to settle now than after handlers depend on the answer.

## Security invariant: proving and spending are separate gates

A shielded transaction clears two gates. **Entitlement** is the Groth16 circuit, which needs the nullifier key — held here, per request, never stored. **Authorization** is the Solana runtime, which needs an Ed25519 signature on the outer transaction — held by `@sdp/custody` and Kora, never here.

`ShieldedAddress` has public fields and needs only `PublicKey::from_ed25519` — a *public* key — so the shielded identity is derived here, never transported. `sign_p256` would sign with the viewing key, a P256 secret this service already holds. So the authority can decrypt notes, derive nullifiers and produce the P256 signatures the protocol wants, and still cannot move funds.

The circuit commits *which* addresses must sign; it does not verify a signature in-circuit. That is what makes the gates separable, and it means **a fully compromised gateway is total privacy loss and zero fund movement.**

## The test that makes keyless registration safe

`public_halves_derive_the_same_shielded_address_as_the_secrets` builds a `ShieldedAddress` the way #1289's register handler will — from the owner's Ed25519 pubkey plus the two public key halves, holding no secret — and asserts it is byte-identical to the one this authority derives from the corresponding secrets.

Without that property, a keyless registration could bind a wallet to an identity its own key material does not describe, and the mismatch would only surface later as notes nobody can decrypt. It needs no handler and no RPC, so it is worth having now rather than alongside the implementation it protects.

## Viewing keys are held index-labelled

`viewing_keys: Vec<(u32, ViewingKey)>`. The label is the key authority's generation number, which is what a stored projection's counters refer to — not a position in this vector. Resolution is by label (`viewing_key_at`), and `current_viewing_key` uses `max_by_key` rather than `.last()`.

With the indices validated as strictly ascending at the request boundary the two agree, so this is not a behaviour change. It is a statement about *what defines the newest key* — its generation number, not where it happens to sit — so that a future change to the ordering guarantee cannot silently start picking the wrong key. Three tests cover it: non-contiguous generations (0, 3, 7) resolve correctly, position 2 does **not** answer to generation 2, and `current_viewing_key` returns generation 7.

**The labels are stripped before the SDK sees the keys.** Upstream has no notion of them and must not gain one, so `WalletAuthority::viewing_keys()` hands over a plain list.

## Merge's custodian path runs through `Merge::new`, not the SDK action

`create_merge` takes a concrete `&ShieldedKeypair`, whose `signing_key` only constructs from a secret (`from_bytes`, `from_ed25519`, `new`). So that entry point is unreachable from a service holding no Ed25519 secret — even though merge's own documentation says it proves ownership in-circuit from the nullifier secret and therefore needs no owner signature.

One level down it is reachable. `Merge::new` and `validate_merge_inputs` are generic over `ShieldedKeypairTrait`, and every method the merge path actually calls — signing *pubkey*, curve, address derivation, nullifier derivation — is satisfiable from public material plus the nullifier secret. `MergeKeypair` is that adapter, and `builds_a_merge_with_no_signing_key_in_the_process` drives `Merge::new` through it in a process holding no signing key.

## `sign_p256` is a documented dead end, not pending work

It was `todo!`, which reads as an invitation to implement it. It is now `unimplemented!` with the reasoning inline, mirroring how `encrypt_anonymous_transfer` already handles a genuinely out-of-scope method in the same file.

At the pinned rev the only real invocation in the tree is inside `key_binding_proof`, which returns `Ok(None)` before reaching it when the signing pubkey is Ed25519 — and every SDP custody wallet is Ed25519. Every other occurrence is a trait declaration, a blanket impl, a test double, or the CLI. The key-authority design agrees from its own side: it exposes a sign-p256 route and marks it unused by the shipping flows.

Note what the comment does *not* say: it **is** satisfiable from the viewing key this authority holds. So if a future rev starts calling it, the fix is to implement it here — not to route it to the key authority, which would move a secret for no reason.

## The residual risk, stated plainly

`ShieldedKeypairTrait::sign` returns `[u8; 64]` with no error channel, so an implementation without a signing key cannot decline — only panic. Merge never calls it at the pinned rev, verified against `Merge::new` and `validate_merge_inputs`.

**That makes the panic unreachable by inspection of upstream's call graph rather than by the type system**, which is a weaker guarantee than the rest of this module enjoys, and it has to be re-checked on every rev bump. `signing_panics_because_this_authority_holds_no_signing_key` is a `#[should_panic]` test pinning the current behaviour so it stays a loud, documented failure rather than a surprise.

- **Three trait methods remain `todo!`**: transfer-output encryption, split-bundle encryption, and — separately — `encrypt_anonymous_transfer` stays a deliberate `unimplemented!` because anonymous transfers are unresolved upstream (there is no `Anonymous` circuit variant; anonymity is a payload-encryption mode).
- **A nullifier secret is 31 bytes, not 32** — a BN254 field element below the field modulus, while viewing keys are 32. The compiler caught this against a contract that had assumed 32. Both widths now live in `src/validate.rs` as the single definition, imported here, so the asymmetry cannot drift between the boundary and the authority.
- **`allow(dead_code)` stays until the handlers construct this.** Removing it early would mean either warnings or a fake call site.

## Upstream asks this raises

Two, both about merge, both worth sending to Helius:

1. **Move `sign` off `ShieldedKeypairTrait`** into a signing-only trait, or make it return `Result`. Then a proof-only custodian satisfies the trait honestly instead of declaring a method it can only panic in, and the residual risk above stops existing.
2. **Provide a custodian submission path for merge.** `submit_merge_transaction` takes a raw fee-payer `Keypair` and submits directly, which a remote signer such as Kora cannot satisfy. Upstream already did the harder half — it accepts `MergeMaterial { signing_pubkey, viewing_pubkey, nullifier_key }` rather than a keypair, explicitly "leaving every signing/viewing/funding secret behind", which is exactly this service's shape. Only the fee payer and the submit-versus-return-bytes shape remain.

## Testing

```
cargo fmt --check
cargo test --locked          # 32 unit + 16 contract tests, all passing
```

Ten unit tests cover this module: address derivation with no signing key, identity agreement across both traits, the public-halves equivalence above, viewing-key and nullifier-key exposure, approval capture, generation-based resolution, newest-key selection, label stripping, and the `#[should_panic]` guard on `sign`. The merge test is the load-bearing one — it constructs `Merge::new` in a process that holds no Ed25519 secret, which is the claim this PR exists to prove.

### DIFF
--- a/apps/sdp-helius-gateway/README.md
+++ b/apps/sdp-helius-gateway/README.md
@@ -96,7 +96,7 @@ Each preflight read is bounded in-process. The agave client retries a `429` five
 ## Running it
 
 ```bash
-cargo test                 # 22 unit + 16 contract tests
+cargo test                 # 32 unit + 16 contract tests
 cargo clippy --all-targets --locked -- -D warnings
 cargo fmt --check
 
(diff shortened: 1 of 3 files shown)


nbaztec (dismissed):

TarekkMA (dismissed):

snowmead:

PR #1292, opened 2026-08-13
build(helius): gateway container image and compose profile

snowmead:
## What

Runtime image plus an opt-in compose service, so the gateway can be built and run locally and is ready for a Cloud Run deploy decision. **No CI cost:** the compose service is behind a profile, and the image is not built in CI.

## The Dockerfile

Staged through `cargo-chef` so the expensive layer — ~450 crates including `groth16-solana`, `light-poseidon` and the `ark-*` graph — keys only on `Cargo.lock` and survives source edits.

The build context is the crate directory, not the repo root, because this is not a pnpm workspace member and has nothing to copy from outside itself. That is also why the ignore file is `Dockerfile.dockerignore` rather than a root `.dockerignore`.

The runner stage carries no toolchain, no source and no cargo registry, runs as UID/GID 1001 to match `sdp-api` and `sdp-web` so a shared k8s `securityContext` works, and both base images are digest-pinned to match the convention in `apps/sdp-api/Dockerfile`.

`HEALTHCHECK` calls the binary's own `--health-check` (implemented in `src/main.rs`, which probes `/health` over loopback and maps the status line to an exit code) rather than installing `curl` — one fewer package to keep patched and to answer for in the Trivy scan that gates every tagged release. Liveness only, matching `sdp-api`: any status below 500 means the HTTP layer is up. Dependency probing belongs on a readiness probe, and `/health` deliberately does not reach the indexer, prover or RPC.

amd64 only. Cloud Run is amd64, and cross-building the Groth16 and Poseidon dependencies under QEMU is slow enough to be impractical in any build that has to finish on a timer.

## Two secrets in compose

The gateway verifies each route against its own caller's HMAC secret, so compose now supplies both `HELIUS_GATEWAY_HMAC_SECRET_SDP_API` and `HELIUS_GATEWAY_HMAC_SECRET_KEY_AUTH` as local-only placeholders, both at least 32 bytes.

Worth knowing when testing against it: **a client configured with only one of them can reach only that caller's half of the API.** That is the control working, not a misconfiguration.

**Upstream endpoints stay env-only by design.** `PHOTON_URL` and `PROVER_URL` are never read from a request: accepting a caller-supplied prover URL alongside a decrypted nullifier key would be an exfiltration primitive.

## Why the profile, and why the smoke job is unaffected

`docker compose --profile rings up sdp-helius-gateway`. Behind a profile for two reasons: it is a release build that would push `docker_compose_smoke` past its 25-minute timeout, and it is useless without a reachable Photon indexer and prover, which the default local stack does not provide.

Verified rather than assumed:

```
$ docker compose config --services
postgres redis sdp-migrate sdp-api sdp-docs sdp-web

$ docker compose --profile rings config --services
postgres redis sdp-migrate sdp-api sdp-docs sdp-helius-gateway sdp-web
```

- **No CI image build.** An earlier draft added a `docker_build_gateway` job to `ci.yml`. It was dropped: it had no path filter, so every PR in the repo would have paid a Rust release build (3–8 min warm, 30–45 cold), and nothing consumes the image yet. It should land with the deploy target that needs it, path-gated the way `rust-gateway.yml` is.
- **Not wired into `release-images.yml`.** That workflow hardcodes `context: .` and builds amd64+arm64 before a multi-arch push; this crate needs a crate-directory context and amd64 only. Wiring it in means per-service `context` and `platforms` in the matrix, which changes the release path for three production images. Worth asking first whether a Cloud Run service with SDP-internal IAM belongs in the self-hosted GHCR bundle at all.
- **The `--mount=type=cache` directives do nothing in a GHA build.** `cache-to: type=gha` exports layers, not cache mounts, so the registry download repeats on a cold cache. The `cargo chef cook` layer is still layer-cached, which is the expensive part.
- **The compose service is not usable end to end yet.** It needs a Photon indexer and prover alongside it; those are not in the stack. Until then it documents the env contract rather than enabling a workflow.
- **Cloud Run still needs `--timeout=700`** for the prove route's layer timeout to be reachable at all, and `--task-timeout=780` for the reconciler. Both are sdp-infra, neither is applied.

## Testing

```
docker compose config --services                    # gateway absent
docker compose --profile rings config --services    # gateway present
docker compose --profile rings config               # both secrets resolve
cargo fmt --check
cargo test --locked          # 32 unit + 16 contract tests, unchanged by this PR
```

CI still has to verify the image actually builds — there is no gateway image job, by the decision above, so the Docker build is unexercised in CI on purpose. A full `docker build` against this Dockerfile has not been run on Linux/amd64; it should be before anything depends on the image.

### DIFF
--- a/apps/sdp-helius-gateway/Dockerfile
+++ b/apps/sdp-helius-gateway/Dockerfile
@@ -1,0 +1,83 @@
+# sdp-helius-gateway — Rust runtime image. Unlike the Node images in this repo,
+# the build context is this directory, not the monorepo root: the crate is not a
+# pnpm workspace member and has nothing to copy from outside itself.
+#
+#     docker build -f apps/sdp-helius-gateway/Dockerfile -t sdp-helius-gateway \
+#       apps/sdp-helius-gateway
+#
+# amd64 only. Cloud Run is amd64, and cross-building the Groth16 and Poseidon
+# dependencies under QEMU is slow enough to be impractical in any build that has to
+# finish on a timer. Apple Silicon developers either emulate or run the gateway
+# remotely.
+#
+# Digest-pinned to match the convention in apps/sdp-api/Dockerfile.
+
+# --- chef: cargo-chef, so the dependency graph caches independently of our source
+FROM rust:1.97.0-slim-bookworm@sha256:6d220bf85c74e842a79da63997af8d2e74455c0b8847d8bb3a5888572334991d AS chef
+WORKDIR /build
+# git: the zolana crates are unpublished and resolved from a git revision.
+#
+# pkg-config/libssl-dev: `openssl-sys` is in the graph, and not for TLS — reqwest
+# here uses rustls. The path is
+# openssl-sys <- openssl <- solana-secp256r1-program <- zolana-user-registry-interface,
+# so it is the P256 curve support that needs it. That also means the runtime image
+# needs libssl3; see the runner stage.
+RUN apt-get update \
+  && apt-get install --no-install-recommends -y git pkg-config libssl-dev ca-certificates \
+  && rm -rf /var/lib/apt/lists/*
+RUN --mount=type=cache,id=cargo-registry,target=/usr/local/cargo/registry \
+  cargo install cargo-chef --version 0.1.77 --locked
+
+# --- planner: reduce the manifest to a dependency-only recipe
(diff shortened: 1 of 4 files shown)


nbaztec (dismissed):

TarekkMA (dismissed):

snowmead:

PR #1300, opened 2026-08-14
feat(earn): add Kamino as a catalogue-only vault provider

keeganthomp:
## Why

Earn V1 was built around one provider shape: **Ground**, a custodial partner fronting an omnibus portfolio wallet that SDP provisions and moves money through. Communal vaults such as Kamino use a second shape, and this PR proves the existing provider seams can absorb it without rewriting the Earn domain.

**Kamino is non-custodial**: a K-Vault is an on-chain vault that the customer's own wallet deposits into. There is no portfolio wallet for SDP to provision, fund, or withdraw from. Kamino therefore implements the base `EarnVaultProvider` catalogue contract plus optional live metrics, but deliberately does not implement the portfolio-wallet capability.

This is a complete **catalogue-only** integration. Kamino deposit/withdraw execution is a separate money movement model and remains out of scope.

### Kamino catalogue provider

- Adds a keyless Kamino provider backed by the public `api.kamino.finance` K-Vault endpoints. Provider availability now derives from the credential keys an availability definition actually reads, so public providers do not require invented secrets.
- Catalogues Kamino's mainnet shelf into both SDP environments so sandbox integrators can browse the real vaults while still receiving an explicit `fundable: false` signal.
- Filters the permissionless registry through declared stablecoin support and `KAMINO_MIN_TVL_USD` ($100k). The committed inventory command, `pnpm --filter @sdp/api earn:inventory:kamino`, records the admitted vaults and largest near-misses so the threshold stays reviewable.
- Treats each provider shelf read as all-or-nothing. Missing page payloads or a live pagination token at the safety cap throw and preserve the existing catalogue instead of turning a partial upstream read into mass delisting.
- Keeps Ground behavior and its portfolio-wallet implementation intact.

### Cluster-aware fundability

Migration `0057_earn_strategy_host_cluster.sql` adds `earn_strategies.host_cluster`, backfills existing rows from their environment, and makes the field required. This records where the instrument actually lives instead of assuming catalogue environment and host chain are identical.

`isClusterFundableInEnvironment` is the single server-side cluster comparison:

1. `assertKnownYieldSources` applies it immediately before provider mutation on create and re-target.
2. `mapToEarnStrategy` exposes `hostCluster` and a request-derived `fundable` boolean. The flag is not stored because the same mainnet-hosted row is fundable for a production caller and not fundable for a sandbox caller.

Provider capability remains a separate gate. Creating a Kamino portfolio program returns 501 through `supportsPortfolioWallets`; no route branches on a literal provider ID.

### Five-minute live metrics

- Adds the optional `EarnLiveMetricsProvider` capability and a five-minute refresh pass for APY, TVL, and volatile risk metadata.
- Refresh is update-only: it cannot admit a vault that the hourly catalogue rejected, and metadata is merged so catalogue-owned fields such as curator survive.
- Ground deliberately does not implement this capability because its rate arrives from the same paged endpoint as its catalogue.
- The managed job runs metrics before the hourly catalogue sync so a stalled provider cannot burn the sync's Redis slot, and the refresh reports through its own Sentry cron monitor.

### Full catalogue visibility in the dashboard

The deposit strategy step now renders the **full active catalogue** for comparison while keeping visibility separate from deposit eligibility.

| Strategy | Sandbox project | Production project |
| --- | --- | --- |
| Ground | Selectable when its cluster and stablecoin are supported | Selectable when its cluster and stablecoin are supported |
| Kamino | Visible, disabled, labelled **Mainnet only** | Visible, disabled, labelled **Browse only** |

`strategyDepositEligibility` owns the client-side decision and prioritizes the most actionable reason: environment mismatch, then unsupported provider capability, then unsupported asset. Ineligible rows keep their APY, pool size, backing, access, and curator visible, but their radio and row interaction are disabled.

Selection is guarded independently at the row, URL preselection/deep-link, and wizard-state boundaries. A stale or crafted strategy ID cannot advance to review. Ground remains the only provider supported by this portfolio-deposit wizard.

The older-API compatibility rule remains intentional: `fundable !== false` admits a response from an API version that predates the field, avoiding a web-only preview that blanks the catalogue before the API deploys. Once the current API is live, every row carries the explicit flag.

## Validation

- Catalogue sync produced 21 Kamino rows per environment, all with `host_cluster = mainnet-beta`; Ground rows were untouched, a second sync preserved IDs and timestamps, and the delist pass removed nothing.
- A sandbox `GET /v1/earn/strategies` returned 21 Kamino rows with `fundable: false` alongside fundable Ground rows.
- `POST /v1/earn/programs` with `provider: "kamino"` returned 501 through the capability gate.
- A Ground program allocating a Kamino reference returned 400.
- Flipping only `host_cluster` on an otherwise-valid Ground row changed it from accepted to refused and back, isolating the cluster gate.
- Corrupting a Kamino row's APY/TVL and running the metrics pass restored live values while retaining catalogue-owned metadata; Ground was skipped.
- The full `sdp-web` unit suite passed: 141 files and 1,165 tests. Web typecheck and targeted Biome checks also passed.
- Dashboard model and component tests cover visible-but-disabled Kamino rows, eligibility copy, row/radio interaction, deep-link rejection, selection clearing, and normal Ground selection.

## Not in scope

- Kamino deposit/withdraw transaction construction
- Kamino user position reads
- Non-stablecoin vaults outside `EARN_DEPOSIT_TOKEN_SYMBOLS`
- Jupiter Lend
- Kamino program/position surfaces beyond catalogue comparison

## Reviewer guide

- Start with `packages/sdp-earn/src/providers/kamino/client.ts` for distillation, pagination safety, and decimal APY truncation.
- Review `apps/sdp-api/src/routes/earn/handlers/program.ts` and `packages/sdp-earn/src/support.ts` for the mutation and cluster gates.
- Review `apps/sdp-api/src/cron/earn-metrics-refresh.ts` for cadence, update-only behavior, and monitoring.
- Review `apps/sdp-web/src/app/dashboard/markets/earn/deposit/earn-deposit-model.ts`, `earn-deposit-wizard.tsx`, and `strategy-step.tsx` for visibility-versus-selection behavior.
- Migration 0057 adds, backfills, and then constrains `host_cluster`; it is safe on a populated table and re-runnable.
- Supporting docs cover the architecture flow, provider pluggability rules, Kamino inventory, and package/API/UI ownership notes.


apps/sdp-api/scripts/inventory-kamino-catalogue.ts
```diff
@@ -0,0 +1,323 @@
...
+const DROP_REASON_LABELS: Record<KaminoCatalogueDropReason, string> = {
+  unknown_deposit_mint: "Deposit mint is not a well-known SDP token",
+  not_a_deposit_token: "Known token, but outside Earn's stablecoin set",
+  unnamed: "Vault has no on-chain name",
+  no_metrics: "No metrics row — TVL could not be established",
+  below_tvl_floor: `TVL below the $${KAMINO_MIN_TVL_USD.toLocaleString("en-US")} floor`,
+};
+
+function escapeCell(value: string): string {
+  return value.replace(/\|/g, "\\|");
```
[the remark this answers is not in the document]
keeganthomp: Fixed in ac5db729 — `escapeCell` now escapes the backslash first, then the pipe. The old order left a trailing `\\` to pair with the escape we added, so a vault named `foo\\` rendered `foo\\\\|` and the pipe became a live column separator. Vault names are third-party controlled (anyone can create and name a Kamino vault), so this is the sanitiser rather than a formality.

apps/sdp-api/scripts/inventory-kamino-catalogue.ts
```diff
@@ -0,0 +1,323 @@
...
+
+  const inventory: Inventory = {
+    fetchedAt: new Date().toISOString(),
+    minTvlUsd: KAMINO_MIN_TVL_USD,
+    totalVaults: vaults.length,
+    rows,
+  };
+
+  await mkdir(INVENTORY_ROOT, { recursive: true });
+  await writeFile(INVENTORY_FILE, `${JSON.stringify(inventory, null, 2)}\n`, "utf8");
```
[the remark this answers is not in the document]
keeganthomp: Fixed in ac5db729 — the snapshot is now validated on WRITE, not only when read back, so the zod schema sits between the upstream response and the tracked file. Every field is constrained rather than merely typed: base58 patterns on the address-shaped ones (`address`, `tokenMint`) and bounded lengths on the free text (`name`, `apy`, `curator`, …), so an upstream response cannot write unbounded or malformed content into a committed artifact. Re-fetched against the live API to confirm the constraints accept real data — 21 catalogued of 170, unchanged.

apps/sdp-api/src/db/migrations/postgres/0057_earn_strategy_host_cluster.sql
```diff
@@ -0,0 +1,32 @@
...
+-- written by the Ground sync (or the local dev seed, which mirrors it), and both
+-- only ever catalogue a source hosted on the environment's own chain.
+UPDATE earn_strategies
+   SET host_cluster = CASE environment
+                        WHEN 'production' THEN 'mainnet-beta'
+                        ELSE 'devnet'
+                      END
+ WHERE host_cluster IS NULL;
+
+ALTER TABLE earn_strategies ALTER COLUMN host_cluster SET NOT NULL;
```
keeganthomp: [P1] Preserve compatibility with the previous catalogue writer

The production workflow executes migrations before the new service or cron image takes traffic, and its rollback restores the previous cron image. That revision's `INSERT` omits `host_cluster`. Once this line makes the column `NOT NULL`, every old/in-flight or rolled-back catalogue upsert fails the not-null check (including the `ON CONFLICT` path), so a failed rollout leaves the restored job unable to refresh the shelf. Use an expand/contract rollout (nullable in this release, constraint in a later release) or keep a DB-side compatibility value/trigger derived safely from `environment` until every writer is past the old shape.

keeganthomp (commented):
I found three blocking correctness/reliability issues and one minor rate-precision defect at `f5d21ab`. I would request changes, but GitHub does not permit a blocking self-review, so this is submitted as a comment; the P1 findings below are blockers.

I inspected all 48 changed files, the production migration/deploy/rollback order, current CI, and existing review threads. Focused local validation passed: 139 `@sdp/earn` tests, 37 API cron/job tests, 89 Earn web tests, API/Earn typechecks, and `git diff --check`. The full API DB suite could not run locally because this environment has no container runtime; GitHub ran it and its sole unit failure is the unrelated custody audit-witness test. The two Kora failures are also unrelated setup/config failures, but all red checks still need clean reruns before merge.

packages/sdp-earn/src/providers/kamino/client.ts
```diff
@@ -0,0 +1,519 @@
...
+  const total = parts.reduce<number>((sum, part) => sum + (part ?? 0), 0);
+  return Number.isFinite(total) ? total : undefined;
+}
+
+/** Curator is data, not code (ADR 0002): match a house named in the vault's
+ * own name, else attribute the vault to Kamino, which curates its own. */
+export function deriveKaminoCurator(name: string): string {
+  const haystack = name.toLowerCase();
+  for (const [pattern, curator] of KAMINO_CURATOR_HOUSES) {
+    if (haystack.includes(pattern)) {
```
keeganthomp: [P1] Do not derive trusted curator identity from a permissionless name

This PR documents that anyone can create and name a K-Vault, but this substring match—and the fallback to `kamino` below—turns that attacker-controlled label into `riskMetadata.curator`. A creator who maintains the TVL floor can name a vault `Steakhouse ...` and SDP will publish it as curated by Steakhouse; any unmatched vault is automatically represented as Kamino-curated. TVL authenticates neither claim. Populate curator only from verified authority/address data or an audited vault-address allowlist; otherwise omit it. The same trust boundary should be applied before using the name to assert an RWA classification.

apps/sdp-api/src/cron/earn-metrics-refresh.ts
```diff
@@ -0,0 +1,186 @@
...
+async function refreshProviderMetrics(
+  repo: EarnRepository,
+  client: EarnLiveMetricsProvider,
+  ctx: EarnRuntimeContext
+): Promise<void> {
+  const logContext = { provider: client.provider, environment: ctx.environment };
+
+  let metrics: ProviderStrategyMetrics[];
+  try {
+    metrics = await client.listStrategyMetrics(ctx);
```
keeganthomp: [P1] Bound the public provider read below the managed-job timeout

`listStrategyMetrics` ultimately uses `fetch` without an `AbortSignal` or any timeout. If Kamino accepts a connection but stalls its headers/body, this await never reaches the catch before the managed Cloud Run job's documented 120-second timeout; the process is killed, and because this refresh runs before `runEarnCatalogueSyncIfDue`, every tick can lose the catalogue sync too. Kamino is keyless and always participates whenever Earn is enabled, so this violates the promised per-provider degradation. Add an abortable request/provider deadline comfortably below the job timeout and continue to the remaining work on expiry.

packages/sdp-earn/src/providers/kamino/client.ts
```diff
@@ -0,0 +1,519 @@
...
+export function truncateKaminoApy(value: string | null | undefined): string | undefined {
+  const trimmed = value?.trim();
+  if (!trimmed || !KAMINO_PLAIN_DECIMAL.test(trimmed)) {
+    return undefined;
+  }
+  const point = trimmed.indexOf(".");
+  if (point === -1) {
+    return trimmed;
+  }
+  const truncated = trimmed.slice(0, point + 1 + APY_DECIMAL_PLACES).replace(/\.?0+$/, "");
```
keeganthomp: [P3] Floor negative rates instead of slicing toward zero

For a real negative rate, slicing moves the quote upward: `-0.0123456789` becomes `-0.012345`, and a sub-precision loss becomes `0`. That contradicts the stated invariant that SDP never quotes above the provider. When a negative value has a non-zero discarded tail, floor it at the retained precision (away from zero), and add a test that asserts the returned numeric value is `<=` the provider value.

apps/sdp-api/src/db/migrations/postgres/0057_earn_strategy_host_cluster.sql
keeganthomp: Confirmed and fixed in 64b27cbd — expand-only.

I checked the deploy ordering rather than assume it: `deploy-sdp-api-gcp.yml` runs **Run database migrations** *before* both **Deploy service** and **Update cron job image**, so between those steps the cron job is still the previous image, whose catalogue INSERT does not list `host_cluster`. Rollback is the worse half — it restores the old image over the applied schema, so the failure is permanent rather than a window.

- `SET NOT NULL` is gone; the migration is ADD COLUMN + backfill only, with a header saying why and that the contract half belongs in a later release once no deployable writer predates the column.
- The read is the compatibility value you suggested, derived safely from `environment`: `mapStrategyRow` resolves a NULL to `CLUSTER_BY_SDP_ENVIRONMENT[row.environment]` — the same rule as the backfill, and true of every writer that predates the column. Deliberately not fail-closed: a NULL coming back un-fundable would drop live Ground strategies out of the wizard, which is the writer's problem showing up as a customer-visible outage.
- Nullability isn't a hole in the invariant, since `UpsertEarnStrategyInput.hostCluster` stays required — every writer on this release states it.
- Both halves are pinned by `earn.repository.test.ts` → "admits a row from a writer that predates host_cluster, and reads it as this environment's cluster", which INSERTs without the column in both environments: the write must be accepted AND the read must resolve it.

Also wrote the general rule into `docs/contributing/earn-pluggability-playbook.md` ("A new catalogue column is EXPAND-ONLY in the release that adds it") so the next provider that needs a column doesn't rediscover this.

packages/sdp-earn/src/providers/kamino/client.ts
keeganthomp: Agreed, and removed entirely in 64b27cbd.

Verified the premise first: Kamino's own SDK exposes `KaminoManager.createVaultIxs` (anyone can init a vault), and the live shelf carries "PC Test Vault Dawid" — someone's personal test vault — so the name really is attacker-chosen free text. The $100k floor is a cost, not an authorization, and it's recoverable: clear it for one hourly sync, get attributed, withdraw.

- `deriveKaminoCurator` and `KAMINO_CURATOR_HOUSES` deleted. Snapshots carry **no** `curator` — including the `kamino` fallback, which was the same forgery with a different name on it.
- `KAMINO_RWA_NAME` and `classifyKaminoSourceKind` deleted. Every vault is `defi`, which is the one classification a K-Vault's own mechanics establish (it allocates into Klend reserves). `declaredSupport.sourceKinds` narrows to `["defi"]` so the envelope states what the client can honestly emit.
- The nine display-only curator labels this PR added to `EARN_KNOWN_CURATOR_LABELS` are removed with it — nothing writes those ids now.
- Census script drops its Curator column; the header says plainly that some catalogued vaults really are RWA-backed and really are run by a named house, and that an assertion anyone could forge is worse than an absent one.
- `client.test.ts` → "never derives a curator or an RWA classification from the vault name" runs a real house name and every RWA term the old regex matched through distillation, asserting `sourceKind === "defi"` and no `curator` — while still asserting the name is carried through, since quoting it is fine, vouching for it isn't.

Worth flagging the cost since it's a product regression, not a free win: three genuinely RWA-backed vaults (Kamino Institutional Commodity Yield, RWA USDC, Kamino Private Credit USDC) now catalogue as `defi`, so `?sourceKind=rwa` returns nothing for Kamino. That's the right trade — an unverifiable claim an attacker can forge is worse than no claim — but the fix is real: a verified authority/address source or an audited vault-address allowlist. Both fields are documented in the client and the playbook as returning once that lands.

apps/sdp-api/src/cron/earn-metrics-refresh.ts
keeganthomp: Fixed in 64b27cbd, at both levels you named.

Confirmed the blast radius before sizing anything: `src/job.ts` runs `runEarnMetricsRefreshTick` **first and awaits it**, then `runEarnCatalogueSyncIfDue`, and `sdp_api_cron_timeout` is `120s` in both dev and prod. So you're right that this wasn't just stale rates — a hung read spent the execution and the catalogue sync never ran, every tick, for as long as the provider stayed slow.

**Request level.** `ProviderRequestInit` takes an opt-in `timeoutMs`, and `providerFetch` builds one `AbortSignal.timeout` covering the fetch **and** `response.text()`. The body has to be inside it: a provider that answers headers promptly and then stalls mid-stream hangs exactly as long as no timeout at all — that's the failure undici's 300s headers default misses. An abort surfaces as `PROVIDER_UNAVAILABLE` naming the elapsed budget, which is what every caller already handles. Opt-in rather than a global default because the right bound belongs to the caller's deadline, not to HTTP — a request inside a 120s job and one inside a customer's API call can't share a number. Kamino sets 10s (~7× the measured 1.4s pass).

**Provider level.** A per-request cap doesn't bound a read that makes many requests — a server echoing a pagination token forever is twenty prompt responses, not one slow one. So `EARN_PROVIDER_METRICS_DEADLINE_MS` (20s) bounds one provider's whole pass per environment, and expiry logs at error and **continues to the next provider**. Sized against the job: 2 environments × 1 live-metrics provider × 20s = 40s worst case, leaving the catalogue sync the larger half of the 120s. The abandoned promise's rejection is swallowed so an orphan can't surface as an unhandled rejection after the pass has moved on.

Test: "gives up on a provider that outruns the deadline and refreshes the rest" — a provider whose `listStrategyMetrics` never settles, on fake timers, asserting the hung one is abandoned in both environments while Kamino still writes. Plus two `fetch.test.ts` cases (stalled headers, stalled body) and one pinning that a caller setting no `timeoutMs` still gets no signal, so Ground's behaviour is unchanged.

packages/sdp-earn/src/providers/kamino/client.ts
keeganthomp: Right, and fixed in 64b27cbd. The invariant is directional, and "truncate" only implements it on one side of zero.

`truncateKaminoApy` now floors a negative value AWAY from zero whenever the discarded tail is non-zero, by long addition on the digit string — the no-float rule still holds, since `0.012345 + 0.000001` isn't `0.012346` in a double. Carry is handled (`-0.9999999` → `-1`), and a negative value whose tail is all zeros is already exact and is left alone.

Both cases you called out:
- `-0.0123456789` → `-0.012346` (was `-0.012345`)
- `-0.0000000001` → `-0.000001`, the smallest loss six places can express (was `0` — which was the worse of the two, since it showed a losing vault as flat)

The test you asked for is `"never returns a value above the provider's, on either side of zero"`: nine reported rates spanning both signs, sub-precision, and carry, each asserting `Number(quoted) <= Number(reported)`. Floats are fine in the assertion — it's about ordering, and the value under test is still produced by string surgery. Example-level cases sit alongside it.

multipletwigs (approved):
yeet

PR #1308, opened 2026-08-14
feat(web): list custody connections on the wallets surface

tobySolutions:
Adds a paginated connections inventory at /dashboard/wallets/connections backed by GET /internal/dashboard/custody/connections. Each row shows the credential label, provider, lifecycle status with a short failure hint, the wallets created through the connection (joined from /v1/wallets via custodyConnectionId), and the created date. The wallets overview links to it when the privy-byok flag is on, and the empty state routes into the existing setup flow.

Management actions from the ticket (make default, add wallet through a connection) are not in this slice. The list endpoint also does not return isDefault yet, so the default marker waits on that.

#### Screenshots

Screenshot 2026-08-18 at 11 42 21


apps/sdp-web/src/app/dashboard/custody/connections/connections-list.tsx
```diff
@@ -0,0 +1,267 @@
...
+                ) : null}
+              </TableCell>
+              <TableCell className={`${CREATED_COLUMN_CLASS} text-xs text-secondary`}>
+                {formatDate(connection.createdAt, locale, t)}
+              </TableCell>
+            </TableRow>
+          ))}
+        </TableBody>
+      </Table>
+      {pageCount > 1 ? (
```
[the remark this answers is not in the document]
tobySolutions: confirmed, fixed in 9c1d1888: resolveConnectionsPage clamps a past-the-end page to the last real page and refetches, regression tests pin the clamp and the genuinely-empty case

apps/sdp-web/src/app/dashboard/custody/connections/connections.data.ts
```diff
@@ -0,0 +1,134 @@
...
+  filters: ConnectionsFilters
+): Promise<{ result: ConnectionsPageResult; filters: ConnectionsFilters }> {
+  const result = await fetchConnectionsPage(request, filters);
+  const { total } = result.pagination;
+  const pageCount = Math.max(1, Math.ceil(total / CONNECTIONS_PAGE_SIZE));
+  if (result.connections.length > 0 || total === 0 || filters.page <= pageCount) {
+    return { result, filters };
+  }
+  const clamped = { page: pageCount };
+  return { result: await fetchConnectionsPage(request, clamped), filters: clamped };
```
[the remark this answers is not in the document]
tobySolutions: confirmed, fixed in the follow-up commit: a clamped refetch that is still past the end now falls back to page 1, which is always in range; test pins the shrinking-total race with three sequenced responses

apps/sdp-web/src/app/dashboard/custody/connections/connections.data.ts
```diff
@@ -0,0 +1,147 @@
...
+  const clamped = {
+    page: Math.max(1, Math.ceil(result.pagination.total / CONNECTIONS_PAGE_SIZE)),
+  };
+  const clampedResult = await fetchConnectionsPage(request, clamped);
+  if (isSettledPage(clampedResult, clamped.page)) {
+    return { result: clampedResult, filters: clamped };
+  }
+
+  const firstPage = { page: 1 };
+  return { result: await fetchConnectionsPage(request, firstPage), filters: firstPage };
```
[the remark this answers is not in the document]
tobySolutions: confirmed, terminated the whole class instead of chasing the race another level: the empty state now keys on the empty slice rather than the total, so no stale count can ever render a rowless table on any page; test pins an empty slice against a nonzero total

tobySolutions:
on the 4/5 rationale: when page 1 returns an empty slice alongside a stale nonzero total in the same response, the slice is the direct observation and the total is the derived aggregate, so the empty state is the correct terminal render; revalidating against the total would loop forever under concurrent deletes. all review threads are resolved.

apps/sdp-web/src/app/dashboard/custody/connections/connections-list.tsx
```diff
@@ -0,0 +1,270 @@
...
+    </Button>
+  ) : null;
+
+  // Keyed on the slice, not the total: a deletion race can hand page 1 an
+  // empty slice with a stale nonzero count, and a rowless table is never the
+  // right render for that.
+  if (connections.length === 0) {
+    return (
+      <ListEmptyState
+        icon={<CableIcon className="size-5" />}
```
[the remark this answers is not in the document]
tobySolutions: same race discussed and settled earlier on this PR: when one response carries an empty page-1 slice and a nonzero total, the slice is the server's direct observation and the total is the derived aggregate, so the empty state is the correct terminal render. Revalidating against the total loops forever under concurrent deletes, and rendering a rowless table would misstate the inventory worse. The add-connection action is offered in the populated state too, so it is not wrong here.

pashpashkin:
@greptileai review pls

apps/sdp-web/src/app/dashboard/custody/connections/connections-list.tsx
```diff
@@ -0,0 +1,270 @@
...
+        message={t("DashboardCustody.connectionsEmpty")}
+        description={t("DashboardCustody.connectionsEmptyDescription")}
+        action={addConnectionAction}
+      />
+    );
+  }
```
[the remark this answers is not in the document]
tobySolutions: confirmed, this corner was real and distinct from the earlier rounds. Terminal fix: an empty slice with a nonzero total is never settled on any page; it clamps, then funnels to page 1, whose slice is the inventory's own answer. Test pins the in-range emptied page landing on page 1 with rows.

apps/sdp-web/src/app/dashboard/custody/connections/connections.data.ts
```diff
@@ -0,0 +1,157 @@
...
+export async function fetchWalletsByConnection(
+  request: SdpApiClient["request"]
+): Promise<Map<string, CustodyWalletSummary[]>> {
+  const res = await request("/v1/wallets?includeAllProviders=true");
+  if (!res.ok) {
+    throw new ConnectionsRequestError(res.status);
+  }
+  const json = (await res.json()) as { data?: { wallets?: CustodyWalletSummary[] } };
+  const byConnection = new Map<string, CustodyWalletSummary[]>();
+  for (const wallet of json.data?.wallets ?? []) {
```
multipletwigs: `json.data?.wallets ?? []` (and the `{ data?: { wallets?: ... } }` type on L145) treats `data.wallets` as optional, but `/v1/wallets` always returns it. A malformed response would silently render "No wallets yet" on every row instead of failing. Repo rule is no fallbacks — type it as `{ data: { wallets: CustodyWalletSummary[] } }`, drop both `?`s and the `??`, and let a broken response throw (the page already catches wallet-fetch failures and degrades the column).

apps/sdp-web/src/app/dashboard/custody/connections/connections.data.ts
```diff
@@ -0,0 +1,157 @@
...
+
+/** Failure codes the installation service publishes; anything unknown renders the generic copy. */
+export type ConnectionCheckFailureCode =
+  | "invalid_credentials"
+  | "provider_response_unknown"
+  | "provider_account_already_connected"
+  | "wallet_conflict";
+
+export interface ConnectionLastCheck {
+  status: "running" | "success" | "failed" | "retry_unknown" | (string & {});
```
multipletwigs: These `(string & {})` open unions have no precedent in the repo and defeat exhaustiveness — the `failureHint` switch in `connections-list.tsx` can never be checked as exhaustive. The status set is already closed elsewhere: `byok-actions.ts` unions `"running" | "success" | "failed" | "retry_unknown"`, and the failure codes come from `provider-credential-installation.service.ts`. Please hoist closed unions to `packages/sdp-types/src/custody.ts` next to `CustodyConnectionLifecycle` (the API mapper, `byok-actions.ts`, and this file all re-type them by hand today), then give `failureHint` an exhaustive `never` default. Same for the `providerCredential.status` union on L30.

apps/sdp-web/src/app/dashboard/custody/connections/page.tsx
```diff
@@ -0,0 +1,77 @@
...
+export const dynamic = "force-dynamic";
+
+export default async function CustodyConnectionsPage({
+  searchParams,
+}: {
+  searchParams: Promise<Record<string, string | string[] | undefined>>;
+}) {
+  const { userId, orgId, orgRole } = await auth();
+  if (!userId) redirect(await getAuthEntryPath());
+  if (!orgId) redirect("/dashboard");
```
multipletwigs, line 32: The `privyByok` flag gates only the link on the wallets overview — this page itself renders for anyone who hits the URL directly with the flag off. Gate it the same way `setup/page.tsx` does: `if (!(await privyByok())) redirect("/dashboard/wallets")`.

apps/sdp-web/src/app/dashboard/custody/connections/connections-list.tsx
```diff
@@ -0,0 +1,270 @@
...
+        <p className="text-sm text-secondary">{t("DashboardCustody.connectionsDescription")}</p>
+        {addConnectionAction}
+      </div>
+      <Table className="[&_table]:w-full [&_table]:min-w-0 [&_table]:table-fixed">
+        <TableHeader>
+          <TableRow>
+            <TableHead className="w-[30%] @2xl/connections-table:w-[26%]">
+              {t("DashboardCustody.connectionColumn")}
+            </TableHead>
+            <TableHead className={`${PROVIDER_COLUMN_CLASS} w-[16%]`}>
```
multipletwigs: Classname concatenation via template literal — repo convention is `cn()`: `cn(PROVIDER_COLUMN_CLASS, "w-[16%]")`. Same on L214 and L253.

multipletwigs (commented):
Nice surface — pagination clamp logic and the degrade-only-the-wallet-column split are solid. Four asks before merge: one fallback, one typing hoist, one flag gate, one classname nit.

multipletwigs:
@tobySolutions apologies, i had a few custom skills to evaluate sdp prs, it surfaced a few findings, feel free to dismiss if irrelevant. thank you!

apps/sdp-web/src/app/dashboard/custody/connections/connections.data.ts
tobySolutions: good catch, typed it strictly and dropped both optionals and the fallback. A broken response now throws and the page's existing catch degrades the wallet column, instead of every row quietly reading as having no wallets.

apps/sdp-web/src/app/dashboard/custody/connections/connections.data.ts
tobySolutions: done. Hoisted three closed unions into sdp-types next to CustodyConnectionLifecycle: check status, failure code, and provider credential status. failureHint now has a never default so a new code breaks the build.

apps/sdp-web/src/app/dashboard/custody/connections/page.tsx
tobySolutions: fixed, the route redirects to /dashboard/wallets when the flag is off, same as setup does.

apps/sdp-web/src/app/dashboard/custody/connections/connections-list.tsx
tobySolutions: switched all three to cn().

tobySolutions:
@multipletwigs not irrelevant at all, all four were right and are pushed. Strict typing on the wallets response with no fallback, the three unions hoisted into sdp-types with an exhaustive never on failureHint, the route redirecting when the flag is off, and cn() for the column classes. Appreciate the pass.

multipletwigs:
@tobySolutions could this be one page instead? why are there two steps?

pashpashkin:
@tobySolutions, could you please check if the screenshots are correct?

tobySolutions:
@pashpashkin good catch, they were wrong. Both images were of the Create wallet wizard on Step 2, Provider details, not the connections list this PR adds. I have taken them off the description and will put correct ones up.

@multipletwigs that also answers yours, and I am sorry for the runaround: the two steps you were looking at belong to the Create wallet wizard in those screenshots, not to anything in this PR. This one adds a single page.

The question is still worth asking about that wizard though. It is provider choice, then provider details, and for Privy the first step is a list of one, so most people click through a screen that never asked them anything. Collapsing it is not part of this slice, but if you want it raised I will write it up for Opeyemi to file.

tobySolutions:
Correcting myself on the last paragraph: I said the first step is a list of one for Privy. It is not. `ProviderStep` renders the whole custody catalogue grouped by category, selectable or not, and `wallet-setup-flow.tsx:88-92` already starts on Provider details when you arrive with `?provider=`, which is the path the Integrations page uses.

So the wizard is two steps when you enter it cold from Wallets and one when you enter it from a provider. Whether the cold entry should still be two is a fair question, but it is not the empty screen I described.

tobySolutions:
@pashpashkin replaced. This one is the connections list the PR actually adds, with the three lifecycle states. The previous two were the Create wallet wizard, which this PR does not touch.

pashpashkin (dismissed):

tobySolutions:
Synced with main to clear a conflict, which dismissed the approval. Re-requested.

The conflict was in `wallets-workspace.tsx`: #1374 moved the tab shell from `overview`/`playground` props to a `panels` array. Kept main's structure and re-added `showConnectionsLink`. Nothing else changed.

multipletwigs:
@tobySolutions the margin is sticking

multipletwigs:
> @tobySolutions the margin is sticking

can we adjust it a bit

tobySolutions:
@multipletwigs on the two steps: that was the screenshots, not the code. Both images were of the Create wallet wizard, same thing Pavel caught. This PR is one page. The two-step flow lives in `custody/setup/wallet-setup-flow.tsx`, which this PR does not touch.

On the margin, can you point at it? The full-height card and `max-w-none` both match the other list pages, and nothing in this route is sticky, so I would just be guessing. A screenshot or your viewport width is enough.

multipletwigs (approved):

multipletwigs:
@tobySolutions i meant this image

multipletwigs:
there's a small rounded gap with overlapping borders

PR #1310, opened 2026-08-14
fix(web): give wallet setup step indicator breathing room below the header

tobySolutions:
The step indicator on the wallet setup flow sat almost flush under the header bar with only pt-2, while the issuance draft wizard and the shared wizard frame both use pt-8 for the same stepper band. This bumps the setup flow to pt-8 and mirrors the change in the wallet setup skeleton so the loading state matches. The change is on the container above the step split, so both the standard details form and the credential form variant get the same spacing.


resourcefulmind (approved):
Approving, spacing only.

PR #1311, opened 2026-08-14
fix(web): include SOL in the payments available balance card

tobySolutions:
The available balance card on the payments overview filtered native SOL out of the aggregate balance rows, so wallets holding SOL showed neither a SOL row nor its value in the total. The home page total already counts SOL because it sums the raw aggregate, so the two surfaces disagreed. Removed the SOL filter in normalizeAggregateBalances so SOL renders and totals like any other priced balance, and added unit tests pinning SOL inclusion.


resourcefulmind (approved):
Approving. Good catch that the home total and the payments card disagreed on SOL.

PR #1314, opened 2026-08-14
fix(web): resolve token labels in approvals amount/asset column

tobySolutions:
The approvals inbox showed the raw mint address in the Amount / asset column, wrapped across several lines for SDP-minted tokens. The column now resolves mints through the same shared token helpers and TokenMark treatment the transactions tables use, including org-issued token metadata, and long values truncate instead of wrapping. Unit tests pin the resolved label and the absence of raw mint rendering.

Screenshot 2026-08-14 at 19 22 20


tobySolutions:
also fixed two pre-existing wraps caught while screenshotting this table: the status badge broke mid-word under narrow widths and truncated destination addresses wrapped their tails; both cells are nowrap now

resourcefulmind (approved):
Approving. Nice that each resolution path has its own test.

PR #1317, opened 2026-08-14
fix(web): keep a single divider under the private channels tab strip

tobySolutions:
On private channels sub-pages like /instance, the back action made the top bar draw its own bottom border right above the tab strip, which already draws one, so the header showed a doubled rule. The back link now rides as plain leading content, which is how the header-tab pages keep the top bar borderless, leaving the tab strip's underline as the single divider. No behavior change to the back link itself.


resourcefulmind (approved):
Approving, single-file border fix.

PR #1318, opened 2026-08-14
fix(web): distinguish issuance from payments rows in the recent transactions card

tobySolutions:
Deploy, Receive, and Send rows in the home Recent transactions card read as one flat list, so issuance activity is indistinguishable from payments activity at a glance. Each row's type now leads with a small muted glyph for the surface it came from, the same icons those modules already use, with an sr-only category label for screen readers. No layout or column changes, and the new copy is catalog-backed in the English messages only.

builds on #1313


resourcefulmind (commented):
The content is fine. 
I'm just flagging that this is based on your pro-1677 branch rather than main, so #1313 needs to land first or this wants a rebase before it can go.

PR #1322, opened 2026-08-14
fix(api): let approved signer checks replay after their key retires

tobySolutions:
The dashboard signer check mints a short-lived API key and revokes it right after the call, but approved-operation replay required the original key to still be active and unexpired, so a signer check routed to approval could never execute once approved. Replay auth now accepts a retired key for the custody_signer_check operation type only; the approval itself is the authorization and other operation types still require an active key. The replayed extraction also reuses the memo recorded on the stored operation, since regenerating one failed the stored-operation match. Adds tests pinning both the successful replay after key retirement and the unchanged refusal for other operation types.


tobySolutions:
this now builds on #1315, which carries the stored-memo replay half, so this branch keeps only the key-retirement change. Merge #1315 first.

resourcefulmind (dismissed):
Approving. The allowRetiredKey bypass is scoped to custody_signer_check only and the negative test proves payment_transfer_execute still fails after key retirement, which is the bit I wanted to see.
Note this also fixes the replay break that #1315 introduces, so it needs to land first or alongside it.

multipletwigs:
@tobySolutions we should remove this short lived api key mint entirely, we should not have special checks for a retired key.

multipletwigs:
closing in favor of https://github.com/solana-foundation/solana-developer-platform/pull/1358

PR #1326, opened 2026-08-14
feat: workflows ui improvements and proper webhook management

arseniy-nikitochkin:
# Managed outbound-webhook endpoints

Phase 5 shipped `send_webhook` as an MVP: an inline URL plus an optional user-pasted secret on the rule's `action.params`, HMAC-signed, retried by the engine — no tables, no UI, and no delivery history at all (each retry overwrote the previous attempt's state on the single `workflow_executions` row, and response bodies were discarded). The old signature carried no timestamp, delivery id or key id, so rotating a secret would have broken every receiver instantly.

This PR adds the managed registry that deferral was waiting on: endpoints an issuer can create and rotate themselves, a per-attempt delivery log with replay, and a workflow-builder picker that points a rule at an endpoint instead of a pasted URL.

## API

`/v1/webhook-endpoints`, 8 routes behind the existing-but-unused `webhooks:read` / `webhooks:write` permissions, unifiedAuth + project scoping, OpenAPI-documented.

| Route | Notes |
|---|---|
| `GET /` | paginated list |
| `POST /` | returns the plaintext secret **once** |
| `GET /:id` | 404 once soft-deleted |
| `PATCH /:id` | label / description / status — **url is immutable** |
| `DELETE /:id` | soft delete; reports `referencingWorkflows` rather than blocking |
| `POST /:id/rotate-secret` | `gracePeriodHours` 0–168, default 24 |
| `GET /:id/deliveries` | paginated attempt log |
| `POST /:id/deliveries/:deliveryId/redeliver` | synchronous re-send, byte-identical body |

**Signing.** Secrets are server-generated (`whsec_` + 32 random bytes) and never readable afterwards, which keeps the repo's write-only secret invariant. Rotation runs a dual-key grace window: both keys sign during it, `x-sdp-signature: t=…,v1=…[,v1=…]` over `${t}.${body}` (Stripe composition), so receivers can cut over without downtime. Managed endpoints never send unsigned — an unreadable secret is a transient failure, unlike the legacy path which could still deliver.

**Migration `0051`.** `webhook_endpoints` stores secret handles as JSONB `StoredCredentialSecret` (current + previous + grace expiry), following the `asset_workflows.definition.actionSecret` precedent rather than cloning `provider_credentials`' typed columns. Endpoint ids are stable across rotation because workflow rules reference them. `webhook_deliveries` logs every attempt with the byte-exact signed request body.

## Secret lifecycle

The bulk of the review-driven work on this branch. The invariant: a secret version is either referenced by a committed row or queued for destruction, never neither.

- `queuePendingEndpointSecret` commits the destroy obligation **before** the write that would reference the version; the write discharges it in its own transaction. A post-commit discharge is not equivalent — a failed discharge would leave a queued retirement pointing at a now-live signing key, and the sweeper would destroy it.
- `createEndpoint` / `rotateSecret` / `softDeleteEndpoint` are transactional and read `FOR UPDATE`; the latter two return the handles they displaced, so the caller destroys what the lock saw rather than a stale read.
- Three defects this closed: a create whose insert failed leaked the version (no failure path existed); deleting an endpoint retired neither key; and rotation destroyed the displaced grace key *before* writing the row, which — since an unreadable live previous key fails closed — turned a rejected rotation into a full delivery outage until grace expiry, not just a leak.
- Schemas stay `.strict()` because the backend authorizes secrets by reference, not by tenant. Pinned by a test: a caller can never choose an endpoint's secret ref.

## Dashboard

`/dashboard/issuance/webhooks` — list, endpoint detail, delivery log with expandable request/response bodies and one-click redeliver, one-time secret reveal.

It sits under Issuance rather than beside it, because it exists to serve issuance workflow rules. It is **deliberately not in the sidebar**: the way in is an asset's Workflows tab, which passes the asset it came from so the page can offer a way back to that tab. The `?from=` value is validated against the id shape before it reaches an href.

The builder gained an endpoint picker (registered endpoint | custom URL, mutually exclusive, mode switch clears the other side) and the workflow surfaces went through a UX pass:

- The tier is a marker on the action's icon tile, not an `Automated`/`Sensitive`/`Requires approval` pill. The pill was the widest thing in a rule row, and in the action picker it duplicated the group heading verbatim.
- Markers use a new `--*-mark` token — each status hue at full opacity. `--*-tx` is darkened for text contrast on light surfaces (amber lands on brown, `#854d0e`) and `--*-brd` is that hue at 20%, too faint for a 10px mark.
- Marker tooltips are built on Base UI directly: the shared wrapper only styles text inside a fixed bubble, and a tinted bubble needs an opaque layer under the tint — every light-mode `*-bg` token is an 8% wash, which let row borders show through.
- Action options no longer ellipsise. At the popup's anchor width the text column is ~170px, so "Remove from allowlist" plus "Requires approval" was never going to fit one line.
- Audit rows name the action that actually ran instead of a wall of "Workflow action executed" — the engine already writes `metadata.actionType`.
- The parameters block explains itself per action; `pause` / `unpause` / `record` take none, so they state what they do instead of leaving a void under the review cards.

image
image
image
image
image
image
image

image

image

image


multipletwigs (dismissed):

GuiBibeau:
@arseniy-nikitochkin let's close this since it's not being actively worked on and product direction will likely change

PR #1330, opened 2026-08-14
feat(helius): add rings schema migration

nbaztec:
Creates the eight tables the Rings module persists to: wallets, key refs,
zones, operations, timelocks, the event feed, the asset allowlist and the
runtime health board. Numbered 0057 rather than the 0053 the design sketch
named, since 0053 through 0056 have since landed.

Three choices carry the weight here:

intent_key gets a UNIQUE index. It is the idempotency contract, and it is
what lets reserveIntent() be an INSERT ... ON CONFLICT DO NOTHING that
returns the row already there. Without it a retried request quietly creates
a second operation on the money path.

retry_of_operation_id is ON DELETE SET NULL, never CASCADE. A retry chain is
the audit evidence explaining why an operation was retried; cascading would
let one deletion erase the lineage that explains it.

Operations reach their wallet through the composite (wallet_id,
organization_id, project_id) key, following the pattern 0048 established for
earn movements. A plain wallet_id FK would let an operation in one project
point at a wallet in another, which for a privacy module is a tenant
isolation break rather than a data quality one.

Closed value sets are named CHECK constraints mirroring the as-const arrays
in packages/sdp-helius-rings/src/constants.ts, matching the repo's
no-enums-anywhere convention. Two coherence constraints go further than the
sketch asked: transfer_mode is pinned to op_type, because an anonymous
transfer rendered as registered tells an operator their counterparty was
disclosed when it was not; and failure_code/message/retryable move together
with the failed state, so a failed row is always actionable in the recovery
UI and a live row never carries stale failure text.

The allowlist seeds wrapped SOL and devnet USDC, with mints and decimals
taken from packages/sdp-types/src/well-known-tokens.ts so it cannot drift
from the registry the rest of the platform resolves against. ON CONFLICT DO
NOTHING keeps the re-run a no-op and stops the migration reviving a mint an
operator has since disabled.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

## Verification

- `vitest run src/db/migrations/0057_helius_rings.migration.test.ts` — 17 passing. The testcontainer DB is already fully migrated by the global setup, so these assertions run against the real schema; the tables existing at all is itself proof the migration applies cleanly.
- `vitest run src/test/run-postgres-migrations.node.test.ts src/db/migrations/` — 43 passing, including the runner's own cross-migration ordering invariants.
- `pnpm check:module-boundaries` — clean.
- `pnpm --filter @sdp/api exec tsc --noEmit` — no errors in the added files. Note there are 28 pre-existing errors on `main`'s working tree in untracked earn/payments WIP files, unrelated to this change.

Covered by the test: all 8 tables and their indexes exist, re-running the migration is a no-op and does not duplicate the allowlist seed, the `intent_key` UNIQUE constraint fires on a duplicate insert and leaves the reserved row intact, a retry survives deletion of the operation it retried, a cross-project wallet reference is rejected, and each coherence constraint (`transfer_mode`/`op_type`, failure detail/state, timelock release ordering) rejects the shape it exists to prevent.

## Notes for review

The allowlist seeds wrapped SOL and devnet USDC. The design doc gave both mints truncated and flagged "verify current devnet mint before seeding", so the values here come from `packages/sdp-types/src/well-known-tokens.ts` instead — worth a second pair of eyes confirming that's the registry you want this keyed to.

### DIFF
--- a/apps/sdp-api/src/db/migrations/0057_helius_rings.migration.test.ts
+++ b/apps/sdp-api/src/db/migrations/0057_helius_rings.migration.test.ts
@@ -1,0 +1,537 @@
+import { readFileSync } from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { Client } from "pg";
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it } from "vitest";
+import { env } from "@/test/helpers/env";
+
+// The test database is already fully migrated by src/test/node-global-setup.ts,
+// so 0057's tables exist before this file runs. That makes the assertions here
+// behavioural rather than structural-only: every test opens a transaction,
+// inserts real rows against the real schema, and rolls back.
+
+const migrationPath = path.join(
+  path.dirname(fileURLToPath(import.meta.url)),
+  "postgres/0057_helius_rings.sql"
+);
+const migrationSql = readFileSync(migrationPath, "utf8");
+
+const TABLES = [
+  "helius_rings_wallets",
+  "helius_rings_key_refs",
+  "helius_rings_zones",
+  "helius_rings_operations",
+  "helius_rings_timelocks",
+  "helius_rings_events",
+  "helius_rings_asset_allowlist",
+  "helius_rings_runtime_health",
+];
+
+let client: Client;
+
+/** Postgres SQLSTATEs the constraint assertions below distinguish between. */
+const UNIQUE_VIOLATION = "23505";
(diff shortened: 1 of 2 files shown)


nbaztec:
erroneous push

PR #1331, opened 2026-08-14
feat(helius): redact rings key material in logs

nbaztec:
Adds the two halves of the redaction posture the plan calls a Week-1
non-negotiable: a log redaction registry, and a build-time check that secret
material does not reach a log or a serializer.

The registry lives in apps/sdp-api/src/runtime/log-redaction.ts and is wired
into baseLoggerOptions() as pino's redact option, covering viewingKey,
nullifierKey, ringsMetadata, proof.ref, proof.internal and
keyRefs[*].material. Each entry is expanded to the root and one nesting
level, which is pino's actual limit: fast-redact's wildcard matches exactly
one intermediate key, not arbitrary depth. That limit is tolerable because
wrapped secrets redact themselves through SecretRef.toJSON() at any depth, so
the registry only has to catch plaintext that escaped wrapping. Both halves
of that claim are asserted.

The check departs from the plan's letter. It specified an ESLint rule
forbidding JSON.stringify on values typed SecretRef<*>, but this repo lints
with Biome and has no custom-rule mechanism in use, and more importantly that
rule would guard a pattern that cannot leak: SecretRef.toJSON() already
returns "[REDACTED]", so stringifying the wrapper is safe by construction.
The pattern that leaks is reveal(), whose result is an ordinary string with no
redaction behaviour. So scripts/check-secretref-serialization.mjs parses with
the TypeScript compiler's own parser and fails the build when a reveal()
result flows into JSON.stringify, a logger, String(), or a template literal,
while leaving reveal() alone where it belongs — passed to an adapter or a
signer. Stringifying a SecretRef is still reported, as pointless rather than
dangerous.

Test files are exempt, which is what the "test" RevealScope is for. The check
is a standalone script with a sibling .test.mjs, following the convention
check-module-boundaries.mjs and check-env-contract.mjs established, and runs
in the Unit Tests CI job rather than Module Boundaries because that job is the
one that installs dependencies.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

## Verification

- `node --test scripts/*.test.mjs` — 84 passing, 10 of them new (each rule's positive case, plus the negative cases that must stay legal: passing the wrapper to a logger or serializer, and calling `reveal()` to hand material to an adapter or signer).
- `vitest run src/runtime/log-redaction.test.ts src/runtime/logger.test.ts` — 15 passing.
- `pnpm --filter @sdp/helius-rings test` — 15 passing; `typecheck` clean.
- `pnpm check:module-boundaries`, `biome check` — clean.
- Confirmed the check is not passing vacuously: planting `JSON.stringify({ k: secret.reveal("adapter") })` in a source file made it report `file:line` and exit 1; removing the file returned it to green.
- Repo-wide runtime is ~1.7s, so the CI cost is negligible.

## Notes for review

**This deliberately does not implement what the plan specified.** `HELIUS_TRACKS.md` A23 asks for an ESLint rule forbidding `JSON.stringify` on `SecretRef<*>`. Two problems: the repo lints with Biome and has no custom-rule mechanism in use, and that rule would guard a pattern that cannot leak — `SecretRef.toJSON()` already returns `"[REDACTED]"`. The leak vector is `reveal()`. If you'd rather have the literal rule as well, say so and I'll add it, but it would flag safe code.

One known and intentional false positive: the check matches `reveal()` by name, so an unrelated `.reveal()` method elsewhere would be flagged. There's a test pinning that behaviour so it reads as a decision rather than a surprise.

Also worth knowing, verified empirically against pino 10: `fast-redact`'s `*.` wildcard matches exactly one intermediate key, not arbitrary depth. Registry entries are expanded to the root and one level; deeper plaintext is not covered. That's tolerable only because wrapped secrets redact themselves at any depth — which is the argument for keeping `SecretRef` mandatory rather than treating the registry as the safety net.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -250,6 +250,12 @@ jobs:
       - name: Scripts tests
         run: pnpm test:scripts
 
+      # Runs here rather than in the Module Boundaries job because it parses
+      # TypeScript with the compiler's own parser, and that job installs no
+      # dependencies.
+      - name: Check SecretRef serialization
+        run: pnpm check:secretref-serialization
+
   build_api:
     name: Build API
     runs-on: ubuntu-latest
--- a/apps/sdp-api/src/runtime/log-redaction.test.ts
+++ b/apps/sdp-api/src/runtime/log-redaction.test.ts
@@ -1,0 +1,132 @@
+import pino from "pino";
+import { describe, expect, it } from "vitest";
+import {
+  LOG_REDACTION_PATHS,
+  REDACTED_LEAF_FIELDS,
+  REDACTED_NESTED_PATHS,
+  REDACTION_CENSOR,
+} from "./log-redaction";
+import { baseLoggerOptions } from "./logger";
+
+/**
+ * Stands in for `SecretRef` from `@sdp/helius-rings`, which sdp-api does not yet
+ * depend on — A5 adds that dependency when the repositories need the domain
+ * types. What matters here is the contract pino relies on, `toJSON()` returning
+ * the censor; that `SecretRef` honours it is asserted in
+ * packages/sdp-helius-rings/src/secrets.test.ts.
+ */
+class WrappedSecret {
+  readonly #value: string;
+  constructor(value: string) {
+    this.#value = value;
+  }
+  reveal(): string {
+    return this.#value;
+  }
+  toJSON(): string {
+    return REDACTION_CENSOR;
+  }
+}
+
+/** Captures what the real logger options would actually emit to a sink. */
+function captureLogs(write: (logger: pino.Logger) => void): string {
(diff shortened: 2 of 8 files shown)


resourcefulmind (commented):
Quick question. `hasRevealedArgument` seems to only check the arguments of the call itself. What happens if someone does const k = secret.reveal("adapter") and then logs k on a separate line? And if that value gets interpolated into an Error message, would pino’s redaction still catch it, given it matches on field paths rather than string contents?”

nbaztec:
erroneous push

PR #1336, opened 2026-08-14
feat(api): support Custody Connections in API keys and policies(HOO-1022)

pashpashkin:
## What Changed

- Added Connection-aware wallet provisioning to both API-key creation endpoints:
  - `provisionWallet: true` provisions against the effective custody target.
  - `provisionWallet: { connectionId }` provisions against an exact active Connection.
- Extended selected API-key wallet authorization and wallet read/control-plane access to Config- and Connection-owned wallets.
- Added unique wallet resolution within the authorized Organization/Project scope; missing or ambiguous selectors now fail closed.
- Made policy inventory, wallet-policy management, and API-key policy bindings work with Connection-owned wallets independently of runtime availability.
- Hardened cache hydration, approval replay authorization, wallet alias handling, and transactional API-key binding updates.
- Unresolved wallet bindings now log `api_key_wallet_binding_unresolved` instead of
  dropping silently; an empty wallet allowlist in `listTransfers` now matches nothing.

## Why

HOO-1021 added Connection-owned wallet support, but API-key and policy control-plane flows still contained Config-only assumptions. This PR closes those gaps while preventing ambiguous selectors or implicit fallback from using the wrong wallet.

## Impact

- No database or data migration is required.
- Existing API response shapes remain unchanged.
- Existing `provisionWallet: true` requests remain supported.
- Exact Connection provisioning creates and binds an additional wallet without changing the Connection default wallet or the Project custody target.
- Ambiguous or unresolved selected-wallet bindings are denied rather than widened or redirected.

## Pre/Post Release Actions

- Pre-release: keep `PRIVY_BYOK_ENABLED=OFF`.
- Post-release: no action required. Keep the flag disabled until HOO-1023 is complete.

## Rollout & rollback

- Not gated by `PRIVY_BYOK_ENABLED`. Rollback = revert of this PR.
- Pre-deploy check executed on dev and prod: **0 rows** — no existing API key changes
  behavior at deploy (query below, kept for re-runs).
- After deploy: alert on the `api_key_wallet_binding_unresolved` warn log — it fires
  if a binding ever degrades to deny-only (e.g. a duplicate wallet id appearing later).

<details>
<summary>Pre-deploy check (read-only): keys whose wallet binding will not resolve</summary>

```sql
WITH selectors AS (
  SELECT k.id AS api_key_id, k.organization_id, k.project_id, p.wallet_id,
         'permission_row' AS source
    FROM api_key_wallet_permissions p
    JOIN api_keys k ON k.id = p.api_key_id
   WHERE k.status = 'active'
  UNION ALL
  SELECT k.id, k.organization_id, k.project_id, k.signing_wallet_id,
         'legacy_signing_wallet'
    FROM api_keys k
   WHERE k.status = 'active'
     AND k.signing_wallet_id IS NOT NULL
     AND k.signing_wallet_id <> ''
     AND NOT EXISTS (SELECT 1 FROM api_key_wallet_permissions p WHERE p.api_key_id = k.id)
),
resolved AS (
  SELECT s.*,
         (SELECT count(*)
            FROM (SELECT w.id
                    FROM custody_wallets w
                    JOIN custody_configs c ON c.id = w.custody_config_id
                   WHERE c.organization_id = s.organization_id
                     AND (c.project_id IS NULL OR c.project_id = s.project_id)
                     AND c.status = 'active' AND w.status = 'active'
                     AND w.wallet_id = s.wallet_id
                  UNION ALL
                  SELECT w.id
                    FROM custody_wallets w
                    JOIN custody_connections c ON c.id = w.custody_connection_id
                   WHERE c.organization_id = s.organization_id
                     AND c.project_id = s.project_id
                     AND c.status = 'active' AND w.status = 'active'
                     AND w.wallet_id = s.wallet_id) candidates
         ) AS candidate_count
    FROM selectors s
)
SELECT api_key_id, organization_id, project_id, wallet_id, source, candidate_count
  FROM resolved
 WHERE candidate_count <> 1
 ORDER BY api_key_id, wallet_id;
```

`candidate_count = 0` — dangling binding; `>= 2` — duplicate wallet id in scope.
Both would hydrate deny-only.
</details>

## Result Validation

- Full API suite: 234 test files passed.
- Tests: 2,843 passed, 14 skipped, 0 failed.
- API typecheck passed.
- Module-boundary checks passed.
- Biome checks passed for all PR files.
- OpenAPI contract tests passed.
- Postman examples were updated.

## Does It Affect Current Flows?

Existing unambiguous Config-backed flows remain unchanged while the feature flag is off. The only intentional compatibility tightening is that ambiguous or unresolved selected-wallet bindings now fail closed instead of falling back.

## Notes

- This PR covers API-key and policy control-plane parity.
- Runtime execution parity for payments, issuance, approval execution, and background jobs remains in HOO-1023.
- API-key permission rows continue to store the Provider `walletId`; an exact local wallet is resolved at authorization time only when the selector has one eligible match.
- Exact Connection intent is nested under `provisionWallet` so previous application revisions reject it before any Provider operation.


pashpashkin:
@greptileai review please

resourcefulmind (changes requested):
1. A key that works today can start 403ing on every wallet route after deploy. If the stored wallet id doesn't resolve to exactly one active custody wallet, the binding gets dropped but the scope stays selected, so a single-wallet key ends up scoped to nothing. Could we count the unresolvable rows in prod first? And log something at the drop site, it's silent right now.

2. Same state leaks the other way: an empty allowed list makes the payments repo drop the wallet filter and return the project's pending transfers. The other call site guards for that.

Also, PRIVY_BYOK_ENABLED off doesn't cover this, nothing on the auth path reads a flag. Rollback is a revert. I wanted to bring this to your attention as well.

pashpashkin:
@resourcefulmind Thanks! all three addressed (`a0c748976`, `51bbf81ee`)

**1. Silent drop / count first.** The drop site now logs
`api_key_wallet_binding_unresolved` (key, wallet id, candidate count), unit-tested.
Ran the read-only preflight (query in the PR description) on **dev and prod: 0 rows**
— no existing key changes behavior at deploy.

**2. Empty allowed list.** Traced every consumer of the allowlist — no reachable
leak today: each path either guards the empty list or throws FORBIDDEN before
querying. But the hazard is real (this PR is what makes `[]` reachable), so it's
fixed at the root: `listTransfers` now compiles `walletIds: []` to `1 = 0` instead
of dropping the filter, pinned by repository tests.

**3. Flag / rollback.** Correct, the auth path would not fork on a
BYOK flag. PR description now states it: rollback = revert, preflight before
deploy, the new warn log as the post-deploy monitor.

resourcefulmind (approved):

PR #1343, opened 2026-08-15
[Superseded] feat(earn): deposit into Kamino K-Vaults from an SDP custody wallet

keeganthomp:
> [!NOTE]
> This monolithic draft has been superseded by five smaller draft PRs. Review and merge them bottom-up:
>
> 1. #1350 — Kamino SDK and neutral contracts
> 2. #1351 — SQL movement ledger
> 3. #1352 — execution runtime
> 4. #1353 — policy-gated API
> 5. #1354 — dashboard experience
>
> All five remain drafts. The replacement stack's top tree is byte-for-byte identical to this PR's reviewed head. This closed PR is retained only for its historical security and SQL review record.

---

Stacked on #1340 — **base is `claude/provider-surfacing-config-68f58a`, not `main`.** Review that PR first.

## Purpose

Kamino strategies are real catalogue entries, but a K-Vault is not a custodial Earn program and it has no deposit address. Sending stablecoins to the vault account would destroy them: deposits must be expressed as program instructions, signed by an SDP custody wallet, and submitted on chain.

This PR adds that second Earn money model:

1. resolve an admitted catalogue strategy and an exact SDP custody-wallet row;
2. build and independently validate the provider's transaction plan;
3. enforce wallet policy and API-key wallet scope;
4. simulate and sign with the selected custody wallet;
5. atomically persist the signed outbox, movement, position claim, idempotency winner, and approved-operation effect;
6. let only that winner broadcast;
7. read position balances live from chain while keeping SDP's movement audit durable.

Production deposits remain fail-closed. The same shared capability disables the dashboard action and rejects the API until withdrawal, Active-position, live-quote, and recovery prerequisites land.

## Why this is a substantial change

- **No address exists to fund.** Existing Earn providers provision a custodial wallet; Kamino requires an execution path.
- **The persistence model is different.** Many organizations can hold the same public vault, so the global provider-wallet uniqueness used by custodial Earn is invalid. Migration `0058` adds separate `earn_vault_positions` and `earn_vault_movements` tables.
- **The provider SDK is behind a firewall.** `@kamino-finance/klend-sdk` uses an older Solana Kit and brings a large runtime/WASM dependency graph. `packages/sdp-kamino` keeps that implementation out of the neutral `@sdp/earn` contract and hourly catalogue job.
- **Cluster defaults are unsafe.** The SDK can read devnet state while emitting mainnet K-Vault instructions unless program ids are rebound and output is independently checked. Every execution/read now also proves the RPC endpoint's genesis hash.
- **Signed broadcasts need an outbox.** An RPC timeout does not prove that no money moved. Exact signed bytes, signature, and last-valid block height are stored before broadcast so ambiguous outcomes remain observable and replayable.

### `packages/sdp-kamino`

- Kit-native deposit plans and position reads over the pinned Kamino SDK.
- Per-cluster K-Vault, lending, and farms program bindings plus output allowlist checks.
- Required live `assetIdentity` (`depositTokenMint`, `shareMint`) carried with every plan.
- Exact mint-scale validation and canonical accepted amount/floor values; insignificant trailing zeroes are accepted, non-zero sub-atoms are rejected.
- Exact raw share-account reads above `2^53`, finite/non-negative share and valuation guards, bounded RPC deadlines, shared-slot reads, and hydration concurrency capped at four.
- Measured slot durations: mainnet ≈416 ms, devnet ≈265 ms; both differ from klend-sdk's default of 400 ms.
- Withdrawal capability remains withheld until multi-transaction/LUT execution and resumability exist.

### API and PostgreSQL

- `POST /v1/earn/vault-deposits` and `GET /v1/earn/vault-positions`.
- Route order: global `earn:write` + `wallets:read` → wallet policy → selected-wallet `earn:write` → handler.
- Request identity is the exact `custodyWalletId` (`cwlt_…`), not a provider-local wallet id or public key; ambiguous provider ids fail closed.
- Production capability, strategy status/surfacing/entitlement, catalogue admission, cluster, live asset identity, signer address, requested amount, accepted amount, and slippage-floor semantics all fail closed before signing.
- Required request fingerprint returns `409` on key reuse with a different wallet/vault/environment/amount/floor. A memo makes independently keyed otherwise-identical Solana messages distinct.
- Build → simulate → sign completes before persistence. One PostgreSQL transaction fences the approved effect and inserts the immutable position identity, signed movement/outbox, and idempotency winner; only the winner broadcasts.
- Ambiguous sends remain `pending`; guarded CAS prevents stale writers from overwriting terminal observations.
- Position reads are project/wallet isolated, keyset-paginated, capped at 100, grouped linearly, and hydrated with bounded concurrency. Live snapshots attach only when provider, cluster, owner, reference, token mint, and share mint all match persisted identity.

Migration `0058` is new-table-only and transactional. It includes:

- nullable forensic project attribution with `ON DELETE SET NULL`, preserving shared organization-wallet claims and movement history;
- a composite movement→position foreign key across tenant/environment/provider/vault/wallet identity;
- immutable asset identity on position re-entry;
- exact positive decimal-string and requested/canonical equality checks;
- strict unsigned-integer block height as bounded `NUMERIC`;
- terminal status metadata checks and an enforced repository transition graph;
- position-first locking for concurrent failed-movement cleanup;
- planner-aligned narrow/global position indexes, live-evidence probes, and deterministic unsettled paging for both `pending` and `submitted` rows.

### Dashboard

- A two-step vault-direct modal opened from Opportunities.
- Production rows remain visible but show an explanatory disabled state; sandbox rows are actionable.
- Token-aware USDC/USDT wallet balances and the exact custody row id are submitted.
- Shared onboarding/provider-access bootstrap uses the linked SDP organization id.
- Mount-level focus trapping/return plus wallet → amount → result autofocus and escaped-focus recovery.
- Pending/submitted copy is explicit and every signed response exposes the cluster-correct explorer link.

## Funds and security invariants

- Wallet policy and selected-wallet binding authorization run before custody signing.
- Catalogue metadata never overrides builder-observed asset identity.
- The policy-approved amount/floor must exactly match the instruction plan's accepted values.
- No unsigned intent or phantom position is persisted after build, simulation, signer, or signing failure.
- Position activation and signed intent persistence are atomic; divergent idempotency losers leave no claim.
- Mint identity cannot be rewritten on re-entry.
- Concurrent failures cannot leave an activated ghost position, and a prior confirmed holding remains active.
- A broadcast exception is **unknown**, never proof of failure.
- `confirmed` and `failed` are terminal states and require their audit metadata.
- Production remains unavailable in both API and web until the exit/recovery work below lands.

## Review outcome

At signed commit [`11bcfee6`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843), this PR has **31 review threads: 30 resolved, 1 deliberately open**.

The open thread is PRO-1690. This PR now provides its durable foundation (exact signed bytes, signature, block height, guarded movement state, and deterministic unsettled index), but it does **not** implement the confirmation/rebroadcast worker or movement-status surface.

Independent final SQL and adversarial fund-flow audits found no remaining merge-blocking P0–P2 findings.

## Verified on devnet, end to end

```text
POST /v1/earn/vault-deposits  → submitted
on-chain                      → finalized, err=null
wallet USDC                   → 20 → 19
shares                        → 1 minted
ledger                        → deposit / submitted / cwlt_795d7f19…
```

The UI flow was also exercised through Privy custody: Opportunities → Deposit → wallet → amount → confirm.

## Required outside this PR

Production is currently fail-closed with `VAULT_DIRECT_DEPOSIT_ENVIRONMENTS = ["sandbox"]`. Before adding `production`:

| # | Requirement | Tracking |
|---|---|---|
| 1 | Configure `SOLANA_DEVNET_RPC_URL` and `SOLANA_MAINNET_RPC_URL`, or ensure the fallback endpoint proves the requested cluster. | Doppler / deploy |
| 2 | Land vault withdrawal, Active-position rendering, tenant-scoped movement status, exact-byte rebroadcast/confirmation, and terminal outbox retention/cleanup. | **PRO-1690** |
| 3 | Compute and display a live, bounded `minSharesOut`; reject stale quotes. Production already rejects deposits without a floor. | **PRO-1691** |
| 4 | Persist in-flight modal attempts and poll durable status across reloads. | **PRO-1692** |
| 5 | Finish USDG/general multi-token balance resolution in the web flow. USDT is handled here. | **PRO-1693** |
| 6 | Propagate an operation-scoped `AbortSignal` through the Earn runtime/provider/RPC stack so the 20-second API deadline cancels underlying work while preserving ambiguous-broadcast semantics. | New follow-up ticket required |
| 7 | Add K-Vault/lending/farms programs to both Kora relay allowlists only if sponsorship is desired, and preserve record-before-broadcast signature recovery. | `sdp-infra` |
| 8 | Apply migration `0058`. It has not been deployed, so all integrity changes remain folded into the unreleased migration. | deploy |

## Validation

```text
full workspace typecheck                 19/19 packages
API build                                passed (including WASM asset)
module boundaries                        passed
Biome on all 54 changed source files     passed
@sdp/kamino                              59 passed, 3 live smoke skipped
@sdp/earn                                141 passed
web Earn + focus                         167 passed
focused API/PostgreSQL funds suite       56 passed
value-moving security conformance        21 passed
full API suite                           2,898 passed, 14 skipped, 1 unrelated auth test flaked
isolated failed auth test rerun           passed
git diff --check                         passed
```

On-chain smoke tests are environment-gated; run them against a Surfpool surfnet that forks mainnet as documented in `packages/sdp-kamino/CLAUDE.md`.

### DIFF
--- a/apps/sdp-api/Dockerfile
+++ b/apps/sdp-api/Dockerfile
@@ -30,6 +30,7 @@ COPY packages/sdp-private-channels/package.json ./packages/sdp-private-channels/
 COPY packages/sdp-spc-escrow/package.json ./packages/sdp-spc-escrow/
 COPY packages/sdp-spc-withdraw/package.json ./packages/sdp-spc-withdraw/
 COPY packages/kit-augment/package.json ./packages/kit-augment/
+COPY packages/sdp-kamino/package.json ./packages/sdp-kamino/
 COPY patches ./patches
 
 RUN --mount=type=cache,id=pnpm-store,target=/pnpm/store \
@@ -53,6 +54,7 @@ COPY packages/sdp-private-channels ./packages/sdp-private-channels
 COPY packages/sdp-spc-escrow ./packages/sdp-spc-escrow
 COPY packages/sdp-spc-withdraw ./packages/sdp-spc-withdraw
 COPY packages/kit-augment ./packages/kit-augment
+COPY packages/sdp-kamino ./packages/sdp-kamino
 
 WORKDIR /repo/apps/sdp-api
 RUN node scripts/build-node.mjs
--- a/apps/sdp-api/src/db/migrations/postgres/0058_earn_vault_direct.sql
+++ b/apps/sdp-api/src/db/migrations/postgres/0058_earn_vault_direct.sql
@@ -1,0 +1,175 @@
+-- Solana Earn: non-custodial ("vault_direct") positions and the movements SDP
+-- initiates into them — the execution era arriving for the second provider
+-- shape (PRO-1634, ADR 0002).
+--
+-- ── Why this is not earn_provider_wallets ──────────────────────────────────
+-- That table models a CUSTODIAL program: SDP asks a provider to provision an
+-- omnibus wallet, the customer funds its address, and one provider wallet is
+-- claimable by exactly one org platform-wide (0056's global unique on
+-- (provider, provider_wallet_ref)).
+--
+-- A K-Vault is the opposite. It is a public, permissionless on-chain vault: it
+-- has no address SDP can hand out (stablecoins sent to the vault's program
+-- account are DESTROYED), SDP custodies nothing, and money moves only when a
+-- wallet SDP can sign for submits an instruction. Crucially, MANY orgs and many
+-- wallets legitimately hold the same vault at the same time — so reusing 0056's
+-- global unique would lock a public vault to whichever org deposited first.
+-- That single constraint difference is why these are separate tables rather
+-- than a nullable column on the existing one.
+--
+-- Conventions inherited from 0055 (earn_program_withdrawals):
+-- * provider is open TEXT (ADR 0001/0002 drift rule) — a row can outlive its
+--   provider's registry entry; allowed values live in code registries.
+-- * Money is decimal strings. No numeric columns anywhere.
+-- * project_id / created_by / initiated_by_key_id are forensic attribution on a
+--   money table even where no wire surface reads them.
+-- * created_at/updated_at are TEXT ISO via sdp_iso_now().
+
+-- ───────────────────────────────────────────────────────────────────────────
+-- 1. The claim set: which custody wallet holds which vault.
+--
+-- NOT a balance. Balances and share counts are read LIVE from chain on every
+-- request (ADR 0002: positions are provider truth, and for a non-custodial
+-- vault the chain IS the provider). This table only records that an org opened
+-- a position, so the reads know which (wallet, vault) pairs to hydrate and the
+-- dashboard can list them without scanning every vault on the shelf.
+-- ───────────────────────────────────────────────────────────────────────────
+CREATE TABLE IF NOT EXISTS earn_vault_positions (
+    id TEXT PRIMARY KEY,
+    organization_id TEXT NOT NULL,
+    project_id TEXT NOT NULL,
+    environment TEXT NOT NULL,
+
+    provider TEXT NOT NULL,
+    -- The vault's own on-chain address (the catalogue's provider_reference).
+    provider_reference TEXT NOT NULL,
+    -- The SDP custody wallet that signs for this position and holds the shares.
+    custody_wallet_id TEXT NOT NULL,
+
+    -- Denormalised from the vault at open time so a position still renders when
+    -- the catalogue row is delisted. Never used to build an instruction — the
+    -- builder always re-reads vault state from chain.
+    share_mint TEXT,
+    token_mint TEXT,
+    label TEXT,
+
+    created_by TEXT,
+    created_at TEXT NOT NULL DEFAULT sdp_iso_now(),
+    updated_at TEXT NOT NULL DEFAULT sdp_iso_now(),
+    -- Set when the position is fully exited. Kept rather than deleted: the
+    -- movement history below references it, and a re-entry reuses the row.
+    closed_at TEXT,
+
+    FOREIGN KEY (organization_id) REFERENCES organizations(id) ON DELETE CASCADE,
+    FOREIGN KEY (project_id) REFERENCES projects(id) ON DELETE CASCADE,
+    -- Deliberately NO cascade: a wallet holding vault shares must not be
+    -- deletable out from under them (same fail-loud rule as 0055's wallet FK).
+    FOREIGN KEY (custody_wallet_id) REFERENCES custody_wallets(id),
+    FOREIGN KEY (created_by) REFERENCES users(id),
+
+    CONSTRAINT earn_vault_positions_environment_check
+        CHECK (environment IN ('sandbox', 'production'))
+);
+
+-- One row per (org, environment, provider, vault, wallet).
+--
+-- Scoped to the ORG and the WALLET — emphatically NOT global on
+-- (provider, provider_reference) the way 0056 scopes a custodial wallet. The
+-- registry is permissionless and public: two orgs holding "Steakhouse USDC" is
+-- normal, and so is one org holding it from two different wallets. A global
+-- unique here would refuse the second org's deposit outright.
+CREATE UNIQUE INDEX IF NOT EXISTS idx_earn_vault_positions_claim
+    ON earn_vault_positions(organization_id, environment, provider, provider_reference, custody_wallet_id);
+
+-- List path: id joins created_at as the deterministic pagination tiebreaker
+-- (same lesson as 0055 and the payment_transfers migrations).
+CREATE INDEX IF NOT EXISTS idx_earn_vault_positions_org_created
+    ON earn_vault_positions(organization_id, environment, created_at DESC, id DESC);
+
+-- ───────────────────────────────────────────────────────────────────────────
+-- 2. The movement ledger: money SDP itself moved.
+--
+-- ADR 0002's rule is "SDP ledgers what SDP INITIATES; SDP reads live what the
+-- provider observes". Ground deposits are unledgered because the customer sends
+-- funds and SDP has no intent moment. Here SDP builds, signs and submits BOTH
+-- directions — so both get a row, written at intent and advanced by guarded CAS
+-- on every observation. Without it, a crash between signing and confirmation is
+-- unrecoverable: the transaction is on chain and SDP has no record it exists.
+-- ───────────────────────────────────────────────────────────────────────────
+CREATE TABLE IF NOT EXISTS earn_vault_movements (
+    id TEXT PRIMARY KEY,
+    organization_id TEXT NOT NULL,
+    project_id TEXT NOT NULL,
+    environment TEXT NOT NULL,
+    position_id TEXT NOT NULL,
+
+    provider TEXT NOT NULL,
+    provider_reference TEXT NOT NULL,
+    custody_wallet_id TEXT NOT NULL,
+
+    direction TEXT NOT NULL,
+    -- 'pending' is SDP-only intent (row exists, nothing signed yet); 'submitted'
+    -- means a signature exists and the transaction is on the wire. Both terminal
+    -- states are chain-observed.
+    status TEXT NOT NULL DEFAULT 'pending',
+
+    -- Decimal strings in the vault token's own units. amount is what was asked
+    -- for; shares is what the chain actually minted or burned, so the two are
+    -- deliberately separate and shares stays NULL until confirmation.
+    amount TEXT,
+    shares TEXT,
+
+    -- Base58 transaction signature, set at submit. A row without one never
+    -- reached the chain and is a terminal failure — no funds moved.
+    signature TEXT,
+    failure_reason TEXT,
+
+    -- Caller idempotency key, REQUIRED. A retried deposit without a stable key
+    -- double-spends, and unlike the custodial create there is no provider-side
+    -- dedupe to fall back on: the chain will happily accept the same transfer
+    -- twice.
+    request_id TEXT NOT NULL,
+
+    created_by TEXT,
+    initiated_by_key_id TEXT,
+
+    created_at TEXT NOT NULL DEFAULT sdp_iso_now(),
+    updated_at TEXT NOT NULL DEFAULT sdp_iso_now(),
+    confirmed_at TEXT,
+
+    FOREIGN KEY (organization_id) REFERENCES organizations(id) ON DELETE CASCADE,
+    FOREIGN KEY (project_id) REFERENCES projects(id) ON DELETE CASCADE,
+    -- No cascade: money history outlives the position row.
+    FOREIGN KEY (position_id) REFERENCES earn_vault_positions(id),
+    FOREIGN KEY (custody_wallet_id) REFERENCES custody_wallets(id),
+    FOREIGN KEY (created_by) REFERENCES users(id),
+
+    CONSTRAINT earn_vault_movements_environment_check
+        CHECK (environment IN ('sandbox', 'production')),
+    CONSTRAINT earn_vault_movements_direction_check
+        CHECK (direction IN ('deposit', 'withdraw')),
+    CONSTRAINT earn_vault_movements_status_check
+        CHECK (status IN ('pending', 'submitted', 'confirmed', 'failed'))
+);
+
+-- SDP-side idempotency lock: one movement per caller key per organization.
+-- Org-scoped rather than position-scoped on purpose — a retry that resolves to
+-- a DIFFERENT position (e.g. the caller corrected the vault) is a different
+-- request and must not silently reuse the key.
+CREATE UNIQUE INDEX IF NOT EXISTS idx_earn_vault_movements_request
+    ON earn_vault_movements(organization_id, request_id);
+
+-- Chain correlation for the confirmation sweep. Partial, because a pending row
+-- legitimately has no signature yet.
+CREATE UNIQUE INDEX IF NOT EXISTS idx_earn_vault_movements_signature
+    ON earn_vault_movements(signature)
+    WHERE signature IS NOT NULL;
+
+-- History for one position, newest first, with the id tiebreaker.
+CREATE INDEX IF NOT EXISTS idx_earn_vault_movements_position_created
+    ON earn_vault_movements(position_id, created_at DESC, id DESC);
+
+-- The sweep's work queue: rows that reached the chain but have no outcome yet.
+CREATE INDEX IF NOT EXISTS idx_earn_vault_movements_unsettled
+    ON earn_vault_movements(status, created_at)
+    WHERE status = 'submitted';
--- a/apps/sdp-api/src/db/migrations/postgres/0058_earn_vault_direct.sql
+++ b/apps/sdp-api/src/db/migrations/postgres/0058_earn_vault_direct.sql
@@ -20,7 +20,8 @@
 -- Conventions inherited from 0055 (earn_program_withdrawals):
 -- * provider is open TEXT (ADR 0001/0002 drift rule) — a row can outlive its
 --   provider's registry entry; allowed values live in code registries.
--- * Money is decimal strings. No numeric columns anywhere.
+-- * Money is decimal strings. The only NUMERIC column is the unsigned protocol
+--   block height, so exact fund amounts round-trip without DB coercion.
 -- * project_id / created_by / initiated_by_key_id are forensic attribution on a
 --   money table even where no wire surface reads them.
 -- * created_at/updated_at are TEXT ISO via sdp_iso_now().
@@ -37,7 +38,9 @@
 CREATE TABLE IF NOT EXISTS earn_vault_positions (
     id TEXT PRIMARY KEY,
     organization_id TEXT NOT NULL,
-    project_id TEXT NOT NULL,
+    -- Forensic attribution, not ownership. Organization fallback wallets may
+    -- hold one claim for movements initiated by multiple projects.
+    project_id TEXT,
     environment TEXT NOT NULL,
 
     provider TEXT NOT NULL,
@@ -49,26 +52,39 @@ CREATE TABLE IF NOT EXISTS earn_vault_positions (
     -- Denormalised from the vault at open time so a position still renders when
     -- the catalogue row is delisted. Never used to build an instruction — the
     -- builder always re-reads vault state from chain.
-    share_mint TEXT,
-    token_mint TEXT,
-    label TEXT,
+    share_mint TEXT NOT NULL,
+    token_mint TEXT NOT NULL,
+    label TEXT NOT NULL,
 
     created_by TEXT,
     created_at TEXT NOT NULL DEFAULT sdp_iso_now(),
     updated_at TEXT NOT NULL DEFAULT sdp_iso_now(),
+    -- A claim is activated only in the transaction that durably records the
+    -- signed transaction that can create the holding. Failed preflight and
+    -- idempotency losers create no position; rolled-back claims never render.
+    activated_at TEXT,
     -- Set when the position is fully exited. Kept rather than deleted: the
     -- movement history below references it, and a re-entry reuses the row.
     closed_at TEXT,
 
     FOREIGN KEY (organization_id) REFERENCES organizations(id) ON DELETE CASCADE,
-    FOREIGN KEY (project_id) REFERENCES projects(id) ON DELETE CASCADE,
+    FOREIGN KEY (project_id) REFERENCES projects(id) ON DELETE SET NULL,
     -- Deliberately NO cascade: a wallet holding vault shares must not be
     -- deletable out from under them (same fail-loud rule as 0055's wallet FK).
     FOREIGN KEY (custody_wallet_id) REFERENCES custody_wallets(id),
     FOREIGN KEY (created_by) REFERENCES users(id),
 
     CONSTRAINT earn_vault_positions_environment_check
-        CHECK (environment IN ('sandbox', 'production'))
+        CHECK (environment IN ('sandbox', 'production')),
+    CONSTRAINT earn_vault_positions_movement_identity_key
+        UNIQUE (
+            id,
+            organization_id,
+            environment,
+            provider,
+            provider_reference,
+            custody_wallet_id
+        )
 );
 
 -- One row per (org, environment, provider, vault, wallet).
@@ -81,10 +97,30 @@ CREATE TABLE IF NOT EXISTS earn_vault_positions (
 CREATE UNIQUE INDEX IF NOT EXISTS idx_earn_vault_positions_claim
     ON earn_vault_positions(organization_id, environment, provider, provider_reference, custody_wallet_id);
 
--- List path: id joins created_at as the deterministic pagination tiebreaker
--- (same lesson as 0055 and the payment_transfers migrations).
-CREATE INDEX IF NOT EXISTS idx_earn_vault_positions_org_created
-    ON earn_vault_positions(organization_id, environment, created_at DESC, id DESC);
+-- Project-scoped list path. Every read supplies the current project's custody
+-- wallet row ids (plus organization fallbacks), pages by (created_at, id), and
+-- excludes provisional claims that never reached the signed boundary.
+CREATE INDEX IF NOT EXISTS idx_earn_vault_positions_wallet_created
+    ON earn_vault_positions(
+        organization_id,
+        environment,
+        custody_wallet_id,
+        created_at DESC,
+        id DESC
+    )
+    WHERE activated_at IS NOT NULL AND closed_at IS NULL;
+
+-- Complement the narrow-wallet index above for pages spanning many custody
+-- rows: ORDER BY can stream globally before applying the wallet-array filter.
+CREATE INDEX IF NOT EXISTS idx_earn_vault_positions_created_wallet
+    ON earn_vault_positions(
+        organization_id,
+        environment,
+        created_at DESC,
+        id DESC,
+        custody_wallet_id
+    )
+    WHERE activated_at IS NOT NULL AND closed_at IS NULL;
 
 -- ───────────────────────────────────────────────────────────────────────────
 -- 2. The movement ledger: money SDP itself moved.
@@ -92,14 +128,16 @@ CREATE INDEX IF NOT EXISTS idx_earn_vault_positions_org_created
 -- ADR 0002's rule is "SDP ledgers what SDP INITIATES; SDP reads live what the
 -- provider observes". Ground deposits are unledgered because the customer sends
 -- funds and SDP has no intent moment. Here SDP builds, signs and submits BOTH
--- directions — so both get a row, written at intent and advanced by guarded CAS
--- on every observation. Without it, a crash between signing and confirmation is
--- unrecoverable: the transaction is on chain and SDP has no record it exists.
+-- directions — so both get a row, written with the signed bytes before broadcast
+-- and advanced by guarded CAS on every observation. Without it, a crash between
+-- signing and confirmation is unrecoverable: the transaction may be on chain
+-- while SDP has no record it exists.
 -- ───────────────────────────────────────────────────────────────────────────
 CREATE TABLE IF NOT EXISTS earn_vault_movements (
     id TEXT PRIMARY KEY,
     organization_id TEXT NOT NULL,
-    project_id TEXT NOT NULL,
+    -- Initiating project attribution; history survives project deletion.
+    project_id TEXT,
     environment TEXT NOT NULL,
     position_id TEXT NOT NULL,
 
@@ -108,20 +146,30 @@ CREATE TABLE IF NOT EXISTS earn_vault_movements (
     custody_wallet_id TEXT NOT NULL,
 
     direction TEXT NOT NULL,
-    -- 'pending' is SDP-only intent (row exists, nothing signed yet); 'submitted'
-    -- means a signature exists and the transaction is on the wire. Both terminal
-    -- states are chain-observed.
+    -- 'pending' means a signed transaction is durably recorded but is not yet
+    -- known to be on the wire; 'submitted' means broadcast returned. Both remain
+    -- nonterminal until a chain observer establishes the outcome.
     status TEXT NOT NULL DEFAULT 'pending',
 
-    -- Decimal strings in the vault token's own units. amount is what was asked
-    -- for; shares is what the chain actually minted or burned, so the two are
-    -- deliberately separate and shares stays NULL until confirmation.
-    amount TEXT,
+    -- Decimal strings in each mint's own units. Keep the caller's text for the
+    -- audit trail, but ledger `amount` / `min_shares_out` from the provider plan:
+    -- those are canonical to mint precision and are what the signed instruction
+    -- actually encodes.
+    requested_amount TEXT NOT NULL,
+    amount TEXT NOT NULL,
+    requested_min_shares_out TEXT,
+    min_shares_out TEXT,
+    -- Shares is what the chain actually minted or burned and stays NULL until
+    -- confirmation.
     shares TEXT,
 
-    -- Base58 transaction signature, set at submit. A row without one never
-    -- reached the chain and is a terminal failure — no funds moved.
-    signature TEXT,
+    -- Signed outbox payload, recorded atomically with the movement and position
+    -- activation before the bytes can reach the network. This is enough for a
+    -- reconciler to query the exact signature and, while its blockhash remains
+    -- valid, rebroadcast the exact same transaction without re-signing.
+    signature TEXT NOT NULL,
+    signed_transaction TEXT NOT NULL,
+    last_valid_block_height NUMERIC NOT NULL,
     failure_reason TEXT,
 
     -- Caller idempotency key, REQUIRED. A retried deposit without a stable key
@@ -157,9 +205,24 @@ CREATE TABLE IF NOT EXISTS earn_vault_movements (
     confirmed_at TEXT,
 
     FOREIGN KEY (organization_id) REFERENCES organizations(id) ON DELETE CASCADE,
-    FOREIGN KEY (project_id) REFERENCES projects(id) ON DELETE CASCADE,
-    -- No cascade: money history outlives the position row.
-    FOREIGN KEY (position_id) REFERENCES earn_vault_positions(id),
+    FOREIGN KEY (project_id) REFERENCES projects(id) ON DELETE SET NULL,
+    -- No cascade, and the duplicated movement identity must describe the exact
+    -- parent claim rather than merely point at any position id.
+    FOREIGN KEY (
+        position_id,
+        organization_id,
+        environment,
+        provider,
+        provider_reference,
+        custody_wallet_id
+    ) REFERENCES earn_vault_positions(
+        id,
+        organization_id,
+        environment,
+        provider,
+        provider_reference,
+        custody_wallet_id
+    ),
     FOREIGN KEY (custody_wallet_id) REFERENCES custody_wallets(id),
     FOREIGN KEY (created_by) REFERENCES users(id),
 
@@ -168,7 +231,61 @@ CREATE TABLE IF NOT EXISTS earn_vault_movements (
     CONSTRAINT earn_vault_movements_direction_check
         CHECK (direction IN ('deposit', 'withdraw')),
     CONSTRAINT earn_vault_movements_status_check
-        CHECK (status IN ('pending', 'submitted', 'confirmed', 'failed'))
+        CHECK (status IN ('pending', 'submitted', 'confirmed', 'failed')),
+    CONSTRAINT earn_vault_movements_confirmation_metadata_check
+        CHECK (
+            (status = 'confirmed') =
+            (NULLIF(BTRIM(confirmed_at), '') IS NOT NULL)
+        ),
+    CONSTRAINT earn_vault_movements_failure_metadata_check
+        CHECK (
+            (status = 'failed') =
+            (NULLIF(BTRIM(failure_reason), '') IS NOT NULL)
+        ),
+    CONSTRAINT earn_vault_movements_amount_format_check
+        CHECK (
+            LENGTH(requested_amount) BETWEEN 1 AND 128
+            AND LENGTH(amount) BETWEEN 1 AND 128
+            AND requested_amount ~ '^\d+(\.\d+)?$'
+            AND amount ~ '^\d+(\.\d+)?$'
+            AND requested_amount ~ '[1-9]'
+            AND amount ~ '[1-9]'
+        ),
+    CONSTRAINT earn_vault_movements_amount_identity_check
+        CHECK (
+            CASE
+                WHEN LENGTH(requested_amount) BETWEEN 1 AND 128
+                 AND LENGTH(amount) BETWEEN 1 AND 128
+                 AND requested_amount ~ '^\d+(\.\d+)?$'
+                 AND amount ~ '^\d+(\.\d+)?$'
+                THEN requested_amount::NUMERIC = amount::NUMERIC
+                ELSE FALSE
+            END
+        ),
+    CONSTRAINT earn_vault_movements_floor_pair_check
+        CHECK ((requested_min_shares_out IS NULL) = (min_shares_out IS NULL)),
+    CONSTRAINT earn_vault_movements_floor_identity_check
+        CHECK (
+            CASE
+                WHEN requested_min_shares_out IS NULL AND min_shares_out IS NULL
+                THEN TRUE
+                WHEN requested_min_shares_out IS NOT NULL
+                 AND min_shares_out IS NOT NULL
+                 AND LENGTH(requested_min_shares_out) BETWEEN 1 AND 128
+                 AND LENGTH(min_shares_out) BETWEEN 1 AND 128
+                 AND requested_min_shares_out ~ '^\d+(\.\d+)?$'
+                 AND min_shares_out ~ '^\d+(\.\d+)?$'
+                 AND requested_min_shares_out ~ '[1-9]'
+                 AND min_shares_out ~ '[1-9]'
+                THEN requested_min_shares_out::NUMERIC = min_shares_out::NUMERIC
+                ELSE FALSE
+            END
+        ),
+    CONSTRAINT earn_vault_movements_last_valid_block_height_check
+        CHECK (
+            last_valid_block_height = TRUNC(last_valid_block_height)
+            AND last_valid_block_height BETWEEN 0 AND 18446744073709551615
+        )
 );
 
 -- SDP-side idempotency lock: one movement per caller key per organization.
@@ -178,17 +295,22 @@ CREATE TABLE IF NOT EXISTS earn_vault_movements (
 CREATE UNIQUE INDEX IF NOT EXISTS idx_earn_vault_movements_request
     ON earn_vault_movements(organization_id, request_id);
 
--- Chain correlation for the confirmation sweep. Partial, because a pending row
--- legitimately has no signature yet.
+-- Chain correlation for the confirmation sweep.
 CREATE UNIQUE INDEX IF NOT EXISTS idx_earn_vault_movements_signature
-    ON earn_vault_movements(signature)
-    WHERE signature IS NOT NULL;
+    ON earn_vault_movements(signature);
 
 -- History for one position, newest first, with the id tiebreaker.
 CREATE INDEX IF NOT EXISTS idx_earn_vault_movements_position_created
     ON earn_vault_movements(position_id, created_at DESC, id DESC);
 
--- The sweep's work queue: rows that reached the chain but have no outcome yet.
+-- Hot existence probe for active-list defense and failed-attempt cleanup.
+CREATE INDEX IF NOT EXISTS idx_earn_vault_movements_position_live_evidence
+    ON earn_vault_movements(position_id)
+    WHERE status IN ('pending', 'submitted', 'confirmed');
+
+-- The sweep's work queue: every signed row without a chain outcome. A broadcast
+-- timeout or crash leaves `pending` WITH a signature, so indexing only
+-- `submitted` would omit precisely the ambiguous rows reconciliation is for.
 CREATE INDEX IF NOT EXISTS idx_earn_vault_movements_unsettled
-    ON earn_vault_movements(status, created_at)
-    WHERE status = 'submitted';
+    ON earn_vault_movements(created_at ASC, id ASC)
+    WHERE status IN ('pending', 'submitted');
--- a/apps/sdp-api/src/db/migrations/postgres/0058_earn_vault_direct.sql
+++ b/apps/sdp-api/src/db/migrations/postgres/0058_earn_vault_direct.sql
@@ -129,6 +129,25 @@ CREATE TABLE IF NOT EXISTS earn_vault_movements (
     -- dedupe to fall back on: the chain will happily accept the same transfer
     -- twice.
     request_id TEXT NOT NULL,
+    -- Canonical fingerprint of the request that wrote this row.
+    --
+    -- The key alone is not enough, and the gap is not hypothetical. Matching on
+    -- `request_id` only cannot distinguish a genuine RETRY from a DIFFERENT
+    -- request wearing the same key: reusing a key with another vault, wallet or
+    -- amount would return 200 carrying the ORIGINAL movement's signature. And
+    -- because the position is claimed as part of the same call, such a request
+    -- would also open a real position row for the new vault and then answer
+    -- with the old vault's transaction — a response that is not merely stale
+    -- but self-contradictory.
+    --
+    -- Storing the fingerprint lets the replay path compare INTENT, not just the
+    -- key, and answer 409 on mismatch. NOT NULL follows 0055's earn convention
+    -- rather than the nullable payments one, because the caller key is required
+    -- on this route: a NULL fingerprint would read as "unclaimed" to
+    -- `resolveIdempotencyReplay` and turn the replay backstop into an
+    -- unrecoverable unique-violation instead of a clean 409.
+    -- Built by `buildEarnVaultDepositFingerprint` (src/lib/idempotency.ts).
+    idempotency_fingerprint TEXT NOT NULL,
 
     created_by TEXT,
     initiated_by_key_id TEXT,
--- a/apps/sdp-api/src/db/repositories/earn-vault.repository.ts
+++ b/apps/sdp-api/src/db/repositories/earn-vault.repository.ts
@@ -1,18 +1,8 @@
 import type { SdpEnvironment } from "@sdp/types";
-import type { AppDb } from "@/db";
+import { type AppDb, asTransactionalClient } from "@/db";
 import { conflict } from "@/lib/errors";
 
-/**
- * Persistence for NON-CUSTODIAL ("vault_direct") Earn positions and the money
- * SDP itself moves into and out of them (migration 0058).
- *
- * Kept out of `earn.repository.ts` deliberately: that file models the custodial
- * program — a provider-provisioned wallet with a fundable address, uniquely
- * claimed platform-wide. These tables model the opposite shape, and the
- * constraint that differs (many orgs may hold the same public vault) is the
- * whole reason they are separate. Mixing them invites someone to "unify" the
- * two uniques, which would lock a permissionless vault to its first depositor.
- */
+/** Persistence for signed, non-custodial Earn vault movements and holdings. */
 
 export function generateEarnVaultPositionId(): string {
   return `earn_vault_position_${crypto.randomUUID()}`;
@@ -28,24 +18,25 @@ export type EarnVaultMovementStatus = "pending" | "submitted" | "confirmed" | "f
 export interface EarnVaultPositionRow {
   id: string;
   organization_id: string;
-  project_id: string;
+  project_id: string | null;
   environment: SdpEnvironment;
   provider: string;
   provider_reference: string;
   custody_wallet_id: string;
-  share_mint: string | null;
-  token_mint: string | null;
-  label: string | null;
+  share_mint: string;
+  token_mint: string;
+  label: string;
   created_by: string | null;
   created_at: string;
   updated_at: string;
+  activated_at: string | null;
   closed_at: string | null;
 }
 
 export interface EarnVaultMovementRow {
   id: string;
   organization_id: string;
-  project_id: string;
+  project_id: string | null;
   environment: SdpEnvironment;
   position_id: string;
   provider: string;
@@ -53,19 +44,18 @@ export interface EarnVaultMovementRow {
   custody_wallet_id: string;
   direction: EarnVaultMovementDirection;
   status: EarnVaultMovementStatus;
-  amount: string | null;
+  requested_amount: string;
+  amount: string;
+  requested_min_shares_out: string | null;
+  min_shares_out: string | null;
   shares: string | null;
-  signature: string | null;
+  signature: string;
+  signed_transaction: string;
+  last_valid_block_height: string;
   failure_reason: string | null;
   request_id: string;
-  /**
-   * Canonical fingerprint of the request that created this row, so a reused
-   * `request_id` carrying a different intent is a 409 rather than a silent
-   * replay of someone else's deposit. Built by
-   * `buildEarnVaultDepositFingerprint`; nullable in the TYPE only to satisfy
-   * `resolveIdempotencyReplay`'s generic, NOT NULL in the schema (0059).
-   */
-  idempotency_fingerprint: string | null;
+  /** Canonical request fingerprint; NOT NULL in migration 0058. */
+  idempotency_fingerprint: string;
   created_by: string | null;
   initiated_by_key_id: string | null;
   created_at: string;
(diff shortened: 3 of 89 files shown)


keeganthomp (commented):
I would request changes on this exact head; GitHub requires a COMMENT because the authenticated account authored the PR. The new custody-signing path bypasses wallet policy/scope controls and does not make chain submission recoverable; mainnet entry is enabled without an SDP exit; and dashboard deposits omit a minimum-share floor while treating nonterminal outcomes as success. The two required API container checks also fail because the new workspace package is excluded from the build context. Details inline (11 P1, 7 P2).

apps/sdp-api/Dockerfile
```diff
@@ -30,6 +30,7 @@ COPY packages/sdp-private-channels/package.json ./packages/sdp-private-channels/
 COPY packages/sdp-spc-escrow/package.json ./packages/sdp-spc-escrow/
 COPY packages/sdp-spc-withdraw/package.json ./packages/sdp-spc-withdraw/
 COPY packages/kit-augment/package.json ./packages/kit-augment/
+COPY packages/sdp-kamino/package.json ./packages/sdp-kamino/
```
keeganthomp, line 33: [P1] Re-include `sdp-kamino` in the Docker build context

Both current API container checks fail at this `COPY`. `apps/sdp-api/Dockerfile.dockerignore` excludes `packages/*` and re-includes the existing workspaces, but not `packages/sdp-kamino`, so BuildKit reports both the package manifest and source as not found. Add `!packages/sdp-kamino` beside the other workspace re-inclusions and rerun Docker Build API and Docker Compose Smoke.

apps/sdp-api/src/routes/earn/index.ts
```diff
@@ -38,6 +39,13 @@ earn.use("*", projectContextMiddleware());
 earn.get("/strategies", requirePermissions("earn:read"), listEarnStrategies);
 earn.get("/strategies/:strategyId", requirePermissions("earn:read"), getEarnStrategy);
 
+// Non-custodial ("vault_direct") positions: SDP builds and signs the deposit
+// from a custody wallet, so unlike /programs there is no provider wallet to
+// provision and no address to fund afterwards. Money-in takes earn:write;
+// the position listing is a plain read.
+earn.post("/vault-deposits", requirePermissions("earn:write"), createEarnVaultDeposit);
```
keeganthomp: [P1] Put vault deposits behind wallet policy enforcement

This route reaches `createOrgSigner` and broadcasts a value-moving transaction without `policyGate`, so wallet deny rules, approvals, amount/asset limits, and destination controls are bypassed. Earn is also absent from `value-moving-conformance.node.test.ts`, so the signing-sink inventory does not catch the omission. Add a policy candidate extractor and gate with replay/effect-start handling, and register this route/service in the conformance inventory.

apps/sdp-api/src/routes/earn/handlers/vault.ts
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  const environment = resolveSdpEnvironment(c);
+  const { auth, wallets } = await resolveScope(c);
+  const projectId = requireProjectId(c);
+
+  // Resolve the strategy first: the caller names a catalogue row, never a raw
+  // vault address. That keeps the deposit target inside what SDP catalogues and
+  // means the admission gates the sync applied still bound this path.
+  const strategy = await getEarnRepository(c).getStrategyById(parsed.data.strategyId);
```
keeganthomp, line 208: [P1] Reuse the strategy admission policy for money-in

The raw lookup only checks existence and environment. In particular, operator-paused/deprecated rows are deliberately retained to stop deposits during an exploit or depeg, while catalogue/program paths filter `status = 'active'`; this path can still fund them by ID. It also bypasses the central hidden/curated-row policy. Extract or reuse one money-in admission helper and require the strategy to be active, visible, and fundable before resolving a wallet.

apps/sdp-api/src/routes/earn/handlers/vault.ts
```diff
@@ -0,0 +1,237 @@
...
+    throw badRequest(
+      `${strategy.provider} is a custodial provider; use POST /v1/earn/programs instead.`
+    );
+  }
+
+  // A mainnet instrument must never be funded from a sandbox project (and vice
+  // versa). `host_cluster` states where the vault actually lives; the sync
+  // already refuses to store a mismatch, so this is the second, per-request
+  // guard rather than the only one.
+  const expectedCluster = environment === "production" ? "mainnet-beta" : "devnet";
```
keeganthomp: [P1] Keep production deposits closed until an exit ships

This branch explicitly admits mainnet for production. Kamino is globally surfaced and entitlement is not environment-scoped, so an entitled production org can deposit real funds even though this PR adds no SDP vault-withdraw path or Active-position surface; the PR's own rollout note says mainnet creation must stay off until those exist. Add a server-side, environment-specific fail-closed capability and mirror it in the UI.

apps/sdp-api/src/routes/earn/handlers/vault.ts
```diff
@@ -0,0 +1,237 @@
...
+    auth.organizationId,
+    "earn",
+    strategy.provider as never,
+    environment === "sandbox"
+  );
+
+  // The wallet must be one SDP can sign for, resolved through the same
+  // permission-checked helper the payments and private-channel deposits use.
+  const walletAddress = resolveWalletAddress(wallets, parsed.data.walletId, "walletId", auth, [
+    "wallets:read",
```
keeganthomp: [P1] Require Earn write authority on the selected wallet

The route requires only global `earn:write`, while this wallet binding is checked for `wallets:read`. A key bound to this wallet for read-only access can therefore spend from it; a key with no bindings need not even hold global `wallets:read`, because the binding assertion becomes a no-op. Match the payments pattern: require global `wallets:read` and require `earn:write` on the selected wallet binding.

apps/sdp-api/src/routes/earn/handlers/vault.ts
```diff
@@ -0,0 +1,237 @@
...
+ * NO provider gate here. This is a READ of money the org already holds, and
+ * ADR 0002's exit-safety asymmetry says un-offering or un-entitling a provider
+ * must never hide a position — it closes the door in, never the door out. The
+ * money-in route above takes both gates; this one deliberately takes neither.
+ */
+export async function listEarnVaultPositions(c: AppContext) {
+  const environment = resolveSdpEnvironment(c);
+  const { auth } = await resolveScope(c);
+  const repo = createPostgresEarnVaultRepository(getDb(c.env));
+  const rows = await repo.listPositions({
```
keeganthomp: [P1] Apply selected-wallet scope before listing positions

This query loads every position for the organization/environment and never intersects it with the API key's wallet bindings. A selected-wallet key with `earn:read` can therefore hydrate and read holdings belonging to other wallets. Derive the allowed wallet IDs for `earn:read` and enforce them in the repository query before any provider reads or response mapping.

apps/sdp-api/src/db/repositories/earn-vault.repository.ts
```diff
@@ -0,0 +1,301 @@
...
+    async createMovement(input) {
+      const id = generateEarnVaultMovementId();
+      const inserted = await db
+        .prepare(
+          `INSERT INTO earn_vault_movements (
+             id, organization_id, project_id, environment, position_id,
+             provider, provider_reference, custody_wallet_id,
+             direction, request_id, amount, created_by, initiated_by_key_id
+           ) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+           ON CONFLICT (organization_id, request_id) DO NOTHING
```
keeganthomp: [P1] Reject an idempotency key reused with a different intent

The conflict key is only `(organization_id, request_id)`, and the replay path returns that row without comparing environment, vault, wallet, amount, or `minSharesOut`. Because the service claims the requested position first, reusing a key for deposit B can return deposit A's signature alongside B's new phantom position. Store and compare a canonical request fingerprint (using the existing idempotency helper), return 409 on mismatch, and preflight replay or make position-claim plus movement creation atomic.

apps/sdp-api/src/services/earn/vault-deposit.service.ts
```diff
@@ -0,0 +1,221 @@
...
+  );
+  if (signer.address !== input.wallet.publicKey) {
+    // The same assertion private-channels makes: a resolved signer that is not
+    // the wallet we priced the deposit for would move someone else's money.
+    throw badRequest("Resolved signing wallet does not match the deposit wallet");
+  }
+
+  let signature: string;
+  try {
+    const submitted = await submitVaultPlan(env, {
```
keeganthomp: [P1] Make the signed transaction recoverable before broadcast

`submitVaultPlan` signs and sends before this movement stores its signature. An accepted-RPC timeout is marked failed, while a crash or DB failure after acceptance leaves `pending` with no chain correlation; simulation/signer exceptions also strand `pending`, and replay merely returns it. There is no confirmer consuming the unsettled index, so `submitted` never settles either. Split prepare/sign/broadcast, durably record the signature/effect start before sending, keep ambiguous sends reconcilable, resume or observe replays, and add immediate confirmation plus a sweep.

apps/sdp-api/src/services/earn/execution-registry.ts
```diff
@@ -0,0 +1,66 @@
...
+ * The known limitation, stated rather than hidden: `SOLANA_RPC_URL` is a
+ * PROCESS-level value while the cluster is PER-REQUEST, so one process can only
+ * serve the cluster its endpoint points at. A production deployment (mainnet
+ * endpoint) therefore cannot build devnet instructions for a sandbox project.
+ * That fails LOUDLY rather than silently: `listKaminoDevnetVaults` proves the
+ * chain by genesis hash before returning anything, so a mismatch is a refusal,
+ * never a confidently wrong result. Serving both clusters from one process is a
+ * follow-up that owes the full env registration.
+ */
+export function resolveClusterRpcUrl(env: Env, _cluster: SolanaCluster): string {
```
keeganthomp: [P1] Resolve the endpoint for the requested cluster

`_cluster` is discarded even though one API process serves sandbox/devnet and production/mainnet requests. Whichever chain `SOLANA_RPC_URL` targets, the other environment builds and reads against the wrong endpoint; the catalogue genesis probe mentioned above is not run on these execution/position paths. Configure per-cluster endpoints and assert their genesis hashes, or explicitly refuse the unsupported environment.

apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx
```diff
@@ -0,0 +1,294 @@
...
+      requestIdRef.current = { key: crypto.randomUUID(), token };
+    }
+    return requestIdRef.current.key;
+  };
+
+  const selectedWallet: CustodyWalletSummary | undefined = useMemo(
+    () => (wallets ?? []).find((wallet) => wallet.id === walletRowId),
+    [wallets, walletRowId]
+  );
+  const walletBalance = selectedWallet ? walletUsdcAmount(selectedWallet) : undefined;
```
keeganthomp: [P2] Resolve the vault's actual deposit token throughout the flow

`strategyToken` returns the portfolio-only USDC/USDT union, so supported USDG Kamino rows are classified as unsupported. This line and `WalletStep` then always read/render USDC, so a USDT vault displays and compares the wrong balance. Resolve an `EarnDepositTokenSymbol` from the strategy mint and parameterize the shared wallet selector/balance accessor by that token.

apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx
```diff
@@ -0,0 +1,294 @@
...
+  // unavailable, and the chain is the real authority. An unreadable balance must
+  // never block a deposit the wallet can actually fund.
+  const overBalance = walletBalance !== undefined && amountNumber > walletBalance;
+
+  async function submit() {
+    if (!selectedWallet || !amountValid) return;
+    setSubmitting(true);
+    setError(null);
+    try {
+      const response = await createEarnVaultDeposit({
```
keeganthomp, line 120: [P1] Require an on-chain minimum-share floor

The dashboard never sends `minSharesOut`. The API leaves it optional and the pinned Kamino SDK selects the legacy deposit instruction when it is absent, so there is no implicit floor: a vault-state change before inclusion can mint materially fewer shares than the user reviewed. Obtain a live quote/exchange rate, apply and display a bounded tolerance with expiry, include the exact floor in the request/fingerprint, and reject missing or stale floors on this money-moving route.

apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx
```diff
@@ -0,0 +1,294 @@
...
+        walletId: selectedWallet.publicKey,
+        amount,
+        requestId: requestIdFor(attemptToken),
+      });
+      if (!response.ok) {
+        setError(response.error ?? t("DashboardEarn.deposit.vaultSubmitError"));
+        return;
+      }
+      const deposit = response.data.data;
+      if (deposit.status === "failed") {
```
keeganthomp, line 133: [P1] Keep a nonterminal deposit as an in-flight attempt

Only `failed` is rejected, so both `pending` replays and normal `submitted` responses enter the success screen, which says the deposit is on-chain and will appear under Active. Neither status is confirmation, the API has no reconciler, and Active has no vault-position consumer. Closing/reloading also loses the only request ID held in the component ref, so an ambiguous retry can use a fresh key and deposit twice. Persist the attempt until terminal, query/poll its status, render each state truthfully, and only announce a holding after confirmation.

apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx
```diff
@@ -0,0 +1,294 @@
...
+            <Button onClick={onClose}>{t("DashboardEarn.deposit.vaultDone")}</Button>
+          </div>
+        </StepSection>
+      </Modal>
+    );
+  }
+
+  if (step === "wallet") {
+    return (
+      <Modal
```
keeganthomp, line 219: [P2] Give the new modal a complete focus lifecycle

The shared `Modal` does not provide initial focus, a focus trap, or return focus, and this flow does not reuse the dedicated behavior in `EarnWithdrawModal`. Keyboard focus can remain on or tab into obscured page content, and it is not restored to the triggering row on close. Move this behavior into the shared modal or generalize the existing Earn focus hook, with a component test.

apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx
```diff
@@ -0,0 +1,294 @@
...
+        isOpen
+        ariaLabel={t("DashboardEarn.deposit.vaultWalletTitle")}
+        onClose={onClose}
+        size="md"
+      >
+        <StepSection title={t("DashboardEarn.deposit.vaultWalletTitle")}>
+          <p className="mb-5 text-sm leading-6 text-secondary">
+            {t("DashboardEarn.deposit.vaultWalletBody")}
+          </p>
+          <WalletStep
```
keeganthomp: [P2] Do not reuse custodial wallet copy for a direct-vault deposit

`WalletStep` always says the program gets a provider-managed deposit address and nothing moves until funds are sent later, directly contradicting this modal's preceding explanation that the deposit is an immediate instruction with no address. Hard-coding `fireblocksEnabled={false}` also tells entitled organizations that Fireblocks is unavailable. Keep the reusable wallet list, but parameterize its deposit-mode copy and real provider availability (or extract neutral, semantic wrappers).

apps/sdp-web/src/app/dashboard/markets/earn/earn-opportunities-table.tsx
```diff
@@ -165,7 +176,14 @@ function OpportunityRow({ strategy }: { strategy: EarnStrategy }) {
         </span>
       </TableCell>
       <TableCell align="right">
-        {depositable.kind === "depositable" ? (
+        {depositable.kind === "depositable" && onVaultDeposit ? (
```
keeganthomp: [P2] Dispatch the row action by deposit style

The parent supplies `onVaultDeposit` to the whole table, and this branch checks only depositability plus callback presence. If a custodial provider is surfaced again, its depositable row opens the vault modal and posts to the vault endpoint instead of entering the custodial wizard. Make the action model carry the route/style discriminant or branch on `earnDepositStyle(strategy.provider)` here.

packages/sdp-kamino/src/sdk.ts
```diff
@@ -0,0 +1,234 @@
...
+ * Returned as a single batch: a deposit touches one vault and creates at most
+ * the user's share ATA, which has always fit one transaction in measurement.
+ * Withdrawals are the multi-batch case (see below).
+ */
+export async function buildKaminoDepositPlan(
+  runtime: KaminoRuntime,
+  input: KaminoDepositInput
+): Promise<KaminoInstructionPlan> {
+  const { client, vault, state, config } = await bindVault(runtime, input.vault);
+  const amount = toDecimal(input.amount, "amount");
```
keeganthomp: [P2] Reject values below the mint's atomic precision

The boundary validates decimal syntax/sign only, while the pinned SDK floors both deposit amount and `minSharesOut` to mint atoms. For a six-decimal mint, `1.0000009` is recorded as requested but only `1.000000` is encoded; a positive sub-atom minimum becomes zero and silently removes its protection. Validate scales after loading token/share mint decimals and persist the canonical accepted amounts.

packages/sdp-kamino/src/sdk.ts
```diff
@@ -0,0 +1,234 @@
...
+    input.owner as Kit2,
+    shares,
+    input.slot as Kit2,
+    reserves,
+    null,
+    null,
+    (input.rentPayer ?? input.owner) as Kit2
+  );
+
+  const instructions = asInstructions([
```
keeganthomp, line 271: [P2] Do not advertise an unsized single-batch withdrawal plan

This flattens every unstake/reserve-withdraw/cleanup instruction into one batch and returns no lookup table, although the pinned SDK documents that multi-reserve exits may need multiple transactions and the package promises transaction-sized batches. Large exits can exceed Solana's packet limit, while the API submitter refuses more than one transaction. Load the vault LUT and compile-size/split at valid protocol boundaries with resumable submission, or withhold the withdrawal method/capability until that contract is implemented.

packages/sdp-kamino/src/sdk.ts
```diff
@@ -0,0 +1,234 @@
...
+ * value is not renders "—" for the value, which is the module rule everywhere
+ * else in Earn and strictly better than a fabricated number.
+ */
+export async function readKaminoPosition(
+  runtime: KaminoRuntime,
+  input: { vault: Address; owner: Address; slot: bigint }
+): Promise<KaminoPosition> {
+  const { vault, state, config } = await bindVault(runtime, input.vault);
+
+  const userShares = await vault.getUserShares(input.owner as Kit2);
```
keeganthomp: [P2] Avoid the parsed `uiAmount` number for share balances

In pinned `klend-sdk` 10.0.0, `getUserShares` reads parsed-token `tokenAmount.uiAmount` as a JavaScript number. Converting that result to a string/`Decimal` here cannot recover lost base units, so large or finely fractional holdings can report rounded shares and derived token value despite the exact decimal-string contract. Read raw `amount` or `uiAmountString`, scale with the share-mint decimals, and test a raw balance above `2^53`.

apps/sdp-api/Dockerfile
keeganthomp: Done in b0423756 — added `!packages/sdp-kamino` to `apps/sdp-api/Dockerfile.dockerignore`, beside the other workspace re-inclusions. Verified locally with `docker build --target builder`, which now completes through `scripts/build-node.mjs` with the package present in the image.

apps/sdp-api/src/routes/earn/index.ts
keeganthomp: Done in b0423756. `POST /vault-deposits` is now `requirePermissions("earn:write", "wallets:read")` → `policyGate({ extract: extractEarnVaultDepositPolicyCandidate })` → handler. The extractor resolves strategy/wallet/amount and the gate enforces before `createOrgSigner` is reached; the handler only ledgers.

You were right that the inventory could not catch this: `valueMovingSourceRoots` did not include `apps/sdp-api/src/services/earn`, so `vault-execution.service.ts`'s signing sinks were invisible. Added that root, registered an `earn` family contract, and added the sink entry — which immediately caught a third sink I had missed (the sponsored path's `signAndSend`).

Candidate uses `operationFamily: "program"` rather than a new `earn` family: family rules are opt-in (`families.length === 0` matches everything), so deny rules, amount limits and approvals still apply, and `program` was already in the union.

apps/sdp-api/src/routes/earn/handlers/vault.ts
keeganthomp: Done in b0423756. Extracted `assertStrategyDepositable` into `routes/earn/handlers/admission.ts` — asserts `status = 'active'` and `isClusterFundableInEnvironment`, replacing the hand-rolled cluster comparison this path had (support.ts explicitly forbids a second copy). Surfacing now goes through `assertEarnProviderSurfaced` too, so it 403s like `/programs` instead of 400ing.

Left visibility/curation OUT deliberately, per routes/earn/CLAUDE.md — browse policy is not an allocation gate, and folding `isHiddenStrategy` in here would freeze existing positions. Say the word if you want curated-away vaults blocked for NEW vault positions specifically; that reads like a real product question rather than a bug.

Pinned by `earn.vault.test.ts` (paused, deprecated, wrong-cluster).

apps/sdp-api/src/routes/earn/handlers/vault.ts
keeganthomp: Done in b0423756. Added `VAULT_DIRECT_DEPOSIT_ENVIRONMENTS` / `isVaultDirectDepositEnabled` to `@sdp/types/provider-access` — `["sandbox"]` today. The route refuses production with 403 before any lookup, and the dashboard reads the SAME constant, so the button and the route cannot disagree.

Put it in `@sdp/types` precisely so it is one shared fact rather than two; opening production is adding `"production"` to that array, which is deliberately one line so the prerequisite work (withdraw route + Active surface) is what gates it, not memory.

Related: `minSharesOut` is now REQUIRED in production (thread on line 104), so lifting this gate cannot also silently ship unprotected deposits.

apps/sdp-api/src/routes/earn/handlers/vault.ts
keeganthomp: Done in b0423756 — matched the payments pattern exactly: router takes `requirePermissions("earn:write", "wallets:read")`, handler asserts `["earn:write"]` on the selected binding.

**One consequence worth flagging before this merges:** no route anywhere currently asserts an `earn:*` scope on a WALLET BINDING — only `payments:*` and `wallets:*` appear. `earn:write` is a legal binding value and enforces correctly, but any existing selected-scope key provisioned with payments-family binding permissions will now 403 here. I followed your instruction literally; if you would rather keep continuity with the money-moving precedent, `payments:write` is the one-word change. Happy to switch.

apps/sdp-api/src/routes/earn/handlers/vault.ts
keeganthomp: Done in b0423756. `getAllowedApiKeyWalletIdsForPermissions(auth, ["earn:read"])` is applied in the repository query, before any provider read, so a scoped key never burns RPC on positions it cannot see.

The trap worth recording: that helper returns PROVIDER wallet ids (`privy_…`) while `earn_vault_positions.custody_wallet_id` is the `cwlt_…` row id, so the handler translates through `scope.wallets` first. Passing the allow-list straight through matches nothing and returns an empty page — a filter that looks like it works and hides everything.

`null` (unbound) applies no filter; `[]` short-circuits to an empty page in the handler, and `listPositions` now THROWS on an empty allow-list rather than letting it degrade into no filter.

apps/sdp-api/src/db/repositories/earn-vault.repository.ts
keeganthomp: Done in b0423756. Added `idempotency_fingerprint` to `earn_vault_movements` (folded into 0058 in place — you confirmed it is not applied anywhere yet), plus `buildEarnVaultDepositFingerprint` in the existing `lib/idempotency.ts` alongside the other per-domain builders, resolved through `resolveIdempotencyReplay` → 409 `CONFLICT`.

The ordering half mattered as much as the check: the replay is now resolved BEFORE `claimPosition`, so a mismatched retry writes nothing at all. Previously it claimed the position first, so reusing a key against a different vault opened a real position row for vault B and then answered with vault A's signature.

`minSharesOut` is in the fingerprint deliberately — the floor is baked into the built instruction, so leaving it out would let a caller reuse a key with a weaker floor and get a silent replay of the stricter original.

Pinned by three tests in `services/earn/vault-deposit.service.test.ts`, including one asserting no second position row exists after the 409.

apps/sdp-api/src/services/earn/execution-registry.ts
keeganthomp: Done in b0423756. `resolveClusterRpcUrl` now reads `SOLANA_DEVNET_RPC_URL` / `SOLANA_MAINNET_RPC_URL` with `SOLANA_RPC_URL` as fallback, and `assertClusterEndpoint` proves the endpoint by GENESIS HASH before anything is built against it — memoised per `(cluster, url)`, so it is one round trip per process rather than per deposit. Failures are cached too: a misconfigured endpoint is a deployment fact, not a transient one.

Added `GENESIS_HASH_BY_CLUSTER` to `@sdp/types` (both `@sdp/earn` and the API need it and neither may depend on the other) — there was previously only a private devnet constant and no mainnet counterpart.

Full env registration done: `env.d.ts`, `API_LOCAL_ENV_KEYS` in `secret-keys.mjs`. `check-env-configurator-drift` passes. **These two vars need adding in Doppler before production** — noted in the PR description.

packages/sdp-kamino/src/sdk.ts
keeganthomp: Done in b0423756. Added `packages/sdp-kamino/src/amounts.ts` — deliberately OUTSIDE the SDK firewall so it is unit-testable without loading klend-sdk. `acceptAtMintScale` refuses any value finer than its mint and returns the canonical encoded form, surfaced as `KaminoInstructionPlan.accepted` (mirrored on `EarnVaultTransactionPlan.accepted`) for the ledger to persist.

Validating the SCALE rather than clamping was deliberate: clamping would make SDP quietly move a different amount than it was asked for, which is the same bug in a friendlier costume. Also closes the sub-atom `minSharesOut` case — it can no longer floor to zero.

Writing the test caught a real flaw in my own first pass: `mintDecimals(undefined)` returned `0` via `Number("")`, i.e. a missing field would have presented as a plausible zero-decimal mint and disabled every check. It now matches digits explicitly and throws.

packages/sdp-kamino/src/sdk.ts
keeganthomp: Taken the second option you offered — the capability is WITHHELD rather than half-built, in b0423756.

Split `EarnVaultWithdrawProvider` / `supportsVaultWithdraw` out of `EarnVaultDirectProvider`, and `KaminoVaultDirectClient` no longer implements `buildVaultWithdrawal`. `buildKaminoWithdrawPlan` stays in the package (proven against surfnet, and the starting point for the real work) but is no longer re-exported from `index.ts`, so its only callers are the smoke tests.

The split is not just bookkeeping: an exit may legitimately need several transactions and a deposit never does, so folding both into one capability made "can build a deposit" silently assert "can build a correctly batched exit". `client.test.ts` pins `supportsVaultWithdraw === false` with a note that batching the plan is how to make it pass, not deleting the test. Batching + LUT + resumable submission tracked as its own ticket.

packages/sdp-kamino/src/sdk.ts
keeganthomp: Confirmed and fixed in b0423756 — I read the pinned SDK to check, and you are exactly right: `getUserShares` → `getUserSharesBalanceSingleVault` sums `getTokenAccountAmount`, which returns `parsed.info.tokenAmount.uiAmount` (`utils/ata.ts:351`), a JS number.

`readKaminoPosition` now computes the UNSTAKED half itself from `tokenAmount.amount` — the exact integer string — summed as `bigint` and scaled by the share mint's decimals. The STAKED half still comes from the SDK, since that one is derived from farm state as an exact `Decimal` and re-implementing it would duplicate the farm lookup for no precision gain.

Tested above 2^53 as you asked: `9007199254740993` base units round-trips exactly, and the test also asserts the contrast (`BigInt(Number(base)) === 9007199254740992n`).

apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx
keeganthomp: Done in b0423756. Lifted the hook out of `earn-withdraw-modal.tsx` into `src/lib/use-modal-focus.ts` and both modals now use it — the withdraw modal keeps its behaviour unchanged, the deposit modal gains initial focus, a trap, and focus return to the Opportunities row that opened it (keyed by `strategy.id`).

Generalized the fallback attribute from `data-earn-withdraw-focus-fallback` to `data-modal-focus-fallback`, updated the trigger in `earn-workspace.tsx` and its assertion in `earn-workspace.unit.test.tsx`. Kept the value-scoped form — a bare attribute lands focus on whichever card rendered first.

Went with the shared-hook option rather than moving it into `Modal` itself: `Modal` is used by flows that do not want a trap, and this keeps the change to the two modals that do.

apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx
keeganthomp: Done in b0423756. `WalletStep` takes a `depositMode` prop (`"custodial"` default, so the wizard is untouched) selecting the note copy — new `walletVaultDirectTitle`/`walletVaultDirectBody` keys say the wallet signs and holds the shares and there is no deposit address, instead of contradicting the body directly above it.

`fireblocksEnabled` is now real: threaded from `markets/earn/page.tsx` via `fetchProviderAvailability` — the same read `deposit/page.tsx` already used — through `EarnWorkspace` → `EarnOpportunitiesPanel` → the modal. Entitled orgs no longer see the locked card.

apps/sdp-web/src/app/dashboard/markets/earn/earn-opportunities-table.tsx
keeganthomp: Done in b0423756. `OpportunityDepositability`'s `depositable` variant now CARRIES the style, and the row branches on `depositable.style === "vault_direct"` rather than on callback presence.

Your framing was the useful bit — callback presence is a fact about the parent's wiring, the style is a fact about the strategy, and only the second can say which endpoint the row may reach. A missing `onVaultDeposit` now disables the button rather than falling through to the custodial link, since landing there would be a wrong answer wearing a working affordance.

Updated the existing assertion in `earn-deposit-model.unit.test.ts` to the new shape.

apps/sdp-api/src/services/earn/vault-deposit.service.ts
keeganthomp: Partly done in b0423756 — **leaving this open**, the sweep is not built.

Done: split \`signVaultPlan\` (signs, does not send) from \`broadcastVaultTransaction\`. The signature is now recorded on the movement **before** the bytes can reach the network, while the row is still \`pending\` — so a crash or a lost response leaves a row naming the exact transaction to look for. A send error no longer marks \`failed\`: that would assert no money moved, which is precisely the claim a timeout cannot support. Only pre-signature failures are terminal. Pinned by three tests, including one asserting the write ordering from inside the broadcast call.

Also: the sponsored path is now deliberately unreachable from the deposit service, because it *cannot* satisfy record-before-broadcast — the relay signs as fee payer after us, so the signature is unknown until the bytes leave. That is a second prerequisite for sponsorship, on top of the Kora allowlist.

Not done, tracked as **PRO-1690**: the confirmation sweep. `idx_earn_vault_movements_unsettled` still has no consumer, so `submitted` never reaches `confirmed`.

apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx
keeganthomp: Partly done in b0423756 — **leaving this open**, the quote is not built.

Done: `minSharesOut` is now **required in production** (400 without it), the SDK-side sub-atom case is closed (a floor below one atom is refused rather than rounding to `0`), and the floor is part of the idempotency fingerprint so a key cannot be reused with a weaker one. Since production is separately fail-closed, that requirement is a guard that becomes load-bearing the moment the environment gate is lifted — the intent is that opening production cannot *also* silently ship unprotected deposits.

Not done, tracked as **PRO-1691**: the live quote, the displayed bounded tolerance, and the expiry with staleness refusal. The dashboard still sends no floor in sandbox.

apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx
keeganthomp: Partly done in b0423756 — **leaving this open**, persistence and polling are not built.

Done, the truthfulness half: `pending` and `submitted` now render as different screens. `submitted` says signed and broadcast, explicitly *not* confirmed; `pending` says SDP could not establish whether it reached the network and warns against retrying blind. Removed the claim that the position appears under Active once it settles — Active reads custodial `earn_provider_wallets` and does not list vault positions at all, so that sentence was false for every status.

Not done, tracked as **PRO-1692**: the `requestId` still lives only in a component ref, so a reload can still mint a fresh key and deposit twice. Polling depends on PRO-1690, since nothing currently advances `submitted`.

apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx
keeganthomp: **Not addressed in this PR** — carried forward as **PRO-1693**, with your analysis in the ticket.

You are right on both counts: `strategyToken` narrows to the portfolio-only `EarnPortfolioToken = "usdc" | "usdt"`, so a supported USDG vault reads as `asset-unsupported`, and both this line and `WalletStep` read USDC unconditionally, so a USDT vault would show the wrong balance.

I left it out because the fix widens a union threaded through the custodial deposit flow (`fundableStrategies`, `strategyDepositEligibility`, `availableTokens`, `singleStrategyAllocation`) — a web-model change rather than a review fix, and the one item here I did not want to land half-done alongside fifteen others. Note the wider `EarnDepositTokenSymbol = "USDC" | "USDG" | "USDT"` already exists in `@sdp/types` and is server-side only today, so the ticket is mostly plumbing it into web.

Nothing renders a wrong balance today — the vault path is USDC-only in sandbox — but that is luck, not design. Happy to pull it into this PR instead if you would rather not merge with it open.

apps/sdp-api/src/services/earn/vault-deposit.service.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  // 2. Claim the position and write the intent row.
+  //
+  // NOTE the two different wallet identifiers, which are easy to confuse and
+  // fail in different ways: `wallet.id` is the SDP row (`cwlt_…`) and is what
+  // the FK on both tables points at, while `wallet.walletId` is the PROVIDER's
+  // own id (`privy_…`) and is what the signing service resolves an adapter by.
+  // Swapping them yields a foreign-key violation here and a "wallet not found"
+  // at signing time.
+  const position = await repo.claimPosition({
```
keeganthomp: [P1] Commit the fenced intent atomically

`claimPosition` and `createMovement` are separate commits after an unlocked preflight. Two concurrent calls using the same `requestId` for different vaults can both miss the replay and claim distinct positions; the loser then 409s on the movement unique, but its position commit survives. The approval-replay path has a second failure from the same boundary: this service never uses `runApprovedWalletOperationEffectTransaction`, so a crash after the intent insert leaves `execution_effect_started_at` null; recovery replays the pending row as 200 and marks the approval completed although nothing was signed. Put claim plus movement in one transaction and couple that transaction to the approved-effect fence; test a divergent concurrent race and interrupted approval execution.

apps/sdp-api/src/services/earn/vault-deposit.service.ts
```diff
@@ -0,0 +1,315 @@
...
+      amount: input.amount,
+      ...(input.minSharesOut === undefined ? {} : { minSharesOut: input.minSharesOut }),
+    });
+  } catch (error) {
+    getLogger().error({ movementId: movement.id, error }, "vault deposit: build failed");
+    return await fail(error instanceof Error ? error.message : "Failed to build the deposit.");
+  }
+
+  // 3. Simulate before signing — cheaper than a landed failure the customer paid for.
+  const simulation = await simulateVaultPlan(env, {
```
keeganthomp: [P1] Do not strand unsigned intents

`simulateVaultPlan` can reject on RPC/blockhash failure rather than return `{ ok: false }`, and `createOrgSigner` plus the address assertion below can also throw outside the terminal-failure path. The provider SDK calls additionally have no request deadline. Each case leaves this movement `pending` without a signature; the replay branch returns that row and never resumes the work, so one transient dependency failure permanently consumes the idempotency key. Bound every external call and make all definitive pre-broadcast failures transition to `failed`, or add a leased resume state; cover rejected simulation, signer failure, and timeout cases.

apps/sdp-api/src/routes/earn/handlers/vault.ts
```diff
@@ -0,0 +1,396 @@
...
+    // Bound, but nothing qualifies: answer empty WITHOUT querying. `null` means
+    // unbound and takes no filter at all; `[]` here would otherwise widen back
+    // to "no filter" in the repository.
+    if (custodyWalletIds.length === 0) {
+      return success(c, { positions: [] });
+    }
+  }
+
+  const repo = createPostgresEarnVaultRepository(getDb(c.env));
+  const rows = await repo.listPositions({
```
keeganthomp: [P1] Restrict every position query to project-scoped wallets

`custodyWalletIds` is supplied only when an API key has explicit wallet bindings. For an unbound key or session it stays undefined, so this query returns every position in the organization/environment, including custody wallets owned by sibling projects. `resolveScope` deliberately contains only the current project's configs plus organization-level fallbacks, so those sibling rows cannot be hydrated, but their position id, vault, internal wallet id, and timestamps are still disclosed. Start with the wallet row IDs from the resolved scope for every caller, then intersect explicit bindings; add a two-project isolation test.

apps/sdp-api/src/services/earn/execution-registry.ts
```diff
@@ -0,0 +1,150 @@
...
+    if (observed !== expected) {
+      throw providerNotConfigured(
+        `The RPC endpoint configured for ${cluster} reports genesis ${observed}, not ${expected}. ` +
+          "Building against it would address the wrong chain — refusing. Set " +
+          `SOLANA_${cluster === "devnet" ? "DEVNET" : "MAINNET"}_RPC_URL to a ${cluster} endpoint.`
+      );
+    }
+  })();
+
+  clusterProofs.set(key, proof);
```
keeganthomp: [P1] Do not cache a transient RPC failure forever

The cached promise includes the transport-error branch above. If the first genesis probe gets a timeout, 429, or brief outage, that rejected promise remains in `clusterProofs` for the process lifetime and every later deposit fails immediately until the instance restarts. The proof also runs before the ledger replay lookup, so even reading an already-created movement is blocked by the poisoned cache. Cache successful proofs (and, if desired, deterministic mismatches), evict transient rejections or give them a bounded TTL, and resolve an idempotent replay before requiring a fresh RPC proof.

apps/sdp-api/src/db/migrations/postgres/0058_earn_vault_direct.sql
```diff
@@ -0,0 +1,194 @@
...
+CREATE UNIQUE INDEX IF NOT EXISTS idx_earn_vault_movements_signature
+    ON earn_vault_movements(signature)
+    WHERE signature IS NOT NULL;
+
+-- History for one position, newest first, with the id tiebreaker.
+CREATE INDEX IF NOT EXISTS idx_earn_vault_movements_position_created
+    ON earn_vault_movements(position_id, created_at DESC, id DESC);
+
+-- The sweep's work queue: rows that reached the chain but have no outcome yet.
+CREATE INDEX IF NOT EXISTS idx_earn_vault_movements_unsettled
```
keeganthomp, line 314: [P2] Include ambiguous broadcasts in the reconciliation queue

The broadcast-error path deliberately leaves a movement `pending` with a signature because the RPC may have accepted it. This partial index includes only `submitted`, so the exact ambiguous state that most needs chain reconciliation is absent from the declared work queue; a crash after recording the signature and before send has the same shape. Before 0058 is applied, index every signed nonterminal row (or introduce an explicit signed/broadcast-unknown state), and use `(created_at, id)` for deterministic sweep paging rather than leading with the predicate's constant `status`.

keeganthomp (commented):
Second pass on the updated sandbox-only contract and every prior reply. I would request changes if GitHub allowed it on a self-authored PR. Production money-in is now fail-closed and CI is green, but the supported sandbox path still has blocking transaction-fencing, unsigned-intent recovery, project-isolation, and RPC-recovery defects. The SQL-specific gaps are the non-atomic claim/movement write, a reconciliation index that omits ambiguous signed rows, and an unbounded position query plus live-RPC fanout. Details inline. I accepted the explicitly deferred production-only web work as tracked follow-up; the existing recovery thread remains open.

apps/sdp-api/src/db/repositories/earn-vault.repository.ts
```diff
@@ -0,0 +1,375 @@
...
+        throw new Error(
+          "listPositions received an empty custodyWalletIds allow-list; the caller must " +
+            "short-circuit to an empty page instead of querying."
+        );
+      }
+
+      const placeholders = scoped?.map(() => "?").join(", ");
+      const result = await db
+        .prepare(
+          `SELECT * FROM earn_vault_positions
```
keeganthomp, line 371: [P2] Bound the position list end to end

This query has no cursor or limit even though its index comment promises keyset pagination. Every returned row is then grouped with repeated array spreads and hydrated through nested unbounded `Promise.all` calls; the Kamino client fans each wallet's references into more concurrent RPC reads. Cost therefore grows from an unbounded database result into quadratic grouping work and an RPC burst on every request. Add a page limit/cursor, an index serving `(organization_id, environment, custody_wallet_id, created_at, id)` for scoped reads, linear grouping, and bounded/batched hydration.

apps/sdp-api/src/services/earn/vault-deposit.service.ts
```diff
@@ -0,0 +1,315 @@
...
+    initiatedByKeyId: input.apiKeyId ?? null,
+  });
+
+  // Lost a race with a concurrent IDENTICAL request (the differing-intent case
+  // threw above and inside createMovement). The winner owns the money movement.
+  if (replayed) {
+    return { position, movement, replayed: true };
+  }
+
+  const fail = async (reason: string) => {
```
keeganthomp: [P2] Keep failed attempts out of the holdings claim set

The position is inserted or reopened before build/simulation/signing, while `fail` updates only the movement. A definitive first-attempt failure therefore leaves a permanent position row that `GET /vault-positions` reports even though the wallet never held shares; a failed re-entry also clears `closed_at`. No writer closes or removes these rows. The claim also receives none of the strategy's `token_mint`, `share_mint`, or label, so those denormalized fields remain null. Preserve the movement audit row, but add an explicit position lifecycle and expose/open it only after an established deposit, while persisting the trusted mint metadata.

apps/sdp-web/src/app/dashboard/markets/earn/deposit/earn-deposit-model.ts
```diff
@@ -111,8 +125,18 @@ export function opportunityDepositability(strategy: EarnStrategy): OpportunityDe
...
 
   const style = earnDepositStyle(strategy.provider);
+  // `vault_direct` becomes depositable when the vault run is enabled: SDP builds
+  // the instruction, signs it with one of the org's own custody wallets, and
+  // submits it (POST /v1/earn/vault-deposits). Nothing is ever sent to an
+  // address — the vault's account is a program account and funds sent there are
+  // lost — so this is true only because a path was BUILT, never by relaxing a
+  // check. `EARN_VAULT_DEPOSITS_ENABLED` is what keeps the affordance and the
+  // flow shipping together.
+  if (style === "vault_direct" && EARN_VAULT_DEPOSITS_ENABLED) {
```
keeganthomp, line 152: [P2] Mirror the production capability gate in the dashboard

This branch checks that a vault flow exists, but never checks `isVaultDirectDepositEnabled` for the selected project environment. Consequently a fundable mainnet Kamino row is explicitly tested as depositable and opens the modal in production, while the API now returns 403 before lookup. That contradicts both the shared-constant comment and the updated PR description. Thread the project environment/capability into this decision, render an explanatory disabled state for production, and invert the current `kamino-production` unit test.

apps/sdp-web/src/app/dashboard/markets/earn/page.tsx
```diff
@@ -30,18 +31,34 @@ export default async function EarnPage() {
...
+  // their own account.
+  let fireblocksEnabled = false;
   try {
     const { projectClient } = await createRequestScopedSdpApiClients();
     if (projectClient) {
       const keysResult = await fetchActiveApiKeys(projectClient.request);
       if (keysResult.ok && keysResult.data) {
         apiKeys = keysResult.data;
       }
+      const availability = await fetchProviderAvailability(projectClient.request, orgId).catch(
```
keeganthomp: [P2] Resolve provider access with the SDP organization id

`orgId` here is Clerk's organization id, but `/v1/organizations/:orgId/provider-access` validates the linked SDP organization id. The established Earn deposit page already resolves that value from `/v1/onboarding/status`. This request therefore 403s and the catch silently leaves `fireblocksEnabled = false`, so entitled organizations still see the locked Fireblocks state this change intended to fix. Reuse/extract the existing onboarding-plus-provider-availability bootstrap and pass `onboarding.organization.id`.

apps/sdp-web/src/lib/use-modal-focus.ts
```diff
@@ -0,0 +1,90 @@
...
+    const focusFrame = window.requestAnimationFrame(() => {
+      contentRef.current
+        ?.querySelector<HTMLElement>('input:not([type="hidden"]):not([disabled])')
+        ?.focus();
+    });
+
+    const trapFocus = (event: KeyboardEvent) => {
+      if (event.key !== "Tab") return;
+      const dialog = contentRef.current?.closest<HTMLElement>('[role="dialog"]');
+      if (!dialog?.contains(document.activeElement)) return;
```
keeganthomp: [P2] Recover focus when the modal panel changes

The effect runs only when `fallbackId` changes. Pressing Continue or Submit removes the focused control while the same modal remains mounted, which leaves focus on `body`; this guard then makes the trap a no-op because the active element is outside the dialog, and the new amount/result panel receives no focus or announcement. Initial focus also searches only for an input, so the result panel has no target. Keep mount-level return focus separate from panel-change autofocus, recover an escaped active element to the first control/heading, and add a DOM test covering wallet to amount to result transitions.

apps/sdp-api/src/routes/earn/handlers/vault.ts
```diff
@@ -0,0 +1,396 @@
...
+          row.provider_reference,
+        ]);
+      }
+      const walletAddresses = await resolveWalletAddresses(c, [...byWallet.keys()]);
+      await Promise.all(
+        [...byWallet.entries()].map(async ([walletId, references]) => {
+          const owner = walletAddresses.get(walletId);
+          if (!owner) return;
+          try {
+            const snapshots = await client.readVaultPositions(earnRuntime(c), {
```
keeganthomp: [P2] Apply the genesis proof to position reads too

The only call to `assertClusterEndpoint` is in the deposit service; this read invokes the executing client directly despite the registry claiming every position read is proved. If a fallback or per-cluster URL targets the other chain, these calls run there and their failures are swallowed below as absent balances; an address collision could hydrate from the wrong deployment. Put the proof in the shared executing-client/runtime boundary, or assert it once here before fanout, so build and read paths cannot drift.

packages/sdp-kamino/src/amounts.ts
```diff
@@ -0,0 +1,84 @@
...
+ * is the same class of bug wearing a friendlier costume. The caller gets a
+ * refusal it can surface, and the number it asked for is the number that moves.
+ */
+export function acceptAtMintScale(field: string, value: string, decimals: number): string {
+  const normalized = value.trim();
+  // `isDecimalString` is the syntax gate and also the SIGN gate: it accepts
+  // digits and at most one dot, so "-1", "1e5" and "Infinity" are all rejected
+  // here rather than needing a separate numeric parse.
+  if (!isDecimalString(normalized)) throw invalidAmount(field, value);
+  if (decimalScale(normalized) > decimals) throw amountTooPrecise(field, value, decimals);
```
keeganthomp: [P2] Ignore insignificant zeroes when checking mint precision

`decimalScale` counts trailing fractional zeroes before canonicalization, so a six-decimal mint rejects exact values such as `1.0000000` and `1.5000000` even though they encode exactly as `1` and `1.5`. That contradicts this module's own test rule that trailing zeroes are noise. Trim insignificant fractional zeroes (or canonicalize exactly) before comparing the meaningful scale, while continuing to reject nonzero sub-atom digits; add cases whose raw scale exceeds the mint only by zeroes.

packages/sdp-kamino/src/client.ts
```diff
@@ -0,0 +1,180 @@
...
+              address: String(account.address),
+              role: Number(account.role),
+            })),
+            // Base64 keeps the contract JSON-safe: a plan may cross a queue or a
+            // log before it is compiled, and a Uint8Array does not survive that.
+            data: Buffer.from(instruction.data ?? new Uint8Array()).toString("base64"),
+          })
+        )
+      ),
+      lookupTables: plan.lookupTables.map(String),
```
keeganthomp: [P2] Preserve and ledger the builder's accepted amounts

`buildKaminoDepositPlan` returns the canonical amounts actually encoded, but `toWire` drops `plan.accepted`; the API then inserts raw `input.amount`. This breaks the neutral plan contract and the resolved-review claim that accepted amounts reach the ledger (current service mocks include `accepted`, which masks the adapter omission). Copy `accepted` across this boundary and persist its amount/floor before signing, with an adapter-to-service test asserting the movement records the canonical encoded value.

apps/sdp-api/src/services/earn/vault-deposit.service.ts
keeganthomp: Addressed in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): Position claim, signed movement/outbox insert, and approved-operation effect fence now commit in one PostgreSQL transaction. Concurrent divergent-idempotency and interrupted-effect regressions prove losers roll back cleanly.

apps/sdp-api/src/services/earn/vault-deposit.service.ts
keeganthomp: Addressed in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): Unsigned intents are no longer persisted: build, simulation, signer resolution, and signing all complete under bounded calls before the atomic insert. Rejected simulation/signer/timeout tests prove pre-sign failures leave no movement or position.

apps/sdp-api/src/routes/earn/handlers/vault.ts
keeganthomp: Addressed in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): Every position read now starts from exact custody-wallet row IDs in the resolved project scope and intersects selected bindings. Two-project and ambiguous-provider-wallet tests prove sibling metadata is not returned.

apps/sdp-api/src/services/earn/execution-registry.ts
keeganthomp: Addressed in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): Durable replay now resolves before chain work. Genesis probes coalesce only for a 30-second trust window, transient rejections evict immediately, and expiry/retry behavior is covered by focused tests.

apps/sdp-api/src/db/migrations/postgres/0058_earn_vault_direct.sql
keeganthomp: Addressed in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): The unsettled partial index now pages deterministically by created_at/id and includes both pending and submitted signed rows, covering ambiguous broadcasts and pre-broadcast crashes.

apps/sdp-api/src/db/repositories/earn-vault.repository.ts
keeganthomp: Addressed in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): Position reads now use bounded keyset pagination, planner-aligned narrow/global indexes, linear grouping, API hydration concurrency 8, and Kamino vault concurrency 4. Pagination and concurrency-shape tests cover the full path.

apps/sdp-api/src/services/earn/vault-deposit.service.ts
keeganthomp: Addressed in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): Positions activate only in the atomic signed-intent transaction, carry immutable live-observed mints plus label, and are deactivated only after serialized last-failure checks. Pre-sign, concurrent-failure, and prior-confirmed-holding regressions pass.

apps/sdp-web/src/app/dashboard/markets/earn/deposit/earn-deposit-model.ts
keeganthomp: Addressed in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): The Opportunities model now consumes the shared environment capability: production vault rows remain visible with an explanatory disabled badge/action, while sandbox remains depositable. Production and sandbox UI tests cover both states.

apps/sdp-web/src/app/dashboard/markets/earn/page.tsx
keeganthomp: Addressed in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): Both Earn pages now reuse a server bootstrap that resolves the linked SDP organization from onboarding status before querying provider access. Success and failure-path tests cover the corrected identity.

apps/sdp-web/src/lib/use-modal-focus.ts
keeganthomp: Addressed in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): Return focus is mount-scoped while panel autofocus is keyed per wallet, amount, and result step; escaped focus is recovered inside the dialog. DOM tests cover every transition and restoration.

apps/sdp-api/src/routes/earn/handlers/vault.ts
keeganthomp: Addressed in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): Vault-position hydration now proves the resolved cluster endpoint before provider fanout and fails closed on proof failure. It also verifies provider, cluster, owner, reference, and both mints before attaching live balances.

packages/sdp-kamino/src/amounts.ts
keeganthomp: Addressed in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): Mint-scale validation now trims insignificant fractional zeroes before checking precision, while retaining exact rejection of nonzero sub-atom digits. Added representable trailing-zero regressions.

packages/sdp-kamino/src/client.ts
keeganthomp: Addressed in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): The Kamino adapter now preserves builder-accepted amount/floor and live asset identity. The service requires exact requested/accepted amount and floor semantics before signing, then persists those canonical values; adapter and DB tests cover the boundary.

apps/sdp-api/src/services/earn/vault-deposit.service.ts
keeganthomp: Updated in [`11bcfee685f5`](https://github.com/solana-foundation/solana-developer-platform/commit/11bcfee685f5103e6320c02a3d5cf7ff93d70843): the movement now atomically persists the exact signed bytes, signature, last-valid block height, approved-effect fence, and position claim before the insert winner broadcasts; the deterministic unsettled index includes both `pending` and `submitted` rows. Ambiguous sends and post-insert crashes remain durable and replayable.

Keeping this thread open for **PRO-1690**: the actual tenant-scoped status/reconciliation surface, exact-byte rebroadcast/confirmation worker, and terminal outbox-payload cleanup are intentionally not implemented in this already-large PR.

keeganthomp:
This monolithic draft is superseded by the five-PR draft stack:

1. #1350 — Kamino SDK and neutral contracts
2. #1351 — SQL movement ledger
3. #1352 — execution runtime
4. #1353 — policy-gated API
5. #1354 — dashboard experience

Coverage was verified immediately before closure: original head `11bcfee685f5103e6320c02a3d5cf7ff93d70843` and top-stack head `d49ff634138947347cc52c150a9b8b21ae00b366` both resolve to Git tree `821af0cca73ea6c82e088ddda89ca4199e802dbb`; their source diff is empty. All five replacement PRs remain drafts and their current aggregate CI states are successful.

Review and merge the stack bottom-up. This original branch is intentionally retained until the stack has landed.

PR #1347, opened 2026-08-16
feat(web): bring your own RPC credentials from the provider page

tobySolutions:
⚠️ **Top of a stack.** Based on #1346, which is based on #1345. It also merges #1344 (the provider page this section hangs off). #1344 has landed. Remaining merge order: **#1345 → #1346 → this**. Each retargets as the one below lands.

The connection model (#1345) and the relay (#1346) landed with no way to reach them: an organization could not add a key of its own without a database client.

The provider's page now lists the organization's own connections and takes a new one (name, network, endpoint, key), then activates or deactivates them. Reached through **server actions over `/internal/dashboard/rpc`**, the way the Privy custody flow reaches its own routes: those routes refuse API keys and require a session, so a browser-visible proxy route would not have worked anyway.

Not only web, despite the title. It also makes `endpointUrl` optional for the two providers whose host is the same for every account, and it moves the connectivity probe into a shared `rpc-probe` module.

The key is **write-only by construction**: the field clears on save, and the response type has no field that could carry a secret back. A stored credential shows only its host and last four characters; a failed check shows the redacted code, never the provider's response.

**A bug the tests caught:** the endpoint hint originally had `{API_KEY}` in the copy. `translate()` reads braces as an interpolation slot and throws at render, so it would have crashed the page. The placeholder renders as an element now.

**Verified:** 7 section tests (submission carries the typed key, the field clears after save, stored credentials render masked, activation, non-admin sees no controls, redacted failure code), 1191/1191 web tests, typecheck clean, module boundaries pass. The detail skeleton grew a block for the new section, and its test asserts the count.

The i18n audit needed four literals baselined: two Solana cluster ids, a placeholder URL, and the key placeholder token. "Production key" was real prose and went into the catalog instead.

Part of HOO-1089. Closes HOO-1090.

#### Screenshots:

Screenshot 2026-08-16 at 01 45 49
Screenshot 2026-08-16 at 03 30 30
Screenshot 2026-08-16 at 03 30 48
Screenshot 2026-08-16 at 04 43 40
Screenshot 2026-08-16 at 04 43 54
Screenshot 2026-08-16 at 04 44 12


resourcefulmind (changes requested):
Read this from the API side up, since the diff carries more backend than the title suggests.

Signer check is where the #1345 lifecycle problem bites a customer. failed is not deactivated, so the relay reads it as unusable and will not fall back, and this PR routes signer check through that resolver. An admin mistyping a key takes POST /v1/wallets/signer-check down org-wide, and the way back is deactivating, which cannot be undone. Should signer check sit inside the fail-closed boundary?

Second, to have it on the record: the body reads web-only, but this also makes endpointUrl optional and reroutes signer check.

Third, members always get "could not be loaded, reload" because the route is admin-only for reads.

Key handling and masking look right.

tobySolutions:
Signer check comes back off the tenant rail. Your framing settled it: it is API-key reachable and organization-wide, and it reads chain state for a platform operation rather than serving the organization's own RPC traffic, so the blast radius of a mistyped key belongs to the relay. `rpc-byok-parity.test.ts` had been asserting the opposite; it now records the exclusion with the reasoning, so wiring the lookup back in fails rather than passing quietly.

Members: fixed. The internal routes are `org:admin` for reads as well as writes, so the page asked on their behalf, got a 403, and rendered it as a failed read. Non-admins get an admin-only state now and the request is not made.

The lifecycle fixes are on #1345 and #1346. The end-to-end test here is the one you asked for: submit, activate, resolve through the service against real Postgres and real encryption. It used to seed the credential `active` and drive the store directly, which is why it passed while the real path was broken. Removing the promotion fails two of its six cases now.

You were right about the description. Updated to say it also makes `endpointUrl` optional for the two providers whose host is the same for every account, and moves the probe into a shared module.

multipletwigs (dismissed):

resourcefulmind (dismissed):

pashpashkin (approved):

PR #1348, opened 2026-08-16
fix(web): paint the route's own skeleton on cold load

tobySolutions:
Refreshing any dashboard route painted the same generic silhouette — five sidebar pills, two cards, one wide block — so it rarely matched the page that followed and the layout jumped when content arrived.

The shell already resolves the correct skeleton two lines from where it renders the fullscreen indicator, so it now passes that through as children along with the route's content width. Padding tracks the settled shell's content section (`px-3 py-5 md:p-6`) so the skeleton and the page occupy the same measure at every breakpoint — a hardcoded `max-w-7xl` was painting it wider than the page on every route.

The two callers with no route in scope — the pre-workspace state and the dedicated interstitial — keep the generic content.

Pre-existing on main. Verified in a browser at 1440x900 and 390x844.

Closes HOO-1118


multipletwigs:
@tobySolutions checking if the loading state is supposed to be like this, where the background isn't white? 

CleanShot 2026-08-17 at 10 12 34@2x

resourcefulmind (approved):
@multipletwigs the sand background isn't from this PR. --sdp-shell-bg is the app shell colour, painted on html, body in globals.css, so every dashboard screen sits on it. Nothing in the diff touches a colour class.

What you're seeing is more likely that the cold load has no raised white panel, where the settled shell wraps its content in one. That's on main already, but this PR matched the width and the shape, so it's the last difference left. Own ticket?

Separately: the settled section only applies px-3 py-5 md:p-6 when it isn't locking the viewport, so on payments, policies, api-keys and wallets the skeleton is padded where the page is flush. Keep as is?

Approving.

PR #1349, opened 2026-08-16
feat(web): add the guidance primitives

tobySolutions:
The dashboard has no callout component, so every block of explanation or warning is hand-rolled from the status tokens and none of them agree on padding, radius, or whether they announce themselves. Doc links are assembled ad hoc at each call site, one against a hardcoded absolute URL, and the docs origin expression is copied into four files.

Adds three primitives following the `badge.tsx` wrapper pattern, plus `lib/docs-url.ts` to resolve the docs origin once. Adopted in the two files nothing else is touching; the remaining copies are in issuance and are left alone.

Two decisions worth a look:

- `Callout` keeps out of the live region by default. Content that renders with the page is already read in document order, so announcing it makes screen readers speak out of sequence. `live` is opt-in for callouts that appear in response to an action.
- `DefinitionTerm` renders the definition as visually hidden text tied by `aria-describedby`, not only inside the tooltip. The tooltip renders its content in a portal on open, so a tooltip-only definition exists for people using a pointer and for nobody else.

The new-tab hint on `DocLink` is a required prop rather than a catalog lookup, which keeps the primitive free of product copy and avoids touching `messages/en/shared.json` while several PRs are open against it.

Verified in a browser at 1440x900: all four variants resolve to real colours via `getComputedStyle` (including `border-info-border`, which had no existing usage), Tab opens the definition and Escape dismisses it.

Closes PRO-1694


resourcefulmind (approved):
Approving

PR #1351, opened 2026-08-16
feat(api): add the Earn vault movement ledger

keeganthomp:
## Stack

1. #1350 — Kamino SDK and neutral contracts
2. **#1351 — SQL movement ledger (this PR)**
3. #1352 — execution runtime
4. #1353 — policy-gated API
5. #1354 — dashboard experience

All five PRs are drafts and merge bottom-up.

## Why the diff looks so large

This is a stacked PR, and #1350 was rebased after this branch was cut. Until #1351 is restacked onto the latest #1350 head, GitHub's comparison temporarily includes earlier Earn UI and Kamino SDK work, plus roughly 2.9k lines of lockfile churn.

The ledger change itself is one commit with 885 additions across four files: the migration, repository, and shared idempotency helper/test. Once the branch is restacked, the inherited changes will disappear from this PR's diff.

## Why

Vault-direct holdings do not fit the existing custodial Earn wallet model: many organizations may hold the same public vault, and an ambiguous Solana send must remain recoverable by exact signed transaction identity. This layer introduces that durable state model independently of the route that will eventually write it.

## What changes

- Add transactional migration `0058` with separate `earn_vault_positions` and `earn_vault_movements` tables.
- Persist exact signed bytes, signature, last-valid block height, request fingerprint, requested/canonical amounts, and asset identity.
- Make project attribution nullable forensic metadata with `ON DELETE SET NULL` so deleting a project cannot erase organization-wallet fund history.
- Add composite tenant/position foreign-key integrity and immutable position mint identity.
- Enforce decimal-string, terminal metadata, and unsigned block-height constraints in PostgreSQL.
- Add repository transition validation, position-first failure locking, keyset pagination, and planner-aligned partial indexes.
- Add a shared idempotency fingerprint/replay helper.

## Safety boundary

This PR creates unused tables and internal persistence APIs only. It registers no route and cannot sign or broadcast a transaction.

## Validation

- API typecheck: passed
- idempotency unit tests: 14 passed
- migration/repository/funds tests passed against real PostgreSQL in the reviewed snapshot
- aggregate CI on this draft: passed


cutbow7 (changes requested):
Found three issues:

1. **[P1] Cross-tenant wallets can be attached to ledger positions.** The repository inserts `organizationId`, `projectId`, `environment`, and `custodyWalletId` independently, while the schema only verifies that each referenced row exists. I reproduced an organization B position referencing organization A’s custody wallet. Validate ownership in the transaction—ideally with an `INSERT … SELECT` over the wallet owner and project—or enforce equivalent composite integrity.

2. **[P2] A failed re-entry reopens a closed position.** `claimPosition` clears `closed_at`, but failure cleanup refuses to deactivate the position if *any* historical confirmed movement exists. A closed position normally has such history, so a failed re-entry remains active and appears in `listPositions`. Preserve the prior closure or scope live evidence to the current activation lifecycle.

3. **[P2] Supplied share quantities are not validated.** `advanceMovement` can seal blank, negative, oversized, or non-decimal `shares` into a confirmed terminal row. Confirmation without shares appears intentional, but any supplied value should satisfy a bounded unsigned-decimal check in both TypeScript and PostgreSQL.

::code-comment{title="[P1] Bind the claim to the wallet tenant" body="These independently bound IDs are not checked against one another. An existing custody wallet from organization A can be inserted into an organization B position, and project/environment mismatches also pass. Validate the project and wallet ownership inside this transaction or enforce equivalent database integrity." file="/Users/cutbow/solana-developer-platform/apps/sdp-api/src/db/repositories/earn-vault.repository.ts" start=247 end=259 priority=1}

::code-comment{title="[P2] Preserve closure across failed re-entry" body="Clearing closed_at here loses the prior lifecycle state. If the new movement fails, cleanup sees the position's historical confirmed movement and leaves it activated with closed_at NULL, producing a ghost reopened position." file="/Users/cutbow/solana-developer-platform/apps/sdp-api/src/db/repositories/earn-vault.repository.ts" start=241 end=242 priority=2}

::code-comment{title="[P2] Validate persisted share quantities" body="When shares is supplied, any string or null is persisted. Because confirmation is terminal, invalid financial data cannot later be corrected through this repository. Add runtime and database validation for non-null values." file="/Users/cutbow/solana-developer-platform/apps/sdp-api/src/db/repositories/earn-vault.repository.ts" start=414 end=417 priority=2}

keeganthomp:
Addressed all three requested changes in c79e39e4:

- Position claims now use an atomic INSERT ... SELECT that binds the supplied organization, project, environment, and custody wallet through the wallet owner. Project-scoped and organization-fallback wallets are both covered.
- Closed positions retain closed_at through a pending re-entry and reopen only after the new deposit confirms. A failed re-entry is no longer returned by listPositions.
- Non-null shares now require a bounded positive unsigned-decimal value in both TypeScript and the PostgreSQL constraint; confirmed movements may still omit shares.
- Added focused PostgreSQL regressions for cross-tenant/project/environment claims, failed and successful re-entry, runtime share validation, and direct-SQL constraint enforcement.

Validation: focused tests 4/4, API typecheck clean, API lint clean. The full API run reached 2,856 passing tests with 8 unrelated Redis cross-file state-isolation failures; the two affected Redis files pass independently (22/22 and 30/30).

cutbow7 (approved):

PR #1358, opened 2026-08-17
fix(policy): reject unmatchable policy rules and ungate the signer check

multipletwigs:
- Add canonical `WALLET_OPERATION_TYPES` (12) and `WALLET_OPERATION_FAMILIES` (4) to `@sdp/types`; every policy rule schema now validates with `z.enum` over them instead of free strings, so a typo'd `operation_type` rule is rejected instead of silently never matching (PRO-1700)
- Drop the producer-less `program` and `provider_admin` families and `raw_sign`/`custody_signer_check` from the vocabulary
- Ungate `POST /v1/wallets/signer-check` from the policy gate: it is an ownership diagnostic (server-authored memo, READONLY_SIGNER, no value moved), and governing it caused the ephemeral-key/replay carve-outs
- Signer check memo is server-generated (`SDP signer check <uuid>`); the `memo` request field is removed so the org wallet never signs caller-chosen text
- Signer check accepts session/Clerk auth with an explicit `walletId` (org-scoped downstream); api-key auth resolves the bound wallet as before
- Kora spend bounded by a fail-closed metered quota (2/min actor, 10/min org) via the standard route-level `meteredQuota`, matching every other paid route
- Known trade-off: the quota charges before body validation, since body parsing lives in handlers today. A malformed body burning a slot is accepted for now because the real fix is moving request validation to the very front of the middleware chain (zod-validator style) — tracked separately — after which middleware ordering makes charge-after-validation the default everywhere
- Dashboard signer check calls the API directly with the session; the mint-key/revoke ephemeral-key flow is deleted
- Migration 0058 purges historical `raw_sign` operation rows (cascades approvals/evaluations) and tightens the family CHECK to the 4 producing families
- Supersedes #1315, #1319, and #1322


resourcefulmind (dismissed):
Spent a while in this one since it touches the policy path. I have a few observations

1. The API key wizard breaks as you use it. Turn on restrictions, type in the operation types box, and it
throws on the first character, because `api-key-authoring` parses that field on every render. Probably wants
to be a picker now the list is fixed.

2. The migration worries me more. `0058` clears old rows by family, but the new parse in the policy repository
reads the operation type, so the two don't line up. Old transfer-prepare rows survive and throw, and the
parse sits in a list map, so one stale row takes out the whole page.

3. It also drops approvals that are still pending. `0055` cancelled them instead, which felt safer. Can we keep
that?

And to say it out loud rather than because I disagree: with the gate off, nothing in a wallet policy stops a
signer check. Makes sense to me, I'd just like it to be a decision.

resourcefulmind (approved):

PR #1364, opened 2026-08-17
feat(helius): add rings repositories

nbaztec:
- 6 repositories (interface + Postgres impl): wallet, operation+timelocks, key-ref, zone, event, health; factory getters + barrel exports
- `reserveIntent`: `ON CONFLICT (intent_key) DO UPDATE SET updated_at = itself RETURNING *` — a replay returns the existing row; the `reserved` flag tells callers whether to run side effects
- State transitions CAS-guarded under `SELECT … FOR UPDATE`; sparse patches never blank earlier columns
- Event payloads redacted on write (shares the registry from the parent PR); key-ref repo never decrypts; timelock release is single-shot
- 60 repository tests against the real testcontainer schema

Notes for review: size is dominated by tests (~900 lines) and mechanical row mapping. The 3-line `pnpm-lock.yaml` change (workspace link) was hand-applied — the registry currently fails `pnpm install --lockfile-only` with `ERR_PNPM_MISSING_TIME` — and verified with `pnpm install --frozen-lockfile`.


nbaztec:
closing in favor of #1388

PR #1370, opened 2026-08-17
fix(tests): serialize the test-database reset with audit-ledger writers

G1de0n:
## Summary

Fixes the recurring `provider-credentials.test.ts` flake (`AuditPersistenceError: External audit-ledger checkpoint did not advance after commit; writes are locked`) that failed Unit Tests on two unrelated PRs today.

**Root cause:** the between-test reset (`seedTestDatabase`) deleted the external audit checkpoint key without holding the ledger's session lock. A background audit write left in flight by a previous test (credential submissions kick off connection checks with their own audit entries) establishes its pre-commit witness in Redis, then commits. `TRUNCATE` already waits for that transaction via its ACCESS EXCLUSIVE lock — but the Redis delete does not, so it could land exactly between the writer's pre-commit witness and its post-commit advance. The post-commit compare-and-set then found the key gone, failed with "did not advance; writes are locked", and the unhandled rejection was attributed to whichever test was running by then.

**Fix:** the reset now runs under the same advisory lock as the ledger writers (`lockedTransactionWithPostCommit` with `AUDIT_LEDGER_SESSION_LOCK_KEY`), truncating inside the transaction and deleting the checkpoint in the post-commit hook while the lock is still held. An in-flight writer therefore always completes its witness protocol before the reset touches the key.

No production code changes — the ledger's fail-closed behavior is working as designed; the test harness was violating its locking contract.

## Testing

- `vitest run src/routes/internal-custody/provider-credentials.test.ts` — 56 passed, five consecutive runs
- `tsc --noEmit` and `biome check` clean


multipletwigs (approved):

multipletwigs:
thank u

PR #1373, opened 2026-08-18
feat(api): move request-body validation to route-level middleware

multipletwigs:
## Summary
- Adds `validateBody`/`validateQuery`/`validateParams` route-level middleware wrapping hono core's `validator` (no new dependency), with typed handler contexts via `ValidatedContext`.
- Migrates ~80 POST/PATCH/PUT/DELETE routes across 16 modules: the schema lives in the route table, handlers read `c.req.valid("json")`. Zero handler-level body parsing remains.
- Metered routes order `requirePermissions → validate* → meteredQuota → policyGate`, so invalid input never charges a fail-closed quota slot (incl. the two metered places GETs via query/param validation).
- Hardening pass closed every initial skip: custody initialize/switch schemas are `z.strictObject` (unknown selector fields rejected, not stripped; `assertNoExistingProviderObjectSelector` deleted), `ramps/:provider/events` replaced by literal per-provider routes with exact schemas, `pay.ts` validates first, recurring cancel/resume get per-route schemas, onboarding complete + api-key revoke get authored schemas.
- sdp-web aligned after a full call-site audit: `Content-Type` only sent with a body, `proxyToSdpApi` no longer forwards empty bodies with a JSON header, private-channels error parsing reads `details.errors`.

## Notes
- Per review: `validateBody` parses the raw body itself instead of delegating to hono's content-type-gated `validator("json")` — a JSON body is never silently dropped into `{}` on a mistyped header (no 200 no-ops on all-optional update schemas), and a bodyless request with a JSON header no longer 400s (empty/whitespace body → `{}`). Mirrors `@hono/zod-validator` otherwise (incl. `safeParseAsync`); the deviation is documented in the JSDoc.
- Behavior changes: malformed JSON returns a 400 `BAD_REQUEST` envelope (previously an unhandled 500); `.catch(() => ({}))` malformed-JSON swallows removed (issuance deploy, subscriptions, api-key rotate/revoke, onboarding); unknown ramp-event providers now 404; invalid-body 400 `details` unified to `{ errors: fieldErrors, formErrors? }` (previously a mix of `treeifyError` shapes).
- All six policy-gate extractors consume the validated body — no double parsing.
- New `validated-body-wiring.node.test.ts` conformance test: every handler typed `ValidatedBodyContext`/`ValidatedContext` must have the matching `validate*(schema)` in its registration chain (97 checks, mutation-tested), so a missed wiring is a red test instead of a runtime 500.

## Test plan
- [x] `tsc --noEmit` clean (sdp-api, sdp-web, sdp-payments)
- [x] `vitest run` — sdp-api 241 files / 2990 tests, sdp-web 147 files / 1213 tests, all green
- [x] `biome check` clean
- [x] Live curl contract checks against a local stack: malformed JSON → 400 envelope, `{}`/headerless body → field errors, strict rejection of unknown custody selectors, bodyless DELETE revoke, unknown ramp provider → 404
- [x] Dashboard click-through: API key create → rotate → delete (strict confirmation), counterparty create (incl. metered places autocomplete) → delete, zero console errors

Closes HOO-1121


eldarik (commented):
**Approve with changes requested on one point** 🟡 — the architecture is right and the hardening pass is a real win; I'd fix the `Content-Type` gate before merge.

Nice work: schema-next-to-route is the correct shape, validating ahead of `meteredQuota`/`policyGate` fixes a real class of wasted fail-closed quota slots, and swapping `assertNoExistingProviderObjectSelector` for a strict `discriminatedUnion` trades hand-rolled prototype-pollution guards for a structural allowlist. No new dependency, one documented `any`, zero handler-level parsing left, and I set-differenced every `ValidatedBodyContext<typeof X>` against every `validate*(X)` — nothing is unwired. 👍

The one thing I'd change is the choice of hono's `validator("json")` as the primitive. It's **content-type-gated** — `if (!contentType || !jsonRegex.test(contentType)) break` leaves `value = {}` and the body is never read (`hono@4.12.25`) — where the old `c.req.json()` parsed regardless. Two consequences fall out:

- **[Risk] A valid JSON body sent without a JSON `Content-Type` is now silently dropped.** On the all-optional update schemas it passes validation and the handler runs with everything `undefined`. `PATCH /v1/api-keys/:keyId` returns **200 `{success:true}`** while the permission/`allowedIps` changes are discarded (cache invalidation doesn't fire either, and the audit row is empty); `rotate` silently falls back to a 24h grace period; same for `updateOrgSchema`, `updateProjectSchema`, `updateTokenSchema`, `updateMemberSchema`… A 200 for a change we threw away is worse than any 400. (Good news — I checked the two dangerous ones: `updateWalletPolicySchema.destinationAllowlist` is required and `updateRecurringPaymentSchema` has a min-keys refine, so neither can be no-op'd into clearing state.)
- **[Risk] The mirror image: a bodyless POST/DELETE that *does* set `Content-Type: application/json` now 400s** "Malformed JSON", because `c.req.json()` throws on an empty body. `readOptionalJsonBody` — which this PR deletes from `recurring-payments.ts` and `subscriptions.ts` — existed for exactly that (`if (!text.trim()) return {}`); the ~40 test call sites that had to gain `body: "{}"` are the tell. Hits `recurring-payments/:id/{activate,cancel,resume,collect}`, subscription `prepare-*`, `rotate`, `DELETE api-keys/:keyId`, the deploy routes, `onboarding/complete`. Your internal audit does hold up — web client, `proxyToSdpApi`, the committed Postman collection (0 bodyless requests with a JSON header) and `docs/` are all clean — so this is external integrators only.

Both die with one ~12-line change in `validateBody`: read the raw body instead of delegating the content-type check, keeping the same `zodValidation` + `addValidatedData("json", …)` shape so no route registration changes.

```ts
const raw = await c.req.text();
let value: unknown = {};
if (raw.trim()) {
  try { value = JSON.parse(raw); }
  catch { throw badRequest("Malformed JSON in request body"); }
}
c.req.addValidatedData("json", parse(value));
await next();
```

Restores the old tolerance, parses whatever the header says, still 400s on genuinely malformed JSON, and `app.ts`'s `HTTPException` normalization stays as a net. (`c.req.text()` fills `bodyCache`, so a downstream `c.req.json()` is fine.) If you'd rather keep hono's validator: make the drop *loud* (415 on a non-empty body with a non-JSON type) and add `.refine(v => Object.keys(v).length > 0)` to the all-optional update schemas.

Then the nits:

- **[P2] `ValidatedBodyContext` is an unenforced promise** — miss a `validate*` and `c.req.valid("json")` is `undefined` at runtime: TypeError → 500, not a compile error. Nothing is wrong today; across 80 routes I'd add an inventory test in the spirit of `wallet-scope-route-coverage.test.ts` so a future miss is a red test.
- **[P2] The 400 `details` shape changed on ~80 endpoints, unversioned** — `treeifyError` → flat `fieldErrors` (counterparties, counterparty-accounts, earn, recurring, subscriptions, asset-profiles), private-channels `details.fieldErrors` → `details.errors`, plus a new `formErrors`. Only in-repo consumer is fixed, but external clients parsing the old shapes fail silently and there's no changelog note.
- **[P3] Invalid requests are now free** — the flip side of the reorder; both remaining test edits (`assetRail: "usdc.solana"`, places `sessionToken`) exist because invalid bodies stopped reaching `meteredQuota`. Worth confirming the app-level rate limiter still fronts those routes.
- **[P3] The `HTTPException` normalization is broader than its case** — it rewrites *any* 400 `HTTPException`, discarding a custom `err.res` and yielding an empty message when none was set. Narrow it to the malformed-body case.
- **[P3] Specific 400 messages regressed** to the generic "Invalid request body": `pay.ts` "account is required", private-channels' four payload messages and "gatewayUrl and chainRpcUrl are required", playground's "Invalid API key".
- **[P3] Two smells** — `updateCounterparty` mutates `c.req.valid("json")` in place (`body.identity = …`), which is now shared request state; and `proxyToSdpApi` maps `""` → `undefined` but not whitespace-only, so a `" "` body still 400s (the deleted `readOptionalJsonBody` trimmed).

eldarik (commented):
Re-reviewed at `d83c05e2` — the `Content-Type` fix is exactly right, and fixed at the root rather than patched around. 👍

`validateBody` now reads the raw body itself, treats empty-or-whitespace as `{}`, and 400s only on genuinely malformed JSON, so both directions are closed: a JSON body sent with `text/plain` is honoured again, and a bodyless POST that *does* send a JSON header no longer 400s. Some touches I didn't ask for and like — `safeParseAsync` so async refinements work, whitespace tolerance (which also retires the `proxyToSdpApi` `" "` nit), and a comment recording *why* this deviates from `@hono/zod-validator` so nobody "fixes" it back. Also the right call skipping the `.refine(keys > 0)` mitigations I floated: with the parse no longer gated on the header, that vector is gone and the refine was only a stopgap.

`validated-body-wiring.node.test.ts` is a better answer to the unenforced-`ValidatedBodyContext` point than I had in mind — walking the route tree and asserting the chain contains the matching `validate*(schema` catches drift at the registration site, and it covers the `policyGate` extract functions for free since `extract:` sits in the same chain slice. `expect(handlers.length).toBeGreaterThan(60)` is the load-bearing line there; glad it's in.

**Approve from my side.** One ask left, and it isn't code:

- **[P2] Changelog the 400 `details` shape change.** Across ~80 endpoints `z.treeifyError`'s nested tree became flat `fieldErrors`, private-channels' `details.fieldErrors` became `details.errors`, and root-level `formErrors` is new. Given the public docs site and the shipped Postman collection, an external client parsing either old shape now silently finds nothing rather than erroring. A release-note bullet covers it.

And explicitly dropping three things I raised, so nobody chases them:

- The metered-quota-bypass worry is a **non-issue** — `app.ts` applies the global rate limiter at the app level, so it still fires ahead of route middleware. Only the per-endpoint quota is spared, which is the whole point of the reorder.
- I overweighted the missing `bodyLimit`: the full-body read already happened for `application/json`, and anyone abusing it would just send that header, so this change adds no real exposure. Pre-existing gap at most.
- `app.ts`'s `HTTPException(400)` branch is now dead for the case its comment describes — `validateBody` throws `AppError` directly, and hono's query/param targets never throw. Harmless as a general net, but the comment is inaccurate now if you're in that file anyway.

multipletwigs:
@eldarik addressed the issues. standardized error on backend.

multipletwigs:
@greptile-apps could you review the frontend standardization as well?

multipletwigs:
@greptile-apps could you review the frontend standardization as well?

multipletwigs:
@greptile-apps please review — the validation 400 contract changed to a single prettified error message (no details envelope); latest commits migrate earn vault-deposits, the shared formatter, and the dashboard consumer.

pashpashkin (approved):

PR #1378, opened 2026-08-18
feat(helius): add rings gateway port and the not-implemented seam

nbaztec:
- Adds `RingsGatewayPort` — the single seam to Zolana/Photon/prover; Track B's B3 implements it as the live HTTP adapter
- Production wiring gets `NotImplementedRingsGateway`: every method throws `gateway_unavailable` naming its seam, so operations end in `failed:gateway_unavailable` and the UI reports the pending integration honestly
- `InMemoryRingsGateway` (deterministic, injectable clock) is test-only: exported via the `@sdp/helius-rings/testing` subpath, never from the package root, never env-selectable — no mocks ship to production
- 13 tests: seam errors per method, determinism, health override, simulated-proof tagging, indexing-delay clock semantics

### DIFF
--- a/packages/sdp-helius-rings/package.json
+++ b/packages/sdp-helius-rings/package.json
@@ -8,6 +8,11 @@
       "types": "./src/index.ts",
       "import": "./src/index.ts",
       "default": "./src/index.ts"
+    },
+    "./testing": {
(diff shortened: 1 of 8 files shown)


nbaztec:
closing in favor of #1388

PR #1379, opened 2026-08-18
feat(helius): map rings operations onto the wallet policy envelope

nbaztec:
- Maps rings operations onto the wallet-operation policy machinery: every rings op evaluates as `operationFamily: "transfer"` with a `rings_*` operation type
- A policy gating transfers on a wallet therefore gates every rings op rooted in it — rings cannot be an approval bypass; verified by test across all op types
- Finer-grained rules can target individual `rings_*` types (e.g. `approval_required` on `rings_transfer_anonymous` only)
- `buildRingsWalletOperationInput` produces the `CreateWalletOperationInput` for the existing enforcement service; intent key doubles as the idempotency key
- No changes to `@sdp/policy` or `@sdp/types` — `operationType` is already an open string


nbaztec:
closing in favor of #1388

PR #1380, opened 2026-08-18
feat(helius): wrap custody signing and RPC submit for rings

nbaztec:
- Thin wrappers so the custody signer and RPC submit path are exercised for real from day 1 (only proof and Photon stay behind the port)
- `signer-adapter.ts`: `createOrgSigner` → signs the gateway-built compiled wire tx (partial and modifying signer branches); base64 in, base64 out; no SecretRef crosses the boundary; tx bytes never logged
- `rpc-adapter.ts`: broadcasts over the shared `@sdp/rpc/solana` `sendTransaction`
- Failures throw `RingsAdapterError` carrying the operation failure code (`signer_failed` / `submit_failed`) and retryability — `WALLET_NOT_FOUND`-class signer errors are non-retryable, submit failures always retryable (the intent key makes resubmission safe)
- 10 unit tests over a fake partial signer and a stub RPC

### DIFF
--- a/apps/sdp-api/src/services/helius-rings/adapter-error.ts
+++ b/apps/sdp-api/src/services/helius-rings/adapter-error.ts
@@ -1,0 +1,22 @@
+import type { FailureCode } from "@sdp/helius-rings";
+
+/**
+ * Failure from the signer or RPC adapter, carrying exactly what the service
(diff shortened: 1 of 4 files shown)


nbaztec:
closing in favor of #1388

PR #1381, opened 2026-08-18
feat(helius): add the rings service orchestrating the operation lifecycle

nbaztec:
- `HeliusRingsService`: the path every rings action travels — provisioning, prepare-through-policy, execution, retry lineage
- Devnet guard at construction (`SOLANA_NETWORK !== "devnet"` refuses); the schema's network CHECK is the other half
- `prepareOperation`: idempotent intent reservation (sha256 intent key) → draft→preparing → policy envelope → deny ends `failed:policy_denied` (non-retryable), approval pauses at `approval_required`, allow runs the pipeline
- Pipeline: gateway build+proof → custody sign → RPC submit → `indexing`, each hop CAS-guarded and recorded on the event feed; adapter failures take their own fail edges
- `executeOperation`: advances an approved operation; polls `verifyIndexed` idempotently until the Photon hit completes it
- `retryOperation`: failed+retryable only; new intent key, `retry_of_operation_id` lineage, original row untouched
- With the production `NotImplementedRingsGateway`, every port call ends in `failed:gateway_unavailable` (retryable) — nothing is simulated
- 19 DB-backed tests covering both gateway variants; key material is deliberately not persisted (that is the key-authority work, Track B4)

### DIFF
--- a/apps/sdp-api/src/services/helius-rings/index.ts
+++ b/apps/sdp-api/src/services/helius-rings/index.ts
@@ -1,0 +1,19 @@
+export { RingsAdapterError } from "./adapter-error";
+export {
+  buildRingsWalletOperationInput,
+  RINGS_ENVELOPE_KINDS,
+  type RingsEnvelopeKind,
+  ringsEnvelopeKind,
+} from "./policy-envelope";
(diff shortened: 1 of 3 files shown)


nbaztec:
closing in favor of #1388

PR #1382, opened 2026-08-18
feat(helius): expose rings over the API behind the feature flag

nbaztec:
- `/v1/helius-rings` routes behind `HELIUS_RINGS_ENABLED` (router-wide 403 gate mirroring private-channels; the service's devnet guard is the second lock)
- Health board, wallet provisioning (503 seam-marker response until the live gateway — the wizard renders it as "integration pending"), zone CRUD, operation lifecycle: prepare / list / detail-with-timeline / execute / retry
- Prepare runs through **real** policy enforcement — route test proves an implicit-allow policy advances the op to the port and ends `failed:gateway_unavailable`, and that enforcement refuses a wallet the tenant doesn't own
- `HeliusRingsError` → `AppError` mapping (`gateway_unavailable`/`config_error` → 503)
- Flag declared in env.d.ts, env-config fields, both `.env.example`s; 10 route tests

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -29,6 +29,9 @@ ASSET_PROFILES_ENABLED=true
 # reconcilers, and the dashboard workspace. Requires
 # SPC_CREDENTIAL_ENCRYPTION_KEY below when enabled.
 PRIVATE_CHANNELS_ENABLED=false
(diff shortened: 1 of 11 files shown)


nbaztec:
closing in favor of #1388

PR #1384, opened 2026-08-18
feat(helius): add the shielded operation composers and zone management

nbaztec:
- One composer for every shielded op kind: shield, send (registered/anonymous), withdraw, merge, timelock — compose → review → confirm
- Anonymous transfers require an explicit acknowledgement on the review step before Confirm unlocks; copy states they are not lower-risk than registered and policy applies identically
- Zone management card (list + create) is fully functional today — zones are SDP-owned metadata, no gateway involved
- Result step shows the operation's landing state honestly (today: `failed — external integration pending`)
- Replaces the interim per-wallet test button


nbaztec:
closing in favor of #1388

PR #1385, opened 2026-08-18
feat(helius): open operations into a detail drawer with the event timeline

nbaztec:
- Activity rows open a right-side drawer: operation identity, state badge, verbatim failure code + message with retryability, and the append-only event timeline (oldest first)
- Timeline payloads were redacted at write time, so the panel cannot surface secret material regardless of what the operation carried


nbaztec:
closing in favor of #1388

PR #1386, opened 2026-08-18
feat(helius): verify approvals server-side, cap retry chains, add recovery UI

nbaztec:
**Server hardening (A21):**
- `executeOperation` now reads the approval verdict from the stored approval request — never from the caller. The previous shape accepted `approvalStatus` in the request body, which would have let anyone with `payments:write` skip a reviewer
- Retries run the full prepare-through-policy path (a retry re-earns its policy verdict, never inherits one) and lineage is capped at 5 — the walk is bounded even against a hypothetical cycle
- New tests: verdict-from-store (pending is inert, approved advances, rejected/canceled/expired fail non-retryably), depth cap

**Recovery UI:**
- Recovery card lists failed operations with a Retry action; retryability and chain limits are enforced server-side and the card relays the verdict
- Composer review step states plainly that operations will record an honest gateway failure while the seam is red

### DIFF
--- a/apps/sdp-api/src/routes/helius-rings/handlers.ts
+++ b/apps/sdp-api/src/routes/helius-rings/handlers.ts
@@ -20,7 +20,6 @@ import {
 import {
   createRingsWalletSchema,
   createRingsZoneSchema,
-  executeRingsOperationSchema,
   listLimitSchema,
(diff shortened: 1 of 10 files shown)


nbaztec:
closing in favor of #1388

PR #1387, opened 2026-08-18
feat(helius): poll rings indexing on a cron, dormant until the live adapter

nbaztec:
- `poll-rings-indexing` cron (every minute): operations in `indexing` poll `verifyIndexed` through the port — a hit completes them; operations stuck past 30 minutes fail as `indexing_timeout` (retryable), so nothing sits in limbo forever
- Ships inert: early-returns unless `HELIUS_RINGS_ENABLED=true` **and** `HELIUS_RINGS_ADAPTER=http` — the live-gateway selector Track B flips. Until then no production operation can reach `indexing` anyway
- `HELIUS_RINGS_ADAPTER` declared across env.d.ts, env-config fields, and both `.env.example`s
- 3 DB-backed job tests: dormant without the selector, completes on a Photon hit, times out past the budget

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -32,6 +32,8 @@ PRIVATE_CHANNELS_ENABLED=false
 # Helius Rings (opt-in, devnet-only). Gates the shielded-wallet API routes
 # and the dashboard workspace.
 HELIUS_RINGS_ENABLED=false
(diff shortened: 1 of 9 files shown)


nbaztec:
closing in favor of #1388

PR #1388, opened 2026-08-18
docs(helius): add the rings operations reference

nbaztec:
- `docs/ops/helius-rings.md`: architecture and the gateway seam, configuration table, the state machine and failure-code reference, idempotency/retry semantics, secret handling, the indexing-poll job, diagnostics, and exactly which surfaces stay dark until the external integration lands


nbaztec:
closing in favor of #1388

PR #1394, opened 2026-08-18
feat(web): let dashboard routes forward opt-in upstream headers

keeganthomp:
## Summary

- `dashboardFetch` accepts an optional `headers`, merged over the inferred `Content-Type`
- `proxyToSdpApi` accepts an optional `upstreamHeaders` forwarded to the upstream client
- add `IDEMPOTENCY_KEY_HEADER` as the one canonical spelling of the header

## Why

The proxy deliberately never copies the inbound header bag — auth, project scope and tracing stay server-owned. That also meant a route had no way to forward the *one* client-owned header a retry-safe write needs. This makes forwarding opt-in per route and per header rather than all-or-nothing, so #1395 can relay `Idempotency-Key` without opening a general passthrough.

Both parameters are optional; every existing caller is unaffected.

## Validation

- 19/19 packages typecheck
- 1,213 web tests (up 3 — new coverage for header merge and upstream forwarding)

## Stack

Split out of #1389 (102 files, +5,939/-8,328). The seven branches together reproduce that branch tree **byte for byte**.

1. #1391 — share program envelopes + payout capability via `@sdp/types`
2. #1392 — Ground client fails closed, exact APY math
3. #1393 — remove the demo seed
4. **#1394 (this PR)** — opt-in upstream headers
5. #1395 — vault deposit/position BFF routes
6. #1396 — retire the prototype, rebuild the Earn program page
7. #1397 — Treasury Solutions

Review in order; each targets the one before it.


multipletwigs (approved):
good add

PR #1396, opened 2026-08-18
feat(earn): retire the Markets prototype and rebuild the Earn program page on live data

keeganthomp:
## Summary

- delete the Markets prototype: the deposit wizard, `earn-workspace`, the opportunities table, the playground, the route skeletons, and their tests (~6,500 lines)
- rewrite the shared Earn modules against live data — `earn-format`, new `earn-decimal` (exact decimal handling), `earn-surfacing`, `earn-program-data`, `earn-market-presentation`, `earn-modal-focus`, `earn-provider-access.server`
- rebuild `/dashboard/markets/earn` on `EarnProgramWorkspace` plus the button builder, reading authenticated Earn APIs and provider capability
- unsupported actions stay visibly disabled with an explanation rather than fabricating positions or success states

## Why

The prototype established the IA but ran on mock browser state and exposed actions independently of real provider and organization capability. This keeps the UI direction and makes the API, database and provider contracts the source of truth.

The old modules are rewritten in place rather than added alongside: `earn-format` drops `formatApy`/`formatUsdCompact`/`formatTokenAmount`/`projectYearlyYield` and narrows the rest to exact strings. Every consumer of those lived inside this directory, so the blast radius is contained — but it does mean the deletion and the rewrite cannot be separated.

Safe to land before #1397 because the whole module is dark: `MARKETS_ENABLED` is off by default.

## Transitional glue

`dashboard-shell.tsx` and `dashboard-navigation-loading.ts` are retargeted at the new skeleton module with a single `earn-program` loading route (~15 lines). #1397 supersedes this with the full Markets IA rename. Isolating it is what lets Treasury review separately.

## Validation

- 19/19 packages typecheck
- 1,136 web tests pass, biome + module boundaries clean

## Known gap

This deletes `markets/earn/CLAUDE.md` (566 lines) — the only `CLAUDE.md` under `apps/sdp-web` — without a replacement, so the web app is left with no directory docs while the architecture materially changed. Worth writing a successor; not included here to keep the diff to the split.

## Stack

Split out of #1389 (102 files, +5,939/-8,328). The seven branches together reproduce that branch tree **byte for byte**.

1. #1391 — share program envelopes + payout capability via `@sdp/types`
2. #1392 — Ground client fails closed, exact APY math
3. #1393 — remove the demo seed
4. #1394 — opt-in upstream headers
5. #1395 — vault deposit/position BFF routes
6. **#1396 (this PR)** — retire the prototype, rebuild the Earn program page
7. #1397 — Treasury Solutions

Review in order; each targets the one before it.


apps/sdp-web/src/app/dashboard/markets/earn/earn-program-workspace.tsx
```diff
@@ -0,0 +1,269 @@
...
+  if (isLoading) return <EarnProgramSkeleton />;
+
+  const rows = strategies ?? [];
+  const selectedStrategy = rows.find((strategy) => strategy.id === selectedStrategyId);
+  const depositsEnabled = isVaultDirectDepositEnabled(sdpEnvironment);
+
+  const continueToBuilder = () => {
+    if (!selectedStrategy) return;
+    router.push(`${builderHref}?strategy=${encodeURIComponent(selectedStrategy.id)}`);
+  };
```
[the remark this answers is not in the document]
keeganthomp: Correct that `createEarnVaultDeposit` has no product caller in this PR — but that is the stack boundary, not an oversight.

This PR is the Earn **Program** page: select a strategy → button builder → server-side integration snippet. The surface that opens positions and moves money is Treasury Solutions, in #1397, which is where each seam gets its caller:

- `createEarnVaultDeposit` → `apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx:483`
- `useEarnVaultPositions` → `apps/sdp-web/src/app/dashboard/markets/treasury-solutions/treasury-solutions-workspace.tsx:588`
- `useEarnPrograms` → same file, `:594`
- `EarnWithdrawModal` → same file, `:737`

So it is four seams landing one PR ahead of their consumer, not one unreachable mutation. Nothing ships half-wired in the meantime: the whole module is dark behind `MARKETS_ENABLED`, which is `flagDefault(..., false)`.

Wiring a deposit modal here instead would duplicate `earn-vault-deposit-modal.tsx` and conflict when #1397 rebases, so I have documented the boundary rather than built a second one. The rewritten `apps/sdp-web/src/app/dashboard/markets/earn/CLAUDE.md` now carries a "Seams without a caller here — do not delete them as dead code" section naming all four, so the next reader does not mistake them for dead exports and remove them.

apps/sdp-web/src/app/dashboard/markets/earn/earn-withdraw-modal.tsx
```diff
@@ -441,15 +425,26 @@ interface EarnWithdrawModalProps {
...
-  // Always USDC: the one stablecoin Ground pays out on Solana, so it is the
-  // right anchor regardless of where the balance happens to sit.
-  const [token, setToken] = useState<EarnPortfolioToken>("usdc");
+  const withdrawTokenOptions = EARN_PORTFOLIO_TOKENS.filter((candidate) =>
+    earnProgramSolanaPayoutTokens(provider).includes(candidate)
+  );
+  const withdrawalAvailable = withdrawTokenOptions.length > 0;
+  // The caller disables providers with no declared Solana payout lane. The
+  // fallback keeps hook state typed; every read/write below still fails closed.
+  const [token, setToken] = useState<EarnPortfolioToken>(withdrawTokenOptions[0] ?? "usdc");
```
multipletwigs: `?? "usdc"` is a banned fallback: with an empty `withdrawTokenOptions` this seeds state with a lane the provider doesn't pay out, and correctness then depends on every downstream read remembering to fail closed. Type the state `EarnPortfolioToken | undefined` (or return the unavailable view before any state exists) so the empty case can't hold a fabricated token.

apps/sdp-web/src/app/dashboard/markets/earn/earn-program-data.ts
```diff
@@ -424,21 +438,147 @@ export async function fetchEarnStrategies(): Promise<EarnStrategy[]> {
...
+  });
+
+  if (!result.ok) return result;
+  const responseBody = result.data;
+
+  if (isObject(responseBody) && isObject(responseBody.data)) {
+    return {
+      ok: true,
+      status: result.status,
+      data: { kind: "submitted", deposit: responseBody.data as unknown as EarnVaultDeposit },
```
multipletwigs: This whole dispatch (516–574) is hand-rolled narrowing — `isObject`, `typeof`, conditional `...(cond ? { x } : {})` spreads — ending in `as unknown as EarnVaultDeposit`, so the deposit record itself is never actually checked. Parse the response at the boundary with a `z.discriminatedUnion` over the success envelope and the `SIGNING_PENDING` error envelope, derive the outcome type via `z.infer`. That deletes `isObject` and both conditional spreads.

apps/sdp-web/src/app/dashboard/markets/earn/deposit/earn-funding-wallets.ts
```diff
@@ -38,21 +23,37 @@ export const EARN_CONNECT_WALLET_HREF = `/dashboard/wallets/setup?provider=${EAR
...
+    !Array.isArray(body.data.wallets)
+  ) {
+    // An invalid success envelope is an upstream failure, not an empty wallet
+    // list. Treating it as [] would disable deposits while claiming the org
+    // simply has no wallets.
+    throw new Error("Invalid custody wallet response");
+  }
   // Only usable funding sources: an inactive wallet cannot originate a transfer.
-  return (body.data?.wallets ?? []).filter((wallet) => wallet.status === "active");
+  return (body.data.wallets as CustodyWalletSummary[]).filter(
```
multipletwigs: The envelope walk above proves `data.wallets` is an array and then casts the rows unchecked. zod-parse the envelope instead and derive the row type — same loud failure, no `as`.

apps/sdp-web/src/app/dashboard/markets/earn/earn-decimal.ts
```diff
@@ -0,0 +1,64 @@
...
+/** Fractional scale, optionally ignoring harmless trailing zeroes. */
+export function unsignedDecimalScale(
+  decimal: Pick<ParsedUnsignedDecimal, "fraction">,
+  { ignoreTrailingZeros = false }: { ignoreTrailingZeros?: boolean } = {}
+): number {
+  return (ignoreTrailingZeros ? decimal.fraction.replace(/0+$/, "") : decimal.fraction).length;
+}
+
+/** Exact ordering for unsigned decimal strings of arbitrary magnitude. */
+export function compareUnsignedDecimals(left: string, right: string): -1 | 0 | 1 | undefined {
```
multipletwigs, line 59: `compareUnsignedDecimals` and `unsignedDecimalScale` re-implement `compareDecimalAmounts` and `decimalScale` from `@sdp/solana/amount`, which this directory already imports. The only genuinely new capability in this file is `parseUnsignedDecimal`'s strict shape + canonical form — keep that, reuse the rest (or move the strict parser into `@sdp/solana/amount` next to its siblings).

apps/sdp-web/src/app/dashboard/markets/earn/earn-format.ts
```diff
@@ -1,66 +1,50 @@
...
  */
 
-export function formatApy(apy: string | number | undefined): string {
-  // Deliberately falsy, not just undefined: an APY of literal 0 (or "0") is a
-  // rate nothing publishes — it means "no rate", and "—" beats a claimed 0.0%.
-  if (!apy) return "—";
-  const value = typeof apy === "string" ? Number(apy) : apy;
-  if (!Number.isFinite(value)) return "—";
-  return `${(value * 100).toFixed(1)}%`;
+function groupedDecimal(
```
multipletwigs: Since ES2023, `Intl.NumberFormat.prototype.format` accepts decimal strings exactly (no float round-trip), and `roundingMode: "trunc"` reproduces this truncation. `new Intl.NumberFormat(locale, { minimumFractionDigits, maximumFractionDigits, roundingMode: "trunc" }).format(value)` replaces the regex grouping — and fixes that this hardcodes en-US commas while `formatProviderApy` next door is locale-aware.

apps/sdp-web/src/app/dashboard/markets/earn/earn-modal-focus.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+  return null;
+}
+
+/**
+ * Shared focus containment and trigger restoration for portaled Earn modals.
+ * Escape remains owned by the common `Modal`, so its close semantics stay
+ * aligned with every other dashboard modal.
+ */
+export function useEarnModalFocus({
```
multipletwigs: A focus trap is generic a11y machinery, not Earn domain. Parked here it will be re-invented by the next modal-heavy module, and the earn-named fallback attributes make it look domain-specific when nothing in it is. Fold it into the DS `Modal` (which already owns Escape) or `components/ui`, with the fallback attribute as a plain parameter.

apps/sdp-web/src/app/dashboard/markets/earn/earn-button-builder.tsx
```diff
@@ -0,0 +1,384 @@
...
+    labelKey: "DashboardMarkets.earnProgram.styleAccent",
+    descriptionKey: "DashboardMarkets.earnProgram.styleAccentDescription",
+  },
+] as const satisfies ReadonlyArray<{
+  value: EarnButtonStyle;
+  labelKey: MessageKey;
+  descriptionKey: MessageKey;
+}>;
+
+if (STYLE_OPTIONS.some((option) => !EARN_BUTTON_STYLES.includes(option.value))) {
```
multipletwigs: Dead runtime check: `value: EarnButtonStyle` in the `satisfies` clause above already proves membership at compile time. Delete these three lines.

apps/sdp-web/src/app/dashboard/markets/earn/earn-button-preview.tsx
```diff
@@ -0,0 +1,37 @@
+"use client";
+
+import { useTranslations } from "@/i18n/provider";
+import { cn } from "@/lib/utils";
+
+export const EARN_BUTTON_STYLES = ["ink", "light", "accent"] as const;
+export type EarnButtonStyle = (typeof EARN_BUTTON_STYLES)[number];
+
+export const EARN_BUTTON_STYLE_CLASS_NAMES: Record<EarnButtonStyle, string> = {
```
multipletwigs: Repo convention for finite maps is `as const satisfies Record<EarnButtonStyle, string>` — the bare `Record` annotation widens the values and drops per-key literals. Same at `earn-withdraw-modal.tsx:78` (`WITHDRAWAL_STATUS_DESCRIPTIONS`).

apps/sdp-web/src/app/dashboard/markets/earn/earn-withdraw-modal.tsx
```diff
@@ -461,21 +456,25 @@ export function EarnWithdrawModal({
...
+      : undefined;
   // What `Max` puts in the box: the offered amount is conservative (whole
   // cents, empirically fillable) while validation below stays permissive
   // against the full reported figure. Offer only what will work; forbid only
   // what certainly will not.
   const maxFillAmount =
     laneLiquidity.phase === "ready" ? floorUsdToCents(laneLiquidity.withdrawableUsd) : undefined;
   const amount = amountInput.trim();
-  const amountShapeValid = USD_AMOUNT_PATTERN.test(amount) && Number(amount) > 0;
+  const amountShapeValid = isPositiveUsdAmount(amount);
```
multipletwigs: `amountShapeValid` / `shapeValid` trip the anti-slop naming rule (no "shape" in symbol names) — `amountFormatValid` says the same thing.

multipletwigs (commented):
Review pass (reuse / parsing / typing / defaults). 9 inline findings.

One PR-level note: after this PR, nothing imports `earn-program-presentation.ts` (its test is deleted too), and the same is true of the whole program write/read surface in `earn-program-data.ts` (`useEarnPrograms`, `createEarnProgram`, `createEarnVaultDeposit`, …), `EarnWithdrawModal`/`EarnWithdrawalOutcomeTracker`, `TreasurySolutionsSkeleton`, and the ~120 `DashboardMarkets.treasury.*` message keys. Fine as a stacked split if #1397 consumes all of it — worth one pass there to delete anything it doesn't.

multipletwigs:
Severity on the review findings: the three worth fixing before flag-on are the `?? "usdc"` lane fallback (earn-withdraw-modal.tsx:447), the hand-parsed vault-deposit response (earn-program-data.ts:516-574), and the unchecked `as CustodyWalletSummary[]` cast (earn-funding-wallets.ts:47). The other six are cleanup and can ride a later pass.

apps/sdp-web/src/app/dashboard/markets/earn/earn-withdraw-modal.tsx
keeganthomp: Done in b61c1c3a. The state is now `EarnPortfolioToken | undefined` and the form subtree renders only once it is a real lane — the token select, amount, Max, available line and preview panel all sit behind `token === undefined ? <unavailable notice> : ...`, so nothing downstream ever receives a fabricated lane and the sub-components keep their non-optional prop types.

Follow-ons: `withdrawalAvailable` is gone entirely (every guard now reads `token === undefined`, including the two preview effects, `amountValid` and `submit`), and `created` carries the token it was made in — `{ withdrawal, token }` — so the confirmation view names the real lane instead of re-reading state that could have been cleared.

apps/sdp-web/src/app/dashboard/markets/earn/earn-program-data.ts
keeganthomp: Done in b61c1c3a. `isObject` and both conditional spreads are gone, and the outcome type is `z.infer` of the parse.

One deviation worth flagging: it is `z.union`, not `z.discriminatedUnion` — the two envelopes share no common key (`{data}` vs `{error}`), so there is no discriminator to switch on. Each branch `.transform()`s to a tagged result, which is what makes the OUTCOME a discriminated union for callers.

The record itself is now checked rather than asserted, which was the real bug: `as unknown as EarnVaultDeposit` let a movement with no `signature` through as a settled deposit. Added a test for that. The schema is annotated `z.ZodType<EarnVaultDeposit>` so a field added or renamed in `@sdp/types` fails typecheck here instead of being silently stripped.

I kept one explicit guard: an approval hold is accepted only on a 202. `dashboardFetch` has already rejected every non-2xx, so a 201 carrying `SIGNING_PENDING` would mean the API called a deposit both created and held — that contradiction must not resolve in the customer's favour. Also tested.

apps/sdp-web/src/app/dashboard/markets/earn/deposit/earn-funding-wallets.ts
keeganthomp: Done in b61c1c3a. The envelope and the rows are zod-parsed and the row type is derived from the schema — `export type EarnFundingWallet = z.infer<typeof earnFundingWalletSchema>` — so there is no `as` and the two cannot drift.

The seam no longer claims `CustodyWalletSummary`. That type carries the `custodyConfigId` XOR `custodyConnectionId` owner union, which is awkward to model faithfully in zod and which this flow never reads; claiming it while validating none of it was the actual lie. `EarnFundingWallet` is what the parse proves: `id`, `walletId`, `publicKey`, `label`, `purpose`, `status`, `isRuntimeExecutionAllowed`, optional `balances` — every field the deposit flows consume.

Worth noting the existing test was passing `{ id, status }` rows and the cast accepted them as complete wallets. Fixtures are realistic now, plus a case asserting a row missing `publicKey` (the address a deposit signs from) fails at this boundary.

apps/sdp-web/src/app/dashboard/markets/earn/earn-decimal.ts
keeganthomp: Done in b61c1c3a. `unsignedDecimalScale` is deleted (its `ignoreTrailingZeros` option had no production caller — only a test), and the withdraw modal now calls `decimalScale` from `@sdp/solana/amount` directly. `parseUnsignedDecimal` stays, with a comment naming exactly what it adds over `isDecimalString`: digits required on both sides of the point, optional no-trim, a length cap, and the canonical form.

`compareUnsignedDecimals` is kept as a thin delegation to `compareDecimalAmounts` rather than deleted, for one reason: the shared comparator throws `AmountError` on a non-decimal, and both call sites read provider strings during render. `laneLiquidity` at `phase: "ready"` can carry `preview.withdrawableUsd` straight off the wire, unvalidated — so `compareUsdDecimals(maxFillAmount, "0")` inside the Max button's `disabled` would crash the modal mid-exit on a malformed figure, where today it disables the button. ADR 0002 (money out beats money off) rules that out, so the wrapper maps the throw to `undefined` and callers treat "cannot compare" as "do not offer". Tested explicitly.

Happy to move the strict parser into `@sdp/solana/amount` next to its siblings if you prefer — say the word and I'll do it as a follow-up rather than widening this PR.

apps/sdp-web/src/app/dashboard/markets/earn/earn-format.ts
keeganthomp: Done in b61c1c3a. The regex grouping is gone; `formatDecimalString` hands the decimal string straight to `Intl.NumberFormat` with `roundingMode: "trunc"`, and every formatter (`formatProviderAmount`, `formatUsd`, `formatTokenQuantity`) now takes `locale` as its second argument, matching `formatProviderApy`. Threaded `useLocale()` through all seven call sites in the withdraw modal.

Verified the swap reproduces the old output exactly, including `9007199254740993.129` past 2^53 and the truncation case (`1.9999999` -> `1.999999`); added a `de-DE` test so the locale fix is actually covered rather than asserted.

One wrinkle: TS types the string overload as `Intl.StringNumericLiteral`, so a plain `string` does not satisfy it. Rather than cast, `isDecimalString` is restated as a type predicate (`isIntlDecimalLiteral`) — the guard that already had to run now also does the narrowing, so there is no `as`.

Left `formatUsd`'s `$` prefix manual rather than switching to `style: "currency"` — that would move the symbol per locale and this copy is English-only. Say if you want that too.

apps/sdp-web/src/app/dashboard/markets/earn/earn-modal-focus.ts
keeganthomp: Done in b61c1c3a — moved to `src/lib/use-modal-focus.ts` as `useModalFocus`, with `git mv` so history follows.

Put it in `src/lib` rather than `components/ui` because that is where this repo already keeps generic hooks, including the direct analogue `use-escape-key.ts`; `components/ui` is all components. `Modal` still owns Escape, as you note.

`fallbackAttribute` is now a plain `string` — the earn-specific closed union is gone, and nothing in the file references Earn any more except one line of comment explaining where it came from and why it left.

Did not fold it into `Modal` itself: the initial-focus selector, the focus key and the restore timing are all per-caller, so absorbing it would mean `Modal` growing four props that only modals with a disappearing trigger use. Easy follow-up if you would rather it live there.

apps/sdp-web/src/app/dashboard/markets/earn/earn-button-builder.tsx
keeganthomp: Done in b61c1c3a — deleted. The `EARN_BUTTON_STYLES` import went with it, since the check was its only use.

apps/sdp-web/src/app/dashboard/markets/earn/earn-button-preview.tsx
keeganthomp: Done in b61c1c3a. `EARN_BUTTON_STYLE_CLASS_NAMES` and `WITHDRAWAL_STATUS_DESCRIPTIONS` both use `as const satisfies Record<...>` now.

Also did `WITHDRAWAL_STATUS_BADGES` two lines above the one you flagged — same finite-map pattern, and leaving one of the pair annotated the old way would have read as deliberate.

apps/sdp-web/src/app/dashboard/markets/earn/earn-withdraw-modal.tsx
keeganthomp: Done in b61c1c3a — `amountShapeValid` -> `amountFormatValid`, and the `AmountError` prop `shapeValid` -> `formatValid`.

multipletwigs (approved):

PR #1403, opened 2026-08-19
feat(api): run the full reconciliation surface on the managed cron job (PRO-1715)

multipletwigs:
Fixes [PRO-1715](https://linear.app/solana-fndn/issue/PRO-1715/run-the-full-reconciliation-surface-on-the-managed-cloud-run-cron-job)

- run recurring-payment collection and the private-channel deposit/withdrawal reconcilers in the Cloud Run job entrypoint (`src/job.ts`) — previously they never ran on managed deployments
- isolate job ticks — this changes the failure semantics of the whole reconciliation job (before/after below)
- remove `PAYMENTS_RECURRING_ENABLED` / `PAYMENTS_RECURRING_COLLECTION_ENABLED` entirely — recurring payments are an always-on surface, collection is now ungated on both cron paths (`BATCH_SIZE` / `RETRY_AFTER_MINUTES` tunables kept)
- private-channel reconcilers run concurrently under their own monitors so a failing leg never skips the other; no Redis slot needed — their crontabs run at or above the job's five-minute schedule
- move the job sequence rationale into a numbered docstring on `runCronJob`; README reconciliation-cron section updated
- companion sdp-infra PR wires `PRIVATE_CHANNELS_ENABLED` into the dev service + job envs
- note: these two commits also ride PR #1402 for now; that branch can rebase them away once this merges

**before** — one sequential chain, each fatal tick a gate for everything behind it:

1. pending transfers + approved-op replay + sponsorship reconcile *(fatal)*
2. earn vault movements *(fatal)*
3. workflow executions *(fatal)*
4. secret retirements *(non-fatal)*
5. earn metrics refresh *(non-fatal)* → catalogue sync *(fatal)*

Tick 1 fails → 2, 3, 4, 5 never run. Tick 2 fails → 3, 4, 5 never run. And since the job is the *only* tick managed deployments get, a persistent failure (e.g. a bad RPC config in tick 1) meant vault intents stopped draining, workflow executions sat `pending`, and Earn went stale — on every 5-minute execution, indefinitely, until the first tick was fixed.

**after** — every tick runs on every execution:

1. pending transfers + approved-op replay + sponsorship reconcile
2. recurring-payment collection *(new)*
3. private-channel deposit ∥ withdrawal reconcilers *(new, gated on `PRIVATE_CHANNELS_ENABLED`)*
4. earn vault movements
5. workflow executions *(gated on asset profiles — always on for managed)*
6. secret retirements *(non-fatal)*
7. earn metrics refresh *(non-fatal)* → catalogue sync *(both gated on the Earn flags)*

Tick 2 fails → 3–7 still run; its failure is collected and logged with its Sentry monitor slug (identifiable even with Sentry off), and the job still fails loudly at the end — a single failure rethrown unchanged, several as an `AggregateError` carrying every cause. The next execution retries everything.


multipletwigs:
Both P2s addressed in 2f301823e:

- **Monitor cadence**: every monitored tick in `job.ts` now declares the job's own `*/5 * * * *` schedule (`MANAGED_JOB_CRON`) instead of the reconciler's in-process crontab — this job is the monitors' only reporter on managed deployments, so the per-minute declarations (private channels AND the pre-existing pending-transfers one) would have alerted on missed check-ins between healthy executions. The per-tick cron parameter is gone entirely.
- **Nested failure causes**: `describeCronFailure` now expands nested `AggregateError`s recursively, so a multi-leg tick's leaf causes (transfers + sponsorship) survive into the structured exit log; covered by a nested-aggregate test.

multipletwigs:
@greptileai review the latest commits please

G1de0n (approved):

PR #1390, opened 2026-08-18
ci: deploy sdp-web to prod via Vercel git integration on release merges

multipletwigs:
- replace the token-based `vercel pull/build/deploy --prebuilt` prod deploy with Vercel's own git integration, gated so main only builds on release merges
- `apps/sdp-web/vercel.json` ignoreCommand now builds a main commit only when its GitHub author is `sdp-release-bot[bot]` (the release-please app); all other main pushes stay skipped, PR previews unchanged
- delete the `deploy-web-production` job from release-please.yml and the manual `deploy-sdp-web-vercel-prod.yml` workflow — both dead without a working team token
- why the author gate is sound: `VERCEL_GIT_COMMIT_AUTHOR_LOGIN` comes from GitHub's resolved account, not forgeable git metadata; main is squash-merge-only with required signatures, so a squash commit is only bot-authored when the PR itself was opened by the release app — release-please has authored 36/36 release commits and nothing else
- why this needs no Vercel credentials: git-triggered builds run on Vercel's infra and resolve the project's production env vars server-side, so no `VERCEL_TOKEN`, org/project IDs, or CI-side env pulling
- hotfix redeploys: Vercel dashboard Redeploy button (the dispatchable workflow this deletes never worked)
- verify on the first release after merge: the deploy fires when release-please's PR merges; if Vercel skips it, the login string in ignoreCommand is the one assumption to adjust

- rewrite `scripts/release-production-workflows.test.mjs`: drop the assertions on the deleted job/workflow, keep the publish-identity + API deploy contracts, and add a test pinning the ignoreCommand release-bot gate

PR #1406, opened 2026-08-19
feat: HOO-997 enforce PII scrubbing across Sentry, logs, and audit metadata

eldarik:
Closes [HOO-997](https://linear.app/solana-fndn/issue/HOO-997/enforce-pii-scrubbing-across-sentry-and-logs). Maps SDP-008 / SDP-019.

## Why

We had credential redaction, applied by hand at ~15 call sites, that knew nothing about PII. Meanwhile the platform stores regulated counterparty data — identity (`firstName`, `lastName`, `dateOfBirth`, `phone`, postal address), contact email, bank instruments (`accountNumber`, `routingNumber`, `iban`) — encrypted at rest behind `services/pii-cipher`. Anything that reached a log line, a Sentry event, or an audit row had left that envelope, and no sink enforced anything: pino had no scrubbing hook, and not one of the four `Sentry.init` sites had a `beforeSend`. `sendDefaultPii: false` only stops the SDK collecting PII on its own — everything the app attaches still shipped.

## What

One denylist, in a new zero-dependency `packages/sdp-redaction` (`sdp-web` can't depend on `@sdp/custody`, which is why those surfaces bypassed redaction entirely). Three passes, because the sinks answer to different readers:

| Pass | Applied to |
| --- | --- |
| Credential-only (`redactCredentialSecrets`) | Client-facing error bodies |
| Credentials + PII (`scrubTelemetry`) | Logs, Sentry, traces, metrics |
| Credentials + PII, emails masked (`scrubAuditMetadata`) | Audit ledger `metadata` |

Enforced at the sinks, not asked of call sites:

| Sink | Enforcement point |
| --- | --- |
| API logs | `hooks.logMethod` in `runtime/logger.ts` — every argument, including the message |
| API Sentry: errors, transactions, spans, logs, metrics, breadcrumbs | `sentryScrubbingHooks` spread into `getSentryOptions` |
| API Sentry capture payload | `scrubError` in `captureUnexpectedError` |
| Audit ledger | `scrubAuditMetadata` in `AuditService.persist`, **before the row is hashed** |
| Dashboard Sentry (browser, server, edge) | `sentryScrubbingHooks` in all three inits |
| Ramp provider messages | `extractProviderErrorMessage` — the one point every provider's echoed-back counterparty field passes into both a log and a response |
| Fireblocks request tracing | `scrubTelemetry` (it writes via `console`, bypassing pino) |

A scrubber failure inside a Sentry hook **drops** the payload rather than passing it through: an unscrubbed event is an incident, a missing event is a gap in a dashboard.

`packages/sdp-custody/src/redaction.ts` moves rather than gaining a re-export shim — a security policy should have one import path, not two plausible ones.

## Please look at these three decisions

1. **The dashboard's Sentry user record is now the Clerk user id only.** It previously sent email and full name on every event, and the feedback widget reused them via `useSentryUser`. Keeping raw emails in Sentry's issue index couldn't be squared with "covers every sink". **Cost: a feedback submission no longer arrives with a repliable address** — the user id still resolves to a person in Clerk. This reverses a decision recorded earlier in the ticket's plan, so it's the one most worth a second opinion.
2. **Audit metadata masks emails (`j***@example.com`) instead of dropping them.** The invitation events in `routes/members/handlers.ts` deliberately store the invitee; `[REDACTED]` there would make the row useless to a reviewer. Masking runs before the ledger hash, so what the chain commits to is what a reviewer reads back. The full address stays in its owning table.
3. **Solana addresses, resource ids, `countryCode`/`subdivisionCode`, `details`, and `userAgent` are explicitly never redacted** (`NEVER_REDACTED_KEYS`). This is why the policy is an explicit key denylist and not an entropy or shape heuristic — every base58 public key would trip a "this looks secret" detector, and those addresses are the primary handle for tracing a payment. The trade I did accept: `displayName` is redacted in telemetry, which takes token and wallet labels with it.

## Two gaps found while building, both fixed

- **`x-api-key` was never redacted.** Key normalization strips separators to `xapikey`, and only an exact `apikey` rule existed — while that header form is exactly what arrives as a Sentry `request.headers` entry. `apikey` is now a suffix rule.
- **The object walker had no depth bound**, despite now sitting on an attacker-reachable path (a webhook body passes through it before anything else reads the payload). Bounded at 16 levels → `[Truncated]`.

## One gap found by the self-review, fixed

Because the `address` container is deliberately excluded, **each provider's street-line key has to be named individually**. `line1`/`line2` (SDP) and `addressLine1`/`addressLine2` (BVNK) were covered; **Mural's `address1` inside `physicalAddress` was not**, so a payout-recipient object reaching a sink would have kept the street. Latent — no current call site logs that object — but exactly the class of gap this ticket exists to close. Added `address1`, `address2`, `physicaladdress`, `residentialaddress`, `homeaddress`, with tests both ways: provider spellings match, and `beneficiaryAddress` / `cryptoAddress` / `destinationAddress` still do **not** (BVNK's `beneficiaryAddress` carries `destinationWalletAddress` — a crypto address — which is why no blanket `*Address` rule can exist).

## Deliberately unchanged

- **Client-facing error bodies keep credential-only redaction.** A 4xx goes back to the tenant that submitted the data in it; stripping its own counterparty fields would remove the only thing making the error actionable.
- **The audit table's `ip_address` / `user_agent` columns.** They are the security record, in dedicated columns, not caller-supplied metadata.
- **`enableScreenshot: true` on the feedback widget** — flagged as an open risk in the policy doc rather than changed here. Screenshots capture the rendered dashboard and `replayIntegration`'s `maskAllText` does **not** apply to them, so a counterparty detail page can reach Sentry as an image no hook can scrub. Product call.

### Test run

Full run on Node 24 (matching CI) after merging `main` at `fb6d2521`, from a clean `node_modules`:

| Gate | Command | Result |
| --- | --- | --- |
| Workspace typecheck | `pnpm typecheck` | **20/20 packages** pass |
| Workspace unit tests | `pnpm test` | **13/13 packages** pass |
| └ `@sdp/api` | | 254 files · **3203 pass**, 10 skipped |
| └ `@sdp/redaction` | | **29 pass** |
| └ `@sdp/payments` | | **36 pass** |
| └ `@sdp/api-integration` (typecheck mode) | | 22 files · **62 pass**, no type errors |
| Dashboard unit tests | `pnpm --filter sdp-web test:unit` | 158 files · **1197 pass** |
| Script tests | `pnpm test:scripts` | **74 pass** — includes the `vercel.json` production-gate contract |
| Module boundaries | `pnpm check:module-boundaries` | pass |
| Pinned dependencies | `pnpm check:pinned-dependencies` | pass |
| Lint | `biome check` on the 44 changed files | clean |
| Dashboard production build | `pnpm --filter sdp-web build` | compiled, 74 static pages generated |

The build was the gate worth watching: `@sdp/redaction` is a new `sdp-web` dependency that exports raw TS with no build step and no `transpilePackages` entry — the same shape as the existing `@sdp/policy` and `@sdp/types` web deps, and it resolves identically. Confirmed the scrubber genuinely reaches the browser rather than being tree-shaken: `REDACTED_EMAIL` is present in the emitted client chunk, so `beforeSend` is live in the tab and not just on the server.

Merge conflicts with `main` were confined to the `@sdp/redaction` additions colliding with `@sdp/kamino` (#1350) and the Vercel deploy move (#1390); all four resolved as unions, no behaviour dropped from either side.

### End-to-end probes

Verified end-to-end against a running local API and Postgres (isolated database, dropped afterwards):

| Probe | Result |
| --- | --- |
| `POST /v1/counterparties`, invalid identity | 400 naming `identity.dateOfBirth` and `identity.address.countryCode`, no submitted values echoed — actionable, not over-redacted |
| `POST /v1/counterparties`, valid | 201; `audit_logs.metadata` persisted as exactly `{"entityType":"individual"}`; `GET` still returns the full identity to its own tenant |
| `POST /admin/allowlist` `{type:"email"}` | Ledger row persisted `"value":"j***@example.com"` — masking by *value shape*, since the address arrives under `value`, a key no rule matches. The `allowlist` table still holds the exact address |
| same, `{type:"domain"}` | `"value":"hoodies.team"` untouched — a non-address value stays fully readable |

Sweep of the entire server log for every PII value, the API key, and the admin key: **0 matches.**

The review also cleared, rather than assumed: the new string scrubbing on `extractProviderErrorMessage` can't touch any substring the provider clients sniff for control flow (`"already APPROVED"`, `"blockhash not found"`, `"reference ="`, Google Places); there is one `pino()` instance and zero `.child()` calls, so the hook has no bypass; `Map`/`Set`, symbol keys, non-enumerable and prototype props drop rather than leak.

## Follow-up

Overlaps [HOO-1011](https://linear.app/solana-fndn/issue/HOO-1011)'s finding 2 — `spc-invite-email.ts` no longer logs the invite URL at all (it's a bearer credential, and the response body already returns it). The `invite_token_hash` migration and the `inviteUrl` value pattern remain that ticket's; note that its plan points at `packages/sdp-custody/src/redaction.ts`, which is now `packages/sdp-redaction`.

Also not built, from the original 7-phase plan: the `noConsole`-for-`packages/**` lint backstop. Of the four `console.*` sites it named, only Fireblocks carried a payload, and the interpolated `bvnk.ts` / `mural.ts` webhook logs turned out to hold only provider ids, wallet ids, org ids, and amounts.


multipletwigs:
**Stale premise: the policy doc and denylist header cite a retired mechanism.**

`docs/security/pii-scrubbing-policy.md:3-8` and `packages/sdp-redaction/src/policy.ts:11-12` say counterparty PII is stored "encrypted at rest" behind `services/pii-cipher`. That layer was retired (#1497–#1499) and migration `0071_drop_counterparty_pii.sql` dropped the columns — on main the platform never persists counterparty PII. The DB holds `displayName` plus provider external references (`providerData`); identity and bank fields are collected JIT and forwarded to providers.

That makes the case for this PR stronger, not weaker: under store-nothing, logs, Sentry, and the audit ledger are the only places SDP could still end up persisting PII — a logged provider response or a hashed audit row would silently undo the store-nothing decision, and the audit ledger is permanent by design. That's the premise the doc should state.

To fix:
1. Rewrite the intro of `docs/security/pii-scrubbing-policy.md` around JIT transit — no `pii-cipher`, no "the platform stores regulated personal data".
2. Same fix in the `policy.ts` header comment.
3. The PR body's "Why" repeats the claim, and its sink table says API Sentry hooks are spread into `getSentryOptions` while the committed doc correctly says the API has no Sentry since #1602 — worth aligning.

Mechanism itself looks right; this is framing only.

multipletwigs (approved):

PR #1412, opened 2026-08-19
build(veda): reach the private @vedatech npm scope, and make Veda key…

cutbow7:
…less

`packages/sdp-veda` (landing later in this stack) depends on `@vedatech/svm-sdk`, the only private package in this workspace. Everything here is the plumbing that lets an install, an image build and CI fetch it — and the correction that a build credential is not a provider credential.

**Auth is an env REFERENCE, never a value.** `.npmrc` is committed and not gitignored, so it carries `//registry.npmjs.org/:_authToken=${NPM_TOKEN}` and the token arrives from outside: an env file for local installs, the `NPM_TOKEN` repository secret in CI, and a BuildKit `--mount=type=secret` for images. Deliberately not `ARG`/`ENV` — both persist in layer history, where `docker history` prints them. Verified on a real build: the token appears in neither the image history nor its filesystem.

**Absence is a supported state.** SDP accepts provider contributions from forks (CONTRIBUTING.md), and a fork cannot read repository secrets. Every workspace install now runs through `.github/scripts/pnpm-install.sh`, which deselects `@sdp/veda` when the token is missing — measured against pnpm 10.16.0: an unfiltered install fails `ERR_PNPM_FETCH_404` on the private tarball while the filtered one completes from the lockfile, because pnpm never requests a deselected project's tarballs. The script publishes `CI_HAS_NPM_TOKEN`, and the jobs that genuinely need the SDK — the API typecheck, the API and `@sdp/veda` unit suites, the API bundle, the API image, the compose smoke, the Surfpool Access shard — skip with a notice naming the cause instead of failing red on a credential the contributor cannot hold. `jobs.<id>.if` cannot see a secret, which is why this is a step-level condition on a value an earlier step wrote.

`translation_agent` is deliberately untouched: it installs `--ignore-workspace` against its own lockfile, which holds no private package, so handing it the token would widen exposure for nothing.

Also plumbed into the two GCP deploy workflows and `release-images.yml`, which build and push the API image — without it, the first deploy after `@sdp/veda` lands would fail at install.

**Veda is keyless.** The SDK talks to Veda's vaults on-chain over RPC; there is no Veda API to authenticate against. `VEDA_API_KEY` / `VEDA_SANDBOX_API_KEY` are removed from `env.d.ts`, `turbo.json` and `scripts/secret-keys.mjs`, and the availability definition becomes `publicApiDefinition`, joining Kamino in `KeyPairedEarnProviderId`'s exclusion. The tempting mistake is reading NPM_TOKEN as Veda's credential: it is build-time only, absent from every running deployment, so declaring it would report Veda unconfigured everywhere — and under ADR 0002 a missing credential is what blocks the EXIT path. `secret-keys.mjs` is "every env key the SDP API reads", so a declared secret nothing reads is a standing question for whoever next provisions the service.

The availability tests used `veda` as their key-paired exemplar; they now use `upshift`, plus a new case pinning that a keyless provider reports configured with no credentials set and never blocks `assertEarnProviderConfigured`.


keeganthomp (dismissed):

pnpm-workspace.yaml
```diff
@@ -32,6 +32,13 @@ minimumReleaseAge: 10080
 # pnpm 10.16 matches minimumReleaseAgeExclude by exact dependency alias.
 # Keep @solana packages explicit until this workspace can rely on scoped wildcard exclusions.
 minimumReleaseAgeExclude:
+  # Veda's SVM SDK ships as pre-release alphas cut for this integration, so the
+  # cooldown would block the exact version packages/sdp-veda pins. The cooldown
+  # exists to keep a compromised release from being pulled in unnoticed; that
+  # risk is answered differently here — the version is pinned exactly (never a
+  # range), so a new publish cannot reach this repo without a lockfile change
+  # under review.
```
keeganthomp: **The cooldown this exempts has already elapsed, and the exemption is permanent.**

`0.1.0-alpha.1` published `2026-08-14T05:16Z`, which is **11,050 minutes old against `minimumReleaseAge: 10080`**. It cleared the cooldown ~16 hours ago. Its transitive deps clear it too (`@solana/kit@7.0.0` from June, `@solana-program/token-2022@0.13.0` from July). Nothing here is needed to install the SDK today.

The durable concern runs the other way: `minimumReleaseAgeExclude` is unversioned, so this exempts **every future `@vedatech` publish**, forever. The `alpha.1 -> alpha.2` bump is exactly the PR where a compromised publish reads as "just a version bump."

Since the dependency lands later anyway, this is untestable config for a package nothing references. It belongs in the PR that adds it.

If it does stay: the pin argument is stronger than the comment claims. It is not a convention, it is a CI gate (`saveExact: true` above, plus `scripts/check-pinned-dependency-versions.mjs` as the "Pinned Dependency Policy" job, whose regex admits prerelease tags).

```suggestion
  # Veda's SVM SDK ships as pre-release alphas cut for this integration, so the
  # cooldown can block the exact version packages/sdp-veda pins. The cooldown
  # exists to keep a compromised release from being pulled in unnoticed; that
  # risk is answered by the pin itself, which is a gate rather than a habit:
  # `saveExact: true` above plus scripts/check-pinned-dependency-versions.mjs
  # (the "Pinned Dependency Policy" CI job) reject any non-exact direct
  # dependency, and its regex admits prerelease tags. A new publish therefore
  # cannot reach this repo without a lockfile change under review.
```

pnpm-workspace.yaml
```diff
@@ -32,6 +32,13 @@ minimumReleaseAge: 10080
...
+  - '@vedatech/svm-sdk'
```
keeganthomp, line 42: **Heads-up for when `packages/sdp-veda` lands: this SDK drags in a second `@solana/kit` major.**

`@vedatech/svm-sdk@0.1.0-alpha.1` pins exactly:

| Package | SDK wants | Workspace catalog |
| -- | -- | -- |
| `@solana/kit` | **7.0.0** | **6.8.0** |
| `@solana-program/token-2022` | **0.13.0** | **0.14.0** |
| `@solana-program/token` | 0.15.0 | 0.15.0 |
| `@solana-program/system` | 0.13.0 | 0.13.0 |

So the tree gets kit 6.8.0 and 7.0.0 side by side, plus ~26 transitive `@solana/*@7.0.0` packages. Branded `Address` / `Rpc` types and `SolanaError` classes do not cross copies cleanly, and it is real weight in the API image. The catalog already flags this area as fragile ("Trial upgrade to satisfy upstream Solana keychain v1 peer requirements", `strictPeerDependencies: false`).

Not a defect in this diff. But this is the "get the workspace ready" PR, and it prepares for the cooldown while the version collision is the hazard that will actually bite.

apps/sdp-api/src/services/provider-availability.service.test.ts
```diff
@@ -57,10 +57,10 @@ const providerEnvKeys = [
   "BVNK_HAWK_AUTH_ID",
   "BVNK_HAWK_SECRET_KEY",
   "BVNK_WALLET_ID",
-  "VEDA_API_KEY",
-  "VEDA_SANDBOX_API_KEY",
   "UPSHIFT_API_KEY",
   "UPSHIFT_SANDBOX_API_KEY",
+  "PERENA_API_KEY",
+  "PERENA_SANDBOX_API_KEY",
```
keeganthomp, line 63: Adding `PERENA_*` is right (the new all-false assertion on `perena` needs it), but `ground` has the same shape and is still missing. It is a `keyPairCredentialDefinition` provider with both keys declared in `env.d.ts`, so `writeProviderEnv` never clears it.

Latent rather than broken, since no test asserts on `ground` today. Whoever writes that test gets a flake if the runner env has the key set.

```suggestion
  "PERENA_API_KEY",
  "PERENA_SANDBOX_API_KEY",
  "GROUND_API_KEY",
  "GROUND_SANDBOX_API_KEY",
```

apps/sdp-api/src/services/provider-availability.service.ts
```diff
@@ -325,7 +325,10 @@ const PROVIDER_AVAILABILITY_DEFINITIONS = {
     },
   },
   earn: {
-    veda: keyPairCredentialDefinition("Veda", "VEDA"),
+    // Keyless like Kamino, and for the same reason: Veda's vaults are read and
+    // written entirely on-chain through `@sdp/veda`, so there is no provider API
+    // to authenticate against, and no credential to declare here.
```
keeganthomp: Small thing: this borrows Kamino's justification for a different situation. Kamino's docstring rests on "no sandbox account to mistake for production," which is an account argument. Veda's sandbox/production split is devnet vs mainnet, so it is a cluster argument.

It holds if Veda's vaults sit at the same addresses on both clusters, which matches what I have seen, but the reader cannot tell that from here. One clause naming the cluster story would keep it honest, since `assertEarnProviderConfigured(env, "veda", true)` now always passes and a devnet withdrawal against a mainnet-only vault would fail deep in the SDK instead of at the gate.

Nothing is reachable today regardless: the client is still `StubEarnClient` and `EARN_PROVIDER_SURFACING.veda = false`.

keeganthomp (changes requested):
## The title and body describe the opposite of what this diff does

**Title:** "reach the private @vedatech npm scope." **Body:** "`@vedatech/svm-sdk`, the only private package in this workspace," then ~400 words on `.npmrc` with `${NPM_TOKEN}`, `.github/scripts/pnpm-install.sh`, `CI_HAS_NPM_TOKEN` step gating, BuildKit `--mount=type=secret`, fork contributors, and `release-images.yml`.

None of that is in the diff. 7 files, zero CI/Docker/npmrc changes. The commit message is the accurate one and says the reverse: *"published to the PUBLIC npm registry, no credential, no private-registry plumbing."*

I checked. Unauthenticated `GET https://registry.npmjs.org/@vedatech%2Fsvm-sdk` returns 200, one version `0.1.0-alpha.1`, maintainer `joseph@veda.tech`. It is public.

This is already misleading readers: Greptile's auto-summary on this PR repeats the error ("prepares the workspace to consume its private S..."). A human reviewer either hunts for token handling that does not exist, or approves believing supply-chain plumbing was reviewed. **Replacing the title and body with the commit message is the only blocking ask here.**

## What is solid

- **No dangling references.** `VEDA_API_KEY` / `VEDA_SANDBOX_API_KEY` are gone from every file in the repo, docs and compose included.
- **Tests pass**, 27/27 across both touched files locally. All 30 CI checks green.
- **Typecheck clean** for this change. The `tsc` errors I hit are pre-existing `helius-rings` / `sdp-kamino` module-resolution noise, nothing in provider-availability.
- **Coverage is net up.** Moving the key-paired exemplar from `veda` to `upshift` keeps `keyPairCredentialDefinition` exercised, and the new keyless case pins behavior that previously had none.
- **The drift guards do their job.** `EARN_CREDENTIAL_ENV_KEYS` derives from the definitions, so dropping Veda's keys automatically relaxes the turbo.json and secret-keys.mjs assertions, while the inverse guard forces the `KEYLESS_EARN_PROVIDERS` edit to be a deliberate line in the diff. This change flows through it correctly.

## Verdict

The code is fine and I would approve it on the merits. Requesting changes purely on the description, plus a suggestion inline that the `pnpm-workspace.yaml` entry moves to the PR that actually adds the dependency, where an install can prove it is needed.

keeganthomp (approved):

keeganthomp (dismissed):

keeganthomp (approved):

PR #1413, opened 2026-08-19
feat(types): add the Veda deployment registry and narrow Veda's decla…

cutbow7:
…red support

Two changes that both answer "what does SDP actually know about Veda?".

**A per-cluster deployment registry, deliberately empty.** Veda's SDK takes every program address at runtime and ships no defaults, so the addresses are configuration Veda supplies rather than constants anyone can derive. SDP has two candidate sets and they disagree: the `address` baked into each Anchor IDL inside `@vedatech/svm-sdk`, and three different addresses in Veda's integration document. MEASURED 2026-08-19 against the public devnet and mainnet RPCs with `getAccountInfo` — none of the three IDL addresses exists on either cluster. They are the source repository's `declare_id!` defaults, not a deployment.

So `VEDA_DEPLOYMENTS` states `null` for both clusters and every consumer fails closed. That is not a placeholder: a wrong address here fails SILENTLY in the worst direction — a program that does not exist reads as an empty shelf, and one that exists but is not Veda's would accept a `deposit` instruction and do something else with the money. Filling an entry in is a pure data change once Veda confirms addresses per cluster; the catalogue read, the builder and their tests already work against a `VedaDeployment` and are exercised with fixtures.

It lives in `@sdp/types` for the reason `kamino-programs.ts` does: `@sdp/earn` reads vault state for the hourly catalogue and `@sdp/veda` builds instructions against the same programs, and an edge between those two would be a workspace cycle that also drags a chain SDK into the catalogue cron.

**`declaredSupport` narrowed to what Veda establishes.** The scaffold claimed `["defi","rwa"]` and USDC/USDG/USDT. `rwa` is the one classification an integrator filters on to find real-world backing, so asserting it needs something Veda publishes on-chain — a Veda vault reaches strategies through pre-approved CPI digests, and what a vault might hold is not what its state establishes. Same rule that keeps every Kamino snapshot `defi`. The asset list becomes USDC alone, sourced from `VEDA_DEPOSIT_TOKEN_SYMBOLS` in `@sdp/types` so the asset the catalogue admits and the asset the builder will spend cannot drift; USDT additionally has no devnet mint, so a devnet snapshot naming it could never pass `isStrategyWithinDeclaredSupport` and would warn every hour forever.


keeganthomp:
reptile right I believe

packages/sdp-types/src/veda-programs.ts
```diff
@@ -0,0 +1,152 @@
...
+}
+
+/**
+ * Deposit assets SDP is willing to front for Veda, as token SYMBOLS.
+ *
+ * Declared cluster-agnostically and bridged to mints through the well-known
+ * token catalogue, matching `EarnDeclaredStrategySupport`. USDC only: it is the
+ * asset Veda's integration material describes, and it is the only one of the
+ * three Earn deposit symbols that has a devnet mint, so a wider declaration
+ * would be a claim SDP could not exercise in sandbox anyway.
```
keeganthomp, line 99: USDG has a devnet mint (`4F6PM96J…`, in `well-known-tokens.ts` in this same package), so USDC isn't the only one of the three with devnet coverage. `client.ts:26` states it correctly (USDT is the one with no devnet mint); this line generalizes it to USDG too. Worth fixing precisely because this file's rationale is what the next person reads before widening the declaration.

```suggestion
 * token catalogue, matching `EarnDeclaredStrategySupport`. USDC only: it is the
 * asset Veda's integration material describes. The sandbox argument only rules
 * out USDT, which has no devnet mint at all; USDG has one, and what keeps it
 * out is simply that no Veda material supports it.
```

packages/sdp-earn/src/providers/veda/client.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      isStrategyWithinDeclaredSupport(
+        client.declaredSupport,
+        snapshot({ depositMints: [USDC_DEVNET, SOL_MINT] })
+      ),
+      false
+    );
+  });
+
+  it("still throws NOT_IMPLEMENTED for listStrategies until the catalogue read lands", async () => {
+    await assert.rejects(client.listStrategies({ env: {}, environment: "sandbox" }), /veda/);
```
keeganthomp: nit: `/veda/` matches any error whose message names the provider, so this also passes on `PROVIDER_NOT_CONFIGURED`, which is exactly what later PRs in this stack will start throwing on this path. At that point the test keeps passing while asserting nothing it claims to. `index.test.ts` already asserts the code directly:

```suggestion
    await assert.rejects(client.listStrategies({ env: {}, environment: "sandbox" }), {
      code: "NOT_IMPLEMENTED",
    });
```

keeganthomp (dismissed):
Solid. Empty-by-design with per-cluster nulls is the right answer given none of the three IDL addresses resolves on either chain, and the doc comment carries the measurement rather than a shrug. Two nits inline, neither blocking: `veda-programs.ts:97` overclaims the sandbox rationale (USDG does have a devnet mint), and the `listStrategies` test matches on a `/veda/` message rather than the `NOT_IMPLEMENTED` code.

On greptile's fixture point: worth taking for readability, but `isStrategyWithinDeclaredSupport` only reads `sourceKind` + `depositMints`, so the devnet-`hostCluster`-with-mainnet-mint snapshot isn't a coverage gap the way the comment frames it.

keeganthomp (dismissed):

keeganthomp (approved):

PR #1422, opened 2026-08-19
feat(earn): read Veda's vault shelf on chain, SDK-free

cutbow7:
`listStrategies` for Veda, plus the account decoding it needs. `@sdp/earn` runs the hourly catalogue cron in both environments and its only dependency is `@sdp/types`, so this reads bytes over JSON-RPC rather than loading `@vedatech/svm-sdk` — the same rule that keeps klend-sdk out of the Kamino catalogue path.

**The shelf is an ALLOWLIST, never a census.** Veda deploys a vault per customer under one program, so enumerating the program's accounts would put other integrators' vaults on SDP's shelf. Addresses come from `@sdp/types/veda-programs`, which carries only what Veda confirmed — which is also what makes `sourceKind: "defi"` defensible here, since every row traces to an address Veda named rather than to anything a stranger could create.

**The layout is a table, not magic offsets.** Kamino's offsets had to be located by matching live values against an independent API; Veda publishes Anchor IDLs, so every offset is derived from the published field order and the account's exact 512-byte length falls out of it. Stating the order explicitly is what makes it checkable — `@sdp/veda`, which holds the IDL, will recompute the table from it and fail if Veda's ABI moves.

**The read is all-or-nothing.** The sync DELETES rows a provider no longer lists, so a partial read does not degrade — it delists whatever went unread. An unreachable RPC, a genesis mismatch, a configured vault whose account is missing, an account that does not decode, and a JSON-RPC error inside a 200 body all throw. That is stricter than the Kamino devnet path's per-row tolerance, and deliberately so: Kamino reads a permissionless census where an odd account is expected, while every address here is one SDP was given.

**The cluster is measured, not derived.** Veda's integration material implies devnet and mainnet may share program addresses. If they do, the genesis-hash proof is the ONLY thing between reading one chain and reporting the other, so `hostCluster` is stamped from the chain actually read and the proof happens before a single account is fetched.

Mapping decisions worth naming:

- **No `currentApy`.** One reading of an exchange rate is not a rate of return. The dashboard renders a missing rate as "—", which is honest; `EarnLiveMetricsProvider` is therefore not implemented — it would return nothing on every five-minute pass forever.
- **No `curator`.** Nothing Veda publishes on-chain attributes one, and an attribution is SDP vouching rather than quoting.
- **`liquidityTerm` needs BOTH constraints clear.** `instant` only when `withdraw_authority` is unset (the comparison Veda's own SDK makes) AND `lock_duration_seconds` is zero. A lock reports `delayed` with a day count rounded UP, because a lock ending part-way through a day is not over until that day is.
- **The name is composed**, since `BoringVault` carries no name field — from the base asset's symbol and the vault id, both facts the account establishes, falling back to the id alone for an unknown asset.

The base58/base64 helpers, the JSON-RPC transport and the genesis proof move to `src/solana-rpc.ts` and are now shared with the Kamino devnet reader, which had private copies of the first two.


packages/sdp-earn/src/providers/veda/client.ts
```diff
@@ -35,4 +134,92 @@ export class VedaEarnClient extends StubEarnClient {
...
+    deployment: VedaDeployment
+  ): Promise<ProviderStrategySnapshot[]> {
+    const entries = await readVedaVaults(rpcUrl, cluster, deployment);
+
+    const snapshots: ProviderStrategySnapshot[] = [];
+    for (const entry of entries) {
+      const mints = depositMints(entry, cluster);
+      // A vault with no deposit asset SDP fronts is not an error and not a row:
+      // it is a vault this deployment has nothing to offer for.
+      if (mints.length === 0) continue;
```
keeganthomp, line 194: **A paused vault ships as `status: "active"`.** `decodeBoringVault` reads `tellerPaused` and `accountingPaused`, and nothing consumes them: `vault-state.ts` is the only hit for either across `packages/` and `apps/`. Meanwhile `syncEarnCatalogue` submits `status: "active"` for anything a provider lists, on the stated grounds that "being listed is what makes a strategy depositable."

So a vault with a halted teller gets a depositable row that fails at build time. That is the same failure the cluster-aware mint screen exists to prevent, and the `mints.length === 0` `continue` right here reads the same way: a vault this deployment has nothing to offer for.

`riskMetadata` is the alternative (open bag, already carries the fee bps), but a flag a partner has to notice is weaker than not listing the row.

```suggestion
      const mints = depositMints(entry, cluster);
      // A vault with no deposit asset SDP fronts is not an error and not a row:
      // it is a vault this deployment has nothing to offer for.
      if (mints.length === 0) continue;
      // Same for a halted teller. The sync marks everything it lists `active`,
      // so a paused vault would be a depositable row that fails at build time.
      if (entry.vault.tellerPaused) continue;
```

keeganthomp (commented):
Read the whole stack entry against `feat/veda-01-deployment-registry`. Verified locally on 593b7e9f: `typecheck` clean, `biome check` clean, 188/188 tests pass.

The shape is right. Genesis-before-read with a test asserting the call ORDER (not just the outcome), the all-or-nothing throw set, the cluster-aware mint screen, and deriving offsets from a declared field table instead of magic numbers are all the right calls. The mainnet-USDC-in-a-devnet-vault case is a sharp catch. Fixture addresses encoded from bytes with a round-trip test is a nice touch.

Five things below, none of which break anything today because `VEDA_DEPLOYMENTS` is all-null. But this PR's own bar is that "the mapping is the half that must already be right when they do", and by that bar four of them are gaps that ship wrong the moment the registry PR lands.

## Findings

| # | Where | What |
| -- | -- | -- |
| 1 | `client.ts:195` | `tellerPaused` is decoded and dropped; the sync marks every listed row `active`, so a halted vault becomes a depositable row |
| 2 | `client.ts:98` | `redemptionDelayDays` quotes the lock alone when a named withdraw authority also gates the exit |
| 3 | `solana-rpc.ts:126` | The doc inverts the sync's empty-shelf behaviour, and `client.ts` in this same PR states it correctly |
| 4 | `vault-state.ts:32` | The IDL cross-check is described in present tense but does not exist, and no `exports` entry would let it import the layouts |
| 5 | `vault-state.ts:321` | No `owner` check on the vault accounts |

## Also

`packages/sdp-earn/CLAUDE.md` still frames the on-chain read as Kamino's `providers/kamino/devnet.ts`. The genesis proof and JSON-RPC transport now live in a shared module and there is a second on-chain provider, with no Veda section anywhere in that file. Worth a paragraph in this PR rather than a later cleanup.

Advisory, not blocking. #1 and #2 are the two I would want landed before the deployment registry gets populated.

packages/sdp-earn/src/providers/veda/client.ts
```diff
@@ -1,12 +1,111 @@
...
+ * lying in the direction that costs a customer money, so both fail to `delayed`.
+ */
+export function vedaLiquidity(entry: VedaVault): {
+  liquidityTerm: "instant" | "delayed";
+  redemptionDelayDays?: number;
+} {
+  const lockSeconds = entry.vault.lockDurationSeconds;
+  const permissionless = entry.vault.withdrawAuthority === VEDA_UNSET_AUTHORITY;
+  if (permissionless && lockSeconds <= 0n) return { liquidityTerm: "instant" };
+  if (lockSeconds <= 0n) return { liquidityTerm: "delayed" };
```
keeganthomp, line 97: **`redemptionDelayDays` understates an authority-gated exit.** With a named `withdraw_authority` AND a non-zero lock, the day count below comes from the lock alone, so the catalogue publishes "2 days" for a vault whose exit also waits on someone signing, with no bound at all. The field reaches partners through `GET /v1/earn/strategies`.

That is the same direction the doc above rules out for `liquidityTerm`, and it is the rule Ground's mapping states outright: "the catalogue never promises faster liquidity than the provider does." Ground only emits a day count because Ground publishes `typicalMaxUnits`; nothing bounds the authority path here, so omitting it is the honest answer.

Second, smaller thing on the same lines: `lockDurationSeconds` is signed (`readIntLe`), and `<= 0n` sends a negative straight to `instant`. Everywhere else the decoder treats an impossible value as a wrong decode rather than a real reading (decimals > 9, zero mints), so this one fails open where the rest fail closed.

No test covers named-authority-and-lock together, which is why neither shows up.

```suggestion
  // A negative lock is a wrong decode, not an unlocked vault, so it must never
  // reach the instant branch.
  if (lockSeconds < 0n) return { liquidityTerm: "delayed" };
  if (permissionless && lockSeconds === 0n) return { liquidityTerm: "instant" };
  // A named authority gates the exit on someone signing, with no bound, so the
  // lock's day count is NOT the redemption delay: quoting it would promise an
  // exit date the vault does not guarantee.
  if (!permissionless || lockSeconds === 0n) return { liquidityTerm: "delayed" };
```

packages/sdp-earn/src/solana-rpc.ts
```diff
@@ -0,0 +1,193 @@
...
+export type RpcAccount = { data?: [string, string] } | null;
+
+/**
+ * One JSON-RPC call, with the failure mode that matters here handled.
+ *
+ * JSON-RPC reports failure INSIDE a 200 body, so the HTTP layer above cannot
+ * see it. Left unchecked, an errored read looks like an empty shelf — and the
+ * catalogue sync DELETES rows a provider no longer lists, so "empty" is the one
+ * shape that quietly delists a provider's whole catalogue. Both branches below
+ * throw for that reason.
```
keeganthomp, line 129: **This has the sync's empty-shelf behaviour backwards, and `listStrategies` in this same PR has it right.** `deleteUnlistedFromCatalogue` skips the delist pass when the keep set is empty ("never tear down a whole shelf off one empty response"), and its own comment notes the repository refuses it too. Empty is the one shape that CANNOT delist a provider; it freezes the catalogue instead.

The throw is still correct, just for the other reason: a frozen shelf that reads as a fresh one. A partial read is the shape that delists, which is what `vault-state.ts` says and what makes the all-or-nothing rule load-bearing.

Worth correcting here specifically because this is the shared file the next on-chain provider will read. (The Kamino original carried the same error, so this is inherited rather than introduced.)

```suggestion
 * JSON-RPC reports failure INSIDE a 200 body, so the HTTP layer above cannot
 * see it. Left unchecked, an errored read looks like an empty shelf, and an
 * empty shelf makes the sync SKIP its delist pass (it refuses to tear down a
 * whole catalogue off one empty response). The row set then freezes on
 * whatever it last held while reading as a fresh pass. Both branches below
 * throw for that reason.
```

packages/sdp-earn/src/providers/veda/vault-state.ts
```diff
@@ -0,0 +1,390 @@
...
+ * `@sdp/types`, because it runs inside the HOURLY catalogue cron in both
+ * environments. Same rule that keeps klend-sdk out of the Kamino catalogue path.
+ *
+ * ── Why the layout is a TABLE, not a list of magic offsets ──────────────────
+ * Unlike Kamino — whose offsets had to be located by matching live values
+ * against an independent API — Veda publishes Anchor IDLs, and
+ * `@vedatech/svm-sdk` ships them with a `programs.lock.json` recording the
+ * commit and a SHA-256 of each. So every offset here is DERIVED from the
+ * published field order, and stating that order explicitly is what makes it
+ * checkable: `@sdp/veda` (which has the IDL) owns a test that recomputes this
```
keeganthomp, line 32: **The cross-check this argument rests on does not exist yet, and could not import the layouts if it did.** `packages/sdp-veda` is not in the tree, and `@sdp/earn`'s `exports` map has `./providers/veda/client` but no `./providers/veda/vault-state`, so the subpath is unreachable from another package. Present tense ("owns a test that recomputes this table") reads as a guard already in place.

What actually pins the table today is `assert.equal(VEDA_BORING_VAULT_SIZE, 512)`, and that is the SUM: **any permutation of the field list preserves it.** The fixture builder writes through the same `offsets` object, so the decoder tests round-trip under a permutation too. Swap `teller.platformFeeBps` and `teller.performanceFeeBps`, or any two of the pubkeys, and the whole suite still passes on a wrong decode.

So the size check catches a wrong WIDTH but not a wrong ORDER, and order is exactly the mistake that "reads a plausible wrong mint". Adding the `exports` entry now is three lines and is what makes the follow-up test possible; the tense here is worth softening until it lands.

packages/sdp-earn/src/providers/veda/vault-state.ts
```diff
@@ -0,0 +1,390 @@
...
+
+  const vaults: VedaVaultAccount[] = [];
+  for (const [index, address] of addresses.entries()) {
+    const data = accountData(accounts[index] ?? null);
+    if (data === null) {
+      throw internalError(
+        `Veda vault ${address} does not exist on ${cluster}. Check the configured deployment addresses in @sdp/types/veda-programs.`
+      );
+    }
+    const decoded = decodeBoringVault(address, data);
```
keeganthomp, line 321: **Nothing checks the account's owner.** `getMultipleAccounts` returns `owner` on every entry and `deployment.vaultProgramAddress` is in scope, but `RpcAccount` does not model the field, so "this address is Veda's vault" rests on 8 discriminator bytes plus a 512-byte length.

Worth closing because `veda-programs.ts` names this exact hazard one file over: SDP has the candidate addresses only as truncated prefixes, and "one that exists but is not Veda's would take a `deposit` instruction and do something else with the money." A typo'd address that happens to decode is far-fetched; an address from the wrong program in a hand-transcribed table is the realistic version, and the owner comparison is what turns the allowlist into a proof instead of an inference.

Cheap: `owner?: string` on `RpcAccount`, compared here alongside the decode.

keeganthomp (dismissed):

keeganthomp (approved):

keeganthomp (approved):

PR #1423, opened 2026-08-19
feat(veda): add @sdp/veda, the SDK wrapper that builds Veda deposits

cutbow7:
The execution half of the integration: unsigned deposit plans and live position reads over the private `@vedatech/svm-sdk`. It never signs, never submits, never touches a database, and holds no runtime credential — Veda's vaults are reached entirely on chain. Mirrors `@sdp/kamino` in shape; the differences are where Veda differs.

**It is a kit-version firewall.** The SDK is built against `@solana/kit` 7 while this repo pins 6.8, so `src/sdk.ts` is the only module that may import it and everything crossing the package surface is kit 6.8, `@sdp/types`, or a decimal string. The tree now carries four kit majors and none of them meet.

**It is also a PRIVATE-REGISTRY firewall, and that half is new.** `@vedatech/svm-sdk` is the only package here needing a credential to install, and `.github/scripts/pnpm-install.sh` survives a missing `NPM_TOKEN` by deselecting exactly one workspace project. That works only while `@sdp/veda` is the SDK's sole owner, so `scripts/check-veda-dependency-boundary.mjs` enforces it — a second owner would turn every forked pull request red on a dependency it never touched. Measured against the real lockfile: without the token an unfiltered install fails `ERR_PNPM_FETCH_404` on the private tarball and the filtered one completes. The shared consumer-scanning helpers move to `scripts/lib/dependency-boundary.mjs`; the Kamino check now uses them.

**Money in requires a way out.** Every deposit build first runs `validateDeployment()` and `validateCompatibility({ requireQueue: true })`. Demanding the withdrawal QUEUE on the deposit path is ADR 0002's exit-safety rule: SDP will not open a position in a vault whose exit infrastructure is not configured and wired to it. It gates only the way in — reads and any future exit never call it — so it can never trap funds. Verdicts cache for ten minutes per (cluster, endpoint, vault); failures are never cached.

**`buildVaultWithdrawal` is deliberately absent**, so `supportsVaultWithdraw` answers false. Veda has two independent exits and SDP has a route for neither; withholding the capability says the route does not exist, never that the customer may not leave — the shares are in their own custody wallet.

**Slippage protection is never invented.** `minSharesOut` is REQUIRED here, unlike Kamino's optional floor: the SDK refuses an implicit tolerance and SDP passes that refusal through rather than defaulting one. `slippageBps` appears nowhere, and a source test asserts it.

**The deposit asset is resolved, and ambiguity is refused.** `EarnVaultDepositInput` carries no mint, so the builder must reach the same asset the catalogue admitted — same predicate (`isVedaDepositMint`, now shared from `@sdp/types` so there is one copy), same source (the vault's own enabled assets). More than one match fails the build rather than picking the first.

**`idl-layout.test.ts` earns its place.** `@sdp/earn` decodes vault state positionally and cannot check itself — it may not depend on the SDK. This package holds the IDLs, so it recomputes that offset table from the published field order and compares the committed IDLs against the SDK's shipped copies AND their recorded SHA-256s. A silent Veda ABI change now fails on the next `pnpm install`, instead of becoming a wrong share mint on a customer's row.

Nothing here has ever run against a live vault: `VEDA_DEPLOYMENTS` is empty, so every chain call fails closed with `DEPLOYMENT_NOT_CONFIGURED`. The env-gated `sdk.smoke.test.ts` takes its deployment from the environment and is the only thing that can prove the integration once Veda confirms addresses.


packages/sdp-veda/src/sdk.ts
```diff
@@ -0,0 +1,479 @@
...
+ * two idempotent associated token accounts, which fits one packet.
+ */
+export async function buildVedaDepositPlan(
+  runtime: VedaRuntime,
+  config: VedaClusterConfig,
+  input: VedaDepositInput
+): Promise<VedaInstructionPlan> {
+  await assertVedaVaultUsable(runtime, config, input.vault);
+
+  const vaultClient = client(runtime, config).vault(input.vault as Kit7);
```
keeganthomp, line 246: This is the second `createVedaClient` for the same runtime+config in two lines: `assertVedaVaultUsable` built one on 239. That's not free, because `createVedaClient`'s constructor eagerly fires `validateUpgradeableProgram` against the vault, hook and queue programs, and every vault op awaits that promise (`#loadState` does `await ctx.ready`), so a deposit pays 2-3 extra `getAccountInfo`s that the 10-minute verdict cache above was specifically meant to stop, and `readVedaPosition` pays them once per vault on a full-shelf read. Threading one client (have `assertVedaVaultUsable` take or return the client, or build it once here and pass it in) removes the duplicate without touching the TTL semantics.

packages/sdp-veda/src/sdk.ts
```diff
@@ -0,0 +1,479 @@
...
+        `${config.cluster} (${candidates.map((asset) => String(asset.mint)).join(", ")}), ` +
+        "and neither a deposit nor a position read can say which one is meant. " +
+        "Carry the mint on the provider contract before widening VEDA_DEPOSIT_TOKEN_SYMBOLS."
+    );
+  }
+
+  const mint = address(String(only.mint));
+  return {
+    mint,
+    decimals: await readMintDecimals(runtime.rpcUrl, mint),
```
keeganthomp, line 223: On the position path this RPC buys nothing. `asset.decimals` reaches only `valuation`, whose zero-share branch does `formatAtomic(0n, asset.decimals)` (line 370) and that is `"0"` for every possible `decimals`; the non-zero branch uses `quote.assetDecimals` instead. So `readVedaPosition` spends one extra `getAccountInfo` per vault, per read, on a number that can never change the output. Resolving decimals lazily, or only in `buildVedaDepositPlan` where it is genuinely load-bearing, keeps the deposit-path guarantee and drops the read cost.

packages/sdp-veda/src/client.ts
```diff
@@ -0,0 +1,304 @@
...
+          assertActive,
+          (reference) => readVedaPosition(runtime, config, { vault: address(reference), owner })
+        );
+
+        const failures = results.flatMap((result, index) =>
+          result.status === "rejected"
+            ? [{ providerReference: references[index], cause: result.reason }]
+            : []
+        );
+        if (failures.length > 0) {
```
keeganthomp, line 373: Question on the all-or-nothing rule for the *implicit shelf* case specifically. `resolveVaultAsset` throws `UNSUPPORTED_VAULT` when a vault configures no SDP-fronted asset (or two), which is a fact about SDP's own config rather than a transient read failure, so one such vault on `vaultStateAddresses` fails every holder's whole Veda portfolio, including their holdings in the other vaults. "A partial portfolio is not a truthful portfolio" clearly covers a vault the caller asked about, but here they asked about nothing. Worth distinguishing structurally-unsupported from unreadable on the shelf path, or is one curated vault in v1 enough that this stays theoretical?

packages/sdp-veda/src/sdk.ts
```diff
@@ -0,0 +1,479 @@
...
+  }
+}
+
+function formatAtomic(value: bigint, decimals: number): string {
+  const negative = value < 0n;
+  const digits = (negative ? -value : value).toString().padStart(decimals + 1, "0");
+  const whole = digits.slice(0, digits.length - decimals);
+  const fraction = decimals === 0 ? "" : digits.slice(digits.length - decimals).replace(/0+$/, "");
+  return `${negative ? "-" : ""}${whole}${fraction === "" ? "" : `.${fraction}`}`;
+}
```
keeganthomp, line 602: nit: this is `formatDecimalAmount` from `@sdp/solana/amount` reimplemented (I checked, they agree on sign, `decimals === 0`, and trailing-zero trimming). `amounts.ts` in this package already calls that module "the repo's fixed-point arithmetic", so a second copy here is the one that can drift. Needs `formatDecimalAmount` added to the imports:
```suggestion
function formatAtomic(value: bigint, decimals: number): string {
  return formatDecimalAmount(value, decimals);
}
```

packages/sdp-veda/src/sdk.smoke.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      minSharesOut: process.env.VEDA_SMOKE_MIN_SHARES_OUT ?? "0.000001",
+    });
+
+    expect(plan.cluster).toBe("devnet");
+    expect(planInstructionCount(plan)).toBeGreaterThan(0);
+    expect(plan.accepted.amount).toBeDefined();
+    expect(plan.accepted.minSharesOut).toBeDefined();
+    // `assertPlanTargetsCluster` already ran inside the builder; this reports
+    // WHAT it admitted, so a surprising program shows up in the failure output.
+    expect(planProgramAddresses(plan).map(String).sort()).toMatchSnapshot();
```
keeganthomp, line 81: nit: a snapshot is the wrong assertion for an env-gated test whose values are per-deployment by design. The first person to run this writes a `.snap` keyed to their `VEDA_SMOKE_*` addresses, and it then fails for anyone whose deployment differs, with CI never running it either way. Asserting the property instead keeps the intent (a surprising program still shows in the failure output): filter `planProgramAddresses(plan)` against `vedaProgramAllowlist(config())` and expect `[]`.

scripts/check-module-boundaries.mjs
```diff
@@ -117,6 +118,20 @@ const MODULE_METADATA = [
     // @sdp/types, which both reach without a cycle.
     allowedDependencies: ["@sdp/earn", "@sdp/solana", "@sdp/types"],
   },
+  {
+    name: "@sdp/veda",
+    directory: "packages/sdp-veda",
+    purpose:
+      "Kit-native Veda SVM vault deposit plans and position reads over the private @vedatech/svm-sdk.",
```
keeganthomp: nit: "private" is stale, and this string is what generates the row in `docs/architecture/module-map.md`, so a reader concludes `pnpm install` needs a registry token. The package's own CLAUDE.md already records that the scope opened during the integration, and the lockfile resolves it from registry.npmjs.org. Same wording in `index.ts:4` and the header of `check-veda-dependency-boundary.mjs`.

keeganthomp (dismissed):
Solid, and the parts that carry risk are the parts that are pinned hardest. I checked the seam against the real `@vedatech/svm-sdk@0.1.0-alpha.1` rather than the mocks: `AccountRole` is 0-3 in both kit majors, `VedaSdkError.context` defaults to `{}` so the `"issue" in cause.context` branch can't throw, `getUserPosition` returns `0n` for an absent share account rather than throwing, and every `buildDeposit`/`validateCompatibility` key you pass matches the SDK's declared shape. I also recomputed `idl-layout.test.ts` by hand: BoringVault is 512 bytes, all 47 declared offsets and the 8 AssetData ones match the IDL, and the three committed IDLs are byte-identical to the SDK's copies with matching SHA-256s. All three boundary scripts pass locally.

Nothing blocking. Six non-blocking notes inline, the two worth your time being the duplicate `createVedaClient` per operation (each construction eagerly re-probes all three programs, so a deposit pays for it twice and a shelf read once per vault) and `readMintDecimals` on the position path, whose result never reaches the output.

keeganthomp (dismissed):

keeganthomp (approved):

PR #1424, opened 2026-08-19
feat(api): dispatch Veda through the vault-direct execution path

cutbow7:
Wires `@sdp/veda` into the API so a Veda catalogue row can take a deposit through the same route Kamino does. The route itself is unchanged — that is the result, not an accident: `POST /v1/earn/vault-deposits` names no provider, and the only edits needed were the ONE sanctioned provider-id branch and its registry overlay.

`resolveEarnExecutionClient` gains a `veda` case; unknown ids still answer `null` rather than throwing, so a strategy row written by a newer deploy degrades to "cannot execute" instead of 500ing a read that touched it. The shared resolver and deadline runner are lifted out of the Kamino branch rather than copied, and `earn-provider-registry.ts` overlays the executing client the same way, `assertNotPortfolioProvider` included — a client claiming both capabilities would let a portfolio route hand a customer a vault account as a deposit address.

Deposit build failures are now mapped by CODE rather than per error class: `INVALID_AMOUNT`, `DEPOSIT_REFUSED` and `COMPLIANCE_APPROVAL_REQUIRED` become a 400 carrying the provider's own sentence, because each names something the request or the vault's current state makes impossible. Everything else keeps bubbling — telling a customer their request was wrong when SDP does not know that would be a guess, and `VAULT_UNREADABLE` has its own test to pin it.

**Two image-context fixes the Kamino precedent did not need.** `apps/sdp-api/Dockerfile.dockerignore` is an allowlist, so `packages/sdp-veda` had to be re-included or the COPY failed on a directory the build could not see. And the deps stage never copied `.npmrc`, so the mounted secret had nothing to expand into — the install reached the registry with no auth header at all. Both were found by building the image, not by reading it. The stage now FAILS on a missing or empty secret rather than warning: after this change the API cannot be built without one, and a 404 several minutes later explains nothing. Verified end to end — the image builds, `server.js` carries the Veda SDK, and the token appears in neither the image filesystem nor `docker history`. `docker compose build` passes with the token supplied through the compose build secret.

The docs image is deliberately untouched: its own context allowlist excludes `packages/sdp-veda` exactly as it excludes `packages/sdp-kamino`, so `--filter @sdp/api...` there resolves against a workspace without them.

Test coverage follows the layer that owns each property. The route suite gains the surfacing seam (Veda is registered and executable but NOT offered, so the real map refuses it 403) plus dispatch and fail-closed cases; replay lives in the service suite where the movement row is real. `execution-registry.test.ts` pins that Veda refuses on its missing deployment BEFORE any RPC work, since an unconfigured deployment is a fact about SDP that should not depend on a node being reachable.


apps/sdp-api/src/services/earn/vault-deposit.service.ts
```diff
@@ -125,6 +126,37 @@ function requireAcceptedPlan(
...
+ * - `INVALID_AMOUNT` — the number is unusable at the mint's scale, or below
+ *   what the vault will mint a share for.
+ * - `DEPOSIT_REFUSED` — the vault will not take this right now: paused, at its
+ *   deposit cap, the asset disabled, the wallet short of balance.
+ * - `COMPLIANCE_APPROVAL_REQUIRED` — the vault gates deposits on an approval
+ *   from the provider's compliance service, which SDP does not implement. A
+ *   definite, explainable refusal rather than an internal fault.
+ *
+ * Matched on the shared `code` shape rather than per provider, so a new
+ * vault-direct provider inherits the mapping by using the same vocabulary.
```
keeganthomp: nit: this isn't matched on code shape, it's matched per error class and then filtered by code. Line 156 gates on `instanceof SdpKaminoError || instanceof SdpVedaError`, so a third vault-direct provider that adopts the exact same vocabulary still 500s until someone adds an arm here. Keeping the `instanceof` is the right call for a money path (duck-typing any `{code: string}` would widen the trust boundary to every Node system error), so I'd just fix the promise:
```suggestion
 * Matched per error class, then by code. A new vault-direct provider does NOT
 * inherit this by reusing the vocabulary; add it to `refusedBuildMessage`.
```

apps/sdp-api/src/services/earn-provider-registry.ts
```diff
@@ -27,23 +31,34 @@ async function resolveKaminoRpcUrl(
...
   const deadline = createVaultDeadline();
   return deadline.run(label, () => operation(() => deadline.assertActive(label)));
-});
+}
+
+const kamino = new KaminoVaultDirectClient(resolveProvenRpcUrl, runVaultOperation);
 assertNotPortfolioProvider(kamino);
 
+const veda = new VedaVaultDirectClient(resolveProvenRpcUrl, runVaultOperation);
+assertVedaNotPortfolioProvider(veda);
```
keeganthomp, line 47: nit: `earn-provider-registry.test.ts` pins the Kamino overlay (`supportsVaultDirect` true, `supportsPortfolioWallets` false) but got no Veda case, so deleting `veda` from the object below would still pass everything. The route suite won't catch it either, since `depositIntoVault` resolves through `resolveVaultDirectClient`, not this registry. Three lines mirroring the Kamino `it` would cover it.

keeganthomp (dismissed):
Dispatch story holds up. I traced the positions handler specifically: an unconfigured Veda throws inside its own hydration job and `mapSettledWithConcurrency` swallows it per provider, so a Veda row can't poison a page of Kamino positions. Two non-blocking nits inline, nothing that should gate merge.

Two things outside the diff worth a look. The body's fourth paragraph describes copying `.npmrc` into the deps stage and failing on a missing build secret, and none of that is in this branch: Docker Build API is green and `@vedatech/svm-sdk@0.1.0-alpha.1` resolves unauthenticated from registry.npmjs.org with the lockfile's integrity hash, so nothing is broken, but the body claims a security-relevant change that isn't there (`packages/sdp-veda/src/index.ts` calls the SDK private too). Separately, [`EARN_PROVIDER_SURFACING`'s comment](https://github.com/solana-foundation/solana-developer-platform/blob/9fa0e7ac662ea84f2adebe5e8b94dc752d5aeb02/packages/sdp-types/src/provider-access.ts#L101-L104) still explains `veda: false` as "never implemented, catalogues nothing", which is precisely what this stack undoes. Worth rewording there, since the next reader of that line will think flipping it is registration cleanup rather than opening money-in dispatch.

The red Lint check is the repo-wide `biome.json` 2.4.16 schema vs CLI 2.5.7 mismatch, identical on main. Not yours.

keeganthomp (dismissed):

keeganthomp (approved):

PR #1425, opened 2026-08-19
docs(veda): design the vault_direct exit path before building it

cutbow7:
No code. SDP can move money into a non-custodial vault and not back out, and the work that fixes it needs three decisions made BEFORE it rather than discovered during it.

**The queued lifecycle does not fit the movement model.** A Veda queue request is settled by a solver Veda operates: its creating transaction confirms in seconds while the request itself stays open for hours, so putting it on `earn_vault_movements` either makes `confirmed` mean two things or leaves the row nonterminal for the request's whole life — and the sweep's work queue never drains. It needs a new table, which is also the cleanest EXPAND under the deploy ordering the playbook describes.

**The route split matters.** Instant redemption needs no schema change at all: `direction IN ('deposit','withdraw')` already exists and the build → sign → record → broadcast → confirm lifecycle describes it exactly. So it lands first and alone, and is a complete exit for any vault whose redemption is permissionless.

**Money out inherits no money-in gate**, restated concretely: no surfacing, no entitlement, no strategy-status check, no `isVaultDirectDepositEnabled`. A paused strategy, an un-surfaced provider and a revoked override must all still let a customer leave.

Also records what was measured rather than assumed: Veda offers two independent exits and its SDK deliberately refuses to choose between them, `closedOrUnknown` is genuinely ambiguous after a request closes (with a concrete proposal for resolving it from the closing transaction's logs rather than guessing), and opening `VAULT_DIRECT_DEPOSIT_ENVIRONMENTS` to production would open Kamino too — which still has no exit path, so that flip is escalated and explicitly not part of this work.

Five open questions are escalated rather than answered, headed by whether the launch vault exposes instant redemption at all — if it does not, the whole sequence reorders.


docs/decisions/0003-veda-vault-withdrawals.md
```diff
@@ -0,0 +1,255 @@
...
+  correctly batched — a multi-reserve exit can exceed Solana's 1232-byte packet
+  and the builder returns one unsized batch (`packages/sdp-kamino/CLAUDE.md`).
+- **Veda** can build a correct single-transaction exit today. What is missing is
+  everything on SDP's side: **there is no `POST /v1/earn/vault-withdrawals`
+  route at all**, and no service behind it.
+
+`VAULT_DIRECT_DEPOSIT_ENVIRONMENTS` is `["sandbox"]` precisely because of this
+gap. Nothing is fund-trapped — the shares sit in the organization's own custody
+wallet and each provider's own surfaces can redeem them — but SDP's product is
+one-directional until this lands.
```
keeganthomp, line 29: This whole framing was true on 08-19 and isn't any more: #1439 (PRO-1702) landed on 08-21 and is already in this PR's own base branch. Kamino's batching problem is solved, `POST /v1/earn/vault-withdrawals` exists with `vault-withdraw.service.ts` behind it, and `packages/sdp-kamino/CLAUDE.md` (cited here for the opposite claim) now reads "implements `buildVaultWithdrawal`, so `supportsVaultWithdraw` answers true".

```suggestion
SDP can move money out of a `vault_direct` vault today, but only for Kamino.
PRO-1702 shipped the whole exit path: `POST /v1/earn/vault-withdrawals` plus its
reads, `vault-withdraw.service.ts`, and a compile-measured builder that rejects
an over-sized transaction rather than batching it (ADR 0002, "One exit is one
signed movement"). `KaminoVaultDirectClient` implements `buildVaultWithdrawal`,
so `supportsVaultWithdraw` answers true for it.

What is missing is Veda's half. `VedaVaultDirectClient` implements
`buildVaultDeposit` and `readVaultPositions` and deliberately not
`buildVaultWithdrawal`, so the existing route answers 501 for Veda
(`packages/sdp-veda/CLAUDE.md`, "The withdraw capability is WITHHELD"). Nothing
is fund-trapped: the shares sit in the organization's own custody wallet and
Veda's own surfaces can redeem them.

`VAULT_DIRECT_DEPOSIT_ENVIRONMENTS` is `["sandbox"]`, but that gate is about
DEPOSITS, not this gap. ADR 0002 is explicit that the exit route ignores it and
"an exit works in production today, where deposits stay closed pending
PRO-1703".
```

docs/decisions/0003-veda-vault-withdrawals.md
```diff
@@ -0,0 +1,255 @@
...
+   exit path. That decision must be made deliberately, by a human, and it is
+   NOT part of this work.
+
+## Decision
+
+### 1. Two capabilities, not one
+
+Instant redemption is the existing `EarnVaultWithdrawProvider` capability:
+`buildVaultWithdrawal` returns an `EarnVaultTransactionPlan`, discovered by
+`supportsVaultWithdraw`. Veda becomes the first provider to answer true.
```
keeganthomp, line 102: Not the first. `packages/sdp-earn/CLAUDE.md` already says `supportsVaultWithdraw` is the capability "which Kamino implements since PRO-1702".

```suggestion
`supportsVaultWithdraw`. Kamino has answered true since PRO-1702, so Veda
becomes the second.
```

docs/decisions/0003-veda-vault-withdrawals.md
```diff
@@ -0,0 +1,255 @@
...
+The instant path needs NO schema change. `earn_vault_movements` already carries
+`direction IN ('deposit','withdraw')` (migration `0059`), and its lifecycle —
+build → simulate → sign → durably record → broadcast → confirm — describes an
+instant redemption exactly as well as it describes a deposit. The reconcile
+cron (`cron/earn-vault-movements.ts`) needs no new states.
+
+So the first implementation PR is: `buildVaultWithdrawal` in `@sdp/veda`, a
+`vault-withdrawal.service.ts` mirroring `vault-deposit.service.ts`, and
+`POST /v1/earn/vault-withdrawals`. That is a complete, shippable exit for any
+vault whose redemption is permissionless.
```
keeganthomp, line 144: This is the part most likely to produce wrong code, so worth getting exact. Three corrections: the canonical ledger is `earn_movements` (0062), not `earn_vault_movements` (0059), which 0062 is explicitly retiring; the direction value is spelled `'withdrawal'` there (0062 notes 0059's `'withdraw'` was never written by any row); and PR 2's scope here rebuilds a service and route that already exist. Also missing the one rule an implementer would otherwise get wrong: the row is share-denominated.

```suggestion
The instant path needs NO schema change, and no new service or route either.
The canonical ledger is now `earn_movements` (migration `0062`), not
`earn_vault_movements` (`0059`), and the direction value is spelled
`'withdrawal'`: `0062` notes that `0059` wrote `'withdraw'` on a column no row
ever carried. Statuses live in `earn_movement_statuses`, keyed by
`(execution_model, status)` with an explicit `is_terminal`. The reconcile cron
(`cron/earn-vault-movements.ts`) already reads `earn_movements` and already
sweeps withdrawals, so it needs no new states.

One rule the instant path inherits, and this document should state outright: a
vault WITHDRAWAL row is denominated in the SHARE MINT, and `amount_requested`
is the exact share quantity decoded from the instruction bytes, never a
build-time token estimate (ADR 0002).

So the first implementation PR is just `buildVaultWithdrawal` in `@sdp/veda`.
The service (`vault-withdraw.service.ts`), the route and its reads already
exist and are provider-neutral, so implementing the method is what flips
`supportsVaultWithdraw` and turns the existing 501 into a working exit.
```

docs/decisions/0003-veda-vault-withdrawals.md
```diff
@@ -0,0 +1,255 @@
...
+
+A queue request is a long-lived object with terminal states that are not
+observable from the transaction that created it. Its creating transaction is
+`confirmed` within seconds; the request itself stays pending for hours or days
+and is settled by an actor SDP does not control.
+
+Putting that on `earn_vault_movements` forces one of two bad outcomes: either
+`confirmed` means two different things depending on `direction`, or the movement
+stays nonterminal for the request's whole life — and the sweep's work queue
+(`idx_earn_vault_movements_unsettled`) never drains.
```
keeganthomp, line 156: The conclusion (separate table) still looks right to me, but this justification is aimed at 0059's flat status column, which 0062 already replaced. `earn_movement_statuses` is keyed `(execution_model, status)` with an explicit `is_terminal`, and `confirmed` is deliberately NOT terminal for `vault_direct`, so "confirmed means two different things" and "the sweep never drains" are both already designed around. The live question is whether a queue request is a new `execution_model` on `earn_movements` or its own table, and the honest argument for the table is object lifetime (one request spans many transactions and is settled by a third party), not status ambiguity. Worth re-arguing on those terms.

docs/decisions/0003-veda-vault-withdrawals.md
```diff
@@ -0,0 +1,255 @@
...
+
+## Sequencing
+
+| PR | Contents | Gate to start |
+|---|---|---|
+| 1 | This document | — |
+| 2 | `buildVaultWithdrawal` in `@sdp/veda`, withdrawal service, `POST /v1/earn/vault-withdrawals`, `GET .../:movementId` | Veda confirms deployment addresses (`VEDA_DEPLOYMENTS` is empty; nothing can be proven until then) and confirms the launch vault's `withdraw_authority` |
+| 3 | `GET /v1/earn/vault-positions/:id/withdrawal-options` | after 2 |
+| 4 | Queued capability, `earn_vault_withdrawal_requests` migration, request/cancel routes, the closed-request indexer | product answer to open question 5 |
+| — | `VAULT_DIRECT_DEPOSIT_ENVIRONMENTS` including `"production"` | **escalated, not part of this work** — the constant is global and flipping it opens Kamino, which still has no exit path |
```
keeganthomp, line 223: Same inversion as the Context section, and it matters more here because it reads as a fund-safety claim. Kamino has an exit path; and per ADR 0002 this constant does not gate either provider's exit at all, only deposits.

```suggestion
| n/a | `VAULT_DIRECT_DEPOSIT_ENVIRONMENTS` including `"production"` | **escalated, not part of this work**: the constant is global, so flipping it opens Kamino AND Veda DEPOSITS at once. It does not gate either provider's exit, which ADR 0002 keeps open in production. |
```

docs/decisions/0003-veda-vault-withdrawals.md
```diff
@@ -0,0 +1,255 @@
...
+5. **Which exits does SDP expose at launch — instant, queued, or both?** This is
+   the product decision PR 4 waits on, and it is also open question 5 of the
+   original integration plan.
+
+## Consequences
+
+- Veda becomes the first provider where `supportsVaultWithdraw` is true, which
+  makes the capability guard load-bearing for the first time: the withdrawal
+  route must narrow on it and refuse Kamino, in the same deployment, on the same
+  contract.
```
keeganthomp, line 254: Backwards today: the route already narrows on `supportsVaultWithdraw` and the provider it refuses is Veda, not Kamino. The real consequence is the opposite one, and it costs test coverage.

```suggestion
- The capability guard stops refusing anyone. It is already load-bearing: the
  route narrows on `supportsVaultWithdraw` today and answers 501 for Veda while
  serving Kamino. Veda implementing the method inverts that, so both
  `vault_direct` providers answer true and no real provider exercises the 501
  branch any more. Keep a deposit-only provider double in the route tests so
  that branch stays pinned.
```

docs/decisions/0003-veda-vault-withdrawals.md
```diff
@@ -0,0 +1,255 @@
...
+writers never touch it at all, so the deploy ordering hazard the playbook
+describes — migrations run before the service rolls, and a rollback restores the
+old image over the new schema — cannot bite.
+
+If any column is added to an EXISTING table as part of this work, it is nullable
+in the release that adds it and `SET NOT NULL` comes in a later one.
+
+### 5. Money out never inherits a money-in gate
+
+ADR 0002's exit-safety invariant, restated because this is the path it exists
```
keeganthomp, line 172: nit: since the exit route shipped, this section documents enforced behavior rather than a decision to make, and the tests already exist: `earn.vault-withdrawals.test.ts` has a `POST /v1/earn/vault-withdrawals - exit safety (ADR 0002)` block. Pointing PR 2 at that file beats pointing it at the custodial `earn-program.test.ts`. One gap in the list below: ADR 0002 does allow exactly one provider-shaped refusal, capability (501), which is worth naming so "NOTHING else" doesn't read as forbidding it.

keeganthomp (changes requested):
The design calls here are sound (two capabilities not one, never auto-select a route, and the open questions are the right ones to escalate), but the document describes a codebase that stopped existing two days after it was written. #1439 / PRO-1702 landed on 08-21 and is already in this PR's base branch, so the Context section, Decision 3, the Sequencing table and the Consequences are all inverted: Kamino has a working exit, `POST /v1/earn/vault-withdrawals` and `vault-withdraw.service.ts` already exist and are provider-neutral, and `supportsVaultWithdraw` currently refuses Veda rather than Kamino.

That matters more than usual for an ADR, since this one's only product is telling PR 2 what to build, and as written it says to build an existing route against a table (`earn_vault_movements`) that migration 0062 is retiring. It also drops the rule an implementer is most likely to get wrong: a withdrawal row is denominated in the SHARE MINT with the exact decoded share quantity.

Blocking: docs/decisions/0003-veda-vault-withdrawals.md:10-23, :96, :129-138, :217, :245-248. Suggestions inline for each, so this should be a refresh against the new base rather than a rethink. Decision 4's conclusion I'd keep, just re-argued (:147).

keeganthomp (dismissed):

keeganthomp (approved):

PR #1426, opened 2026-08-19
chore(veda): add the catalogue inventory and refresh the Earn agent n…

cutbow7:
…otes

The census artifact every catalogued provider carries, plus the doc updates this stack made necessary.

**The inventory answers a different question than Kamino's.** Kamino's registry is permissionless, so its report is about what a TVL floor refused. Veda's shelf is an explicit allowlist, so the useful question is "for each vault SDP was given, what does the chain actually say, and what did that turn into?" — enabled assets and which SDP fronts, pause flags, compliance mode, share lock, withdraw authority, fees. Every one of those decides what a customer is shown and every one is read positionally off an account nobody inspects by hand.

It runs the SAME code the hourly sync runs — `readVedaVaults`, `isVedaDepositMint`, `vedaVaultName`, `vedaLiquidity` — so the report cannot disagree with the catalogue it describes, and it validates the snapshot on WRITE as well as read because the file is committed and its contents come off chain.

The committed report currently says both clusters were not read, because `VEDA_DEPLOYMENTS` is empty. That is the truthful artifact, not a placeholder: the moment Veda confirms addresses, one regeneration turns it into the real census. Every render path was still exercised — against a populated fixture covering a catalogued vault, a dropped one, an unknown base asset, a share lock, and both pause flags — because a report whose tables have never run is not a report.

`packages/sdp-earn/CLAUDE.md`'s capability list had gone stale twice over: it predated the vault-direct and vault-withdraw capabilities entirely, and its "two provider shapes" section described Kamino as the only non-custodial one. Both now say what is actually true, including WHY neither vault-direct provider implements vault-withdraw — Kamino can build an exit but not a batched one, Veda can build a correct one but SDP has no route — because those are different problems with different fixes.

Generated inventory snapshots are excluded from biome, alongside the vendored IDLs. Their shape is owned by the script that writes them, and a formatter that disagrees with `JSON.stringify` turns every regeneration into a lint failure — which it did, on a single-element array.


packages/sdp-earn/CLAUDE.md
```diff
@@ -330,9 +382,13 @@ rates come from the same paged endpoint the catalogue uses.
...
   `resolveEarnProviderClient` — DB provider ids are open strings and MUST be
   resolved through this, never direct-indexed.
-- Optional capabilities so far: portfolio wallets, withdrawal approvals, and
-  live metrics. All three are method-presence guards in capabilities.ts, and a
-  provider may implement any subset — Kamino has only the third.
+- Optional capabilities so far: portfolio wallets, withdrawal approvals, live
+  metrics, vault-direct (deposit + read) and vault-withdraw. All are
+  method-presence guards in capabilities.ts and a provider may implement any
+  subset — Kamino has live metrics and vault-direct, Veda has vault-direct.
+  NOBODY implements vault-withdraw yet, which is a statement about SDP's
```
keeganthomp, line 415: This contradicts line 182 of this same diff, and the code: `KaminoVaultDirectClient implements EarnVaultWithdrawProvider` (`packages/sdp-kamino/src/client.ts:109`), and its test asserts `supportsVaultWithdraw(client)` is true. An agent trusting the capability matrix here would conclude `POST /v1/earn/vault-withdrawals` doesn't exist and build a second one.

```suggestion
  subset — Kamino has live metrics, vault-direct and vault-withdraw; Veda has
  vault-direct only. Veda withholding vault-withdraw is a statement about SDP's
```

keeganthomp (changes requested):
Script and generated artifact both hold up: I re-derived the committed report from the committed snapshot and it matches `renderReport` byte for byte, the biome exclusion genuinely skips `.earn-catalogue` (`Checked 0 files`, path reported as ignored), the RPC env names match `resolveClusterRpcUrl`'s convention, the admission rule matches the client's `depositMints`, and the schema bounds line up with the decoder (shareDecimals capped at 9, signed lockDurationSeconds). Every factual claim in the new Veda section checks out too, including the `idl-layout.test.ts` ABI-drift test it cites. One blocking thing: `packages/sdp-earn/CLAUDE.md:389` says nobody implements vault-withdraw, which contradicts line 182 of the same file and the actual code.

keeganthomp (dismissed):

keeganthomp (approved):

PR #1430, opened 2026-08-20
feat(rpc): reusable verified-confirmation helper for on-chain writes

G1de0n:
## Summary

Extracts the issuance deploy triple-check (signature status, account existence, transaction re-read) into a reusable `verifyTransactionLanded` helper in `@sdp/rpc`, and wires `deploy.ts` to it. Other money paths can adopt the helper in follow-up PRs.

## Changes

- `packages/sdp-rpc/src/verified-confirmation.ts`: new `verifyTransactionLanded(rpc, signature, { expectAccount? })` returning a discriminated result (`ok` / `not_confirmed` / `account_missing` / `not_indexed`); verdicts are returned, transient RPC failures are retried and only rethrown once the schedule is exhausted.
- `packages/sdp-rpc/src/transient.ts`: `withTransientRpcRetry` moved here from `solana.ts` and exported, with an injectable delay schedule for tests.
- `apps/sdp-api/.../issuance/handlers/deploy.ts`: replaces the inline checks with the helper; error responses unchanged. The signature-status and account-existence reads gain transient retries they previously lacked.
- `@sdp/rpc` gets a `node:test` runner; 12 unit tests cover all verification outcomes plus retry classification.

## Testing

- `pnpm --filter @sdp/rpc test`: 12/12 pass; typecheck and lint clean for `@sdp/rpc` and `@sdp/api`.


resourcefulmind (approved):
I read this one as the gate for the two that depend on it. It's a faithful extraction, all three checks in the same order and all three error responses preserved. The
signature-status and account-existence reads gain transient retry they did not have, which is strictly safer.

multipletwigs:
hey andrey, unrelated to this PR (none of this is in the diff) but poked around `solana.ts` while reviewing the stack and found a few spots reinventing stuff `@solana/kit` already ships. figured I'd drop it here since this PR's the one touching this file most:

- `requestAndConfirmAirdrop` (line ~378) hand-rolls a raw `fetch` + JSON-RPC payload for `requestAirdrop`, then casts `payload.result as Signature` on unvalidated JSON. kit has `airdropFactory({ rpc, rpcSubscriptions })` for exactly this — one typed call, push-confirmed via subscriptions instead of polling, and explicitly devnet/testnet-only which fits us. `createRpcSubscriptions(env)` is already in this file for the second arg.
- `confirmTransaction`'s commitment check (line ~354) hand-rolls the processed < confirmed < finalized ordering with an OR-chain. kit exports `commitmentComparator(a, b)` for this — `commitmentComparator(result.confirmationStatus, commitment) >= 0` is the same thing, just not reinvented.
- `sendTransaction` and `simulateTransaction` both do `Buffer.from(...).toString("base64")` for the tx encoding, but `getTransactionNetworkFee` right there in the same file already does it the kit way with `getBase64Decoder().decode(...)`. worth making the other two consistent with it.

G1de0n:
Good catches, all three check out — did them in #1440 rather than folding them in here, since this one's approved and scoped to the extraction, and the base64 change touches `sendTransaction`/`simulateTransaction` which six value-moving handlers use.

A couple of notes from doing it:

- The airdrop one was the only real correctness issue — `payload.result as Signature` on unvalidated JSON. Conveniently nothing in the repo calls that function today, so it was a free fix. One wrinkle: `airdropFactory` is typed for test clusters only, so our general `SolanaRpc` doesn't satisfy it and the client needs narrowing at that call. Still a much better trade than casting untrusted response data.
- On base64: `Buffer` wasn't genuinely available here either — it was hand-declared at the top of the file because the package builds without Node types. Moving both call sites to `getBase64Decoder()` let that declaration go away entirely.
- `commitmentComparator` matched the ordering the OR-chain encoded exactly; added a null guard since `confirmationStatus` can be absent.

Ran the issuance and transfer-batch suites against it (300 passed) since those are the paths that actually submit and simulate.

PR #1431, opened 2026-08-20
feat(payments): cross-check Solana Pay settlement via verifyTransactionLanded

G1de0n:
## Summary

Stacked on #1430. Adopts the `verifyTransactionLanded` helper in the Solana Pay settlement path (`payment-requests.ts`), so inbound payment requests only settle after an independent signature-status check agrees with `@solana/pay`'s own transaction validation.

## Changes

- `settlePaymentRequestIfPaid` now calls `verifyTransactionLanded(rpc, found.signature)` between `findReference` and `validateTransfer`. Any non-ok verdict (`not_confirmed`, including processed-only and errored transactions, or `not_indexed`) is treated as "no valid payment yet": the row is returned unchanged and the next reconcile re-checks. Money state (`recordInboundTransfer` + `paid`) is written only after two independent checks agree on the signature.
- The status read gains transient RPC retries this path previously lacked.
- Tests: two new cases (unconfirmed signature does not settle; confirmed-but-unindexed does not settle); existing cases updated to mock the helper.

## Testing

`pnpm --filter @sdp/api exec vitest run src/services/payments/payment-requests.test.ts`: 12/12 pass; typecheck and lint clean.


resourcefulmind (dismissed):
This one writes to the settlement path and the cross-check is also fail closed too. Any non-ok verdict returns the row unchanged before validateTransfer
and before the paid write, and both new tests assert that. No branch prefers one source over the other.
Awesome, approving!

multipletwigs (approved):

PR #1432, opened 2026-08-20
fix(rpc): classify transient Solana JSON-RPC server codes as retryable

G1de0n:
## Summary

Stacked on #1430. `isTransientRpcError` only recognized transient failures by HTTP status and transport error text, so Solana JSON-RPC server errors that are transient by definition — `-32005` (node unhealthy/behind), `-32014` (block status not yet available), `-32019` (failed to query long-term storage) — were thrown without retry. In production this surfaced as a 500 + Sentry issue on `GET /v1/payments/transfers` when the RPC provider's archival storage timed out with `-32019`, an error whose own message says "please try again".

## Changes

- `packages/sdp-rpc/src/transient.ts`: the classifier now also matches these codes (both the `@solana/errors` "Solana error #-32019" form and the raw server message text).
- Tests: code form retried, raw message form retried, and a guard that ordinary numbers like `0x32019` are not misread as codes.

## Testing

`pnpm --filter @sdp/rpc test`: 15/15 pass; typecheck and lint clean.

### DIFF
--- a/packages/sdp-rpc/src/transient.ts
+++ b/packages/sdp-rpc/src/transient.ts
@@ -2,6 +2,13 @@
  * Shared RPC helpers used by the Solana RPC layer and the fee-payment adapters.
  */
 
+import {
+  isSolanaError,
+  SOLANA_ERROR__JSON_RPC__SERVER_ERROR_BLOCK_STATUS_NOT_AVAILABLE_YET,
+  SOLANA_ERROR__JSON_RPC__SERVER_ERROR_LONG_TERM_STORAGE_UNREACHABLE,
+  SOLANA_ERROR__JSON_RPC__SERVER_ERROR_NODE_UNHEALTHY,
+} from "@solana/kit";
+
 // Overloaded-gateway / timeout HTTP statuses worth retrying.
 const TRANSIENT_HTTP_STATUS = /\b(408|429|500|502|503|504)\b/;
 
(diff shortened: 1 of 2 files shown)


packages/sdp-rpc/src/transient.ts
```diff
@@ -20,9 +20,20 @@ const TRANSIENT_ERROR_TEXT =
...
 export function isTransientRpcError(error: unknown): boolean {
   const message = error instanceof Error ? error.message : String(error);
-  return TRANSIENT_HTTP_STATUS.test(message) || TRANSIENT_ERROR_TEXT.test(message);
+  return (
+    TRANSIENT_HTTP_STATUS.test(message) ||
+    TRANSIENT_ERROR_TEXT.test(message) ||
+    TRANSIENT_SOLANA_RPC_CODE.test(message) ||
+    TRANSIENT_SOLANA_RPC_TEXT.test(message)
+  );
 }
```
multipletwigs: hey andrey, let's use the error codes from kit instead of regexing the formatted message — `@solana/kit` already re-exports `isSolanaError` and the typed JSON-RPC server error constants (`SOLANA_ERROR__JSON_RPC__SERVER_ERROR_NODE_UNHEALTHY` / `..._BLOCK_STATUS_NOT_AVAILABLE_YET` / `..._LONG_TERM_STORAGE_UNREACHABLE`), and `@solana/kit` is already a dependency here. Matching on the code directly avoids depending on `@solana/errors`' message wording staying stable across versions.

```suggestion
// Transient Solana JSON-RPC server codes: node unhealthy/behind, block
// status not yet available, long-term storage unreachable.
const TRANSIENT_SOLANA_RPC_CODES = [
  SOLANA_ERROR__JSON_RPC__SERVER_ERROR_NODE_UNHEALTHY,
  SOLANA_ERROR__JSON_RPC__SERVER_ERROR_BLOCK_STATUS_NOT_AVAILABLE_YET,
  SOLANA_ERROR__JSON_RPC__SERVER_ERROR_LONG_TERM_STORAGE_UNREACHABLE,
] as const;

export function isTransientRpcError(error: unknown): boolean {
  const message = error instanceof Error ? error.message : String(error);
  return (
    TRANSIENT_HTTP_STATUS.test(message) ||
    TRANSIENT_ERROR_TEXT.test(message) ||
    TRANSIENT_SOLANA_RPC_CODES.some((code) => isSolanaError(error, code))
  );
}
```

Also needs the import at the top of the file:
`import { isSolanaError, SOLANA_ERROR__JSON_RPC__SERVER_ERROR_NODE_UNHEALTHY, SOLANA_ERROR__JSON_RPC__SERVER_ERROR_BLOCK_STATUS_NOT_AVAILABLE_YET, SOLANA_ERROR__JSON_RPC__SERVER_ERROR_LONG_TERM_STORAGE_UNREACHABLE } from "@solana/kit";`

(drops the raw-text regex fallback for these three — if there's a caller that can see one of these as a bare pre-formatted string rather than a thrown `SolanaError`, flag it and we can keep a narrower text fallback for just that case.)

resourcefulmind (dismissed):
Quick pass, this looks good. The retry-amplification worry does not apply, every consumer is bounded.

G1de0n:
Good call, done in 6900a6c — dropped the regex for `isSolanaError` against the three typed codes from kit. Verified the constants resolve to exactly the values I was matching (-32005 / -32014 / -32019), so the coverage is the same for anything thrown by the RPC layer, and it can no longer false-positive on unrelated text.

One consequence worth noting: the typed check only matches real `SolanaError` instances, so an error re-thrown as a plain `Error` carrying the same message would no longer be retried. I checked every caller of `isTransientRpcError` (`withTransientRpcRetry`, `getSignatureStatusOrNull`, and the native adapter's submit retry) — all three receive the error straight from the kit RPC call, so nothing regresses today. Worth remembering if we ever start passing wrapped errors into it.

Tests now build real `SolanaError`s instead of asserting on message wording.

multipletwigs (dismissed):

multipletwigs (approved):

PR #1433, opened 2026-08-20
fix(payments): park transfers with unknown submission outcomes

pashpashkin:
## What changed

A transfer whose submission outcome is unknown is no longer recorded as `failed`.

- `FeePaymentError` now carries `maybeBroadcast` and `preBroadcast` across Kora retries. One shared
  boundary classifies the final verdict for SOL, SPL, and confidential transfers.
- Proven pre-broadcast failures remain terminal. Ambiguous outcomes return
  `409 transfer_submission_outcome_unknown` and park the transfer as `processing` with
  `provider_data.submission_outcome = 'unknown'`.
- The submitted signature is persisted immediately with one retry. If both writes fail, the transfer
  parks with the signature in `provider_data.submitted_signature`.
- Recovery excludes signatureless parked rows and checks transaction history before failing an aged
  signed transfer.
- Sponsorship uses the same verdict: proven pre-broadcast failures release the reservation; ambiguous
  outcomes remain `charged_unknown`.
- OpenAPI and the operations runbook document the new response and reconciliation procedure.

## Why

`main` can report `failed` after a transaction may already have landed: when Kora loses its response,
when a retry follows an ambiguous attempt, or when the signature write fails. A client or recovery job
may then submit the payment again.

## Impact

- `POST /v1/payments/transfers` may return `409` with
  `error.details.reason = transfer_submission_outcome_unknown`. Do not retry it without the same
  `Idempotency-Key`.
- A submitted but unconfirmed transfer now returns `200` with `status: processing` and its signature,
  then settles asynchronously.
- Deterministic Kora refusals now return the appropriate `400` or `429`; ambiguous provider outcomes
  park instead of becoming terminal.
- Parked transfers wait for operator reconciliation. Other activity on the wallet is unaffected.

## Pre/post release actions

- **Pre:** none. No schema change, migration, or feature flag.
- **Post:** monitor and reconcile rows where `status = 'processing'` and
  `provider_data ->> 'submission_outcome' = 'unknown'`; see
  `docs/ops/release-operations.md`.
- **Rollback:** resolve parked rows before deploying a pre-fence image. The old recovery job will mark
  unresolved rows `failed`, after which clients may retry them. Record any unresolved ids if an
  emergency rollback cannot wait.

## Result Validation

- Final stacked head: 267 API test files, 3412 passed, 14 skipped, 0 failed; API typecheck and
  changed-file Biome checks pass.
- Coverage includes verdict precedence, Kora retries and refusals, route parking and replay,
  submitted-but-unconfirmed responses, lost signature writes, sponsorship accounting, recovery
  exclusion, and transaction-history lookup.

Successful transfer submission, authorization, policy, and idempotency behavior are unchanged.
Failure behavior changes for `POST /v1/payments/transfers`. The shared Kora transport and error mapping
also affect other Kora-backed consumers. Batch and recurring parking follow in the next PR.

## Beyond spec

1. Submitted-but-unconfirmed transfers now return `200 + processing` instead of an error and a
   terminally failed row.
2. Sponsorship budgeting reuses the same submission verdict and releases only proven pre-broadcast
   reservations.
3. Kora refusals use the pinned provider enum and one response-aware HTTP client in every
   configuration. This corrects `400`/`429` responses and deployments without a Cloud Run audience.
4. A processing transfer with a signature is checked against transaction history before being
   failed, adding one bounded archival RPC lookup when aged candidates exist.
5. Signature persistence retries at submission time, including when the caller never reads the
   returned row.
6. Redis-backed tests use a database per Vitest worker instead of `FLUSHALL` on a shared instance.

## Question for the API owner

If both signature-column writes fail, the durable row parks with `submitted_signature`, but the
current response is `200 + processing` with the signature. Should this case return the same `409` as
other unresolved outcomes?

### Residual risks

- `409` is advisory: a client that submits again without the same `Idempotency-Key` can still
  double-pay.
- If both marker writes fail during a DB outage, recovery may later time out the unmarked row. The
  failure is logged but cannot be closed with another DB write.
- Parked rows wait indefinitely and currently have no dedicated alert.
- The classifier treats Kora HTTP 4xx responses other than 408/499 as pre-broadcast; that assumption
  must remain true for the deployed provider path.

## Known gaps

`signAndSend` fuses signing with sending, so a signature only exists once the
transaction may already be on chain. That is why this PR has to guess a broadcast
verdict from provider error codes and text, and why a row it cannot resolve waits
for an operator.

Splitting them — `signAsFeePayer` plus a local `sendTransaction`, as
`services/earn/vault-execution.service.ts` already does — makes the signature known
before anything reaches the network: a failure then either provably sent nothing, or
carries a signature the reconciler settles from chain. No parking, no error
classification, no migration.


pashpashkin:
@greptileai review please

pashpashkin:
@greptileai review please

pashpashkin:
@greptileai review please

apps/sdp-api/src/services/jobs/track-pending-transfers.ts
```diff
@@ -291,61 +364,64 @@ async function syncProcessingTransfersOnChain(
...
-          transferId: transfer.id,
-          status: "failed",
-          slot: Number(status.slot),
-          error: JSON.stringify(status.err),
-          updatedAt: nowIso,
-        });
-        continue;
-      }
+  for (let i = 0; i < agedOut.length; i++) {
+    await applyOnChainVerdict(env, repo, agedOut[i], archived[i] ?? null, nowIso);
```
pashpashkin, line 394: Keeping this sequential intentionally: each iteration writes to Postgres, and Promise.all could enqueue up to 256 writes against a 10-connection pool

apps/sdp-api/src/services/jobs/track-pending-transfers.ts
```diff
@@ -291,61 +364,64 @@ async function syncProcessingTransfersOnChain(
...
+      now.getTime() - new Date(transfer.updated_at).getTime() > STUCK_PROCESSING_AFTER_MS
+  );
+
+  // Persist every verdict already returned by the recent cache before a slow
+  // transaction-history lookup can delay the rest of the selected batch.
   for (let i = 0; i < processingWithSig.length; i++) {
     const transfer = processingWithSig[i];
     const status = statuses[i] ?? null;
+    if (status) {
+      await applyOnChainVerdict(env, repo, transfer, status, nowIso);
```
pashpashkin, line 384: Keeping this sequential intentionally: each iteration writes to Postgres, and Promise.all could enqueue up to 256 writes against a 10-connection pool

pashpashkin:
@greptileai The archival lookup is limited to 30 seconds and failures are caught. Transfers remain safely confirmed, and finalization always runs afterward, so this is bounded latency
Does it make sense?

pashpashkin:
@greptileai The 30-second limit is enforced by our RPC transport, not by the provider. createRpc(env) uses DEFAULT_RPC_REQUEST_TIMEOUT_MS = 30_000 and aborts a stalled request via AbortController. This call has no timeout override or retries

pashpashkin:
@greptileai review please

apps/sdp-api/src/services/jobs/track-pending-transfers.ts
```diff
@@ -291,61 +364,64 @@ async function syncProcessingTransfersOnChain(
...
-          await updateTerminalTransfer(env, repo, transfer, {
-            transferId: transfer.id,
-            status: "failed",
-            error: "Transaction not found on chain",
-            updatedAt: nowIso,
-          });
-        }
-        continue;
-      }
+  const archived = await getArchivedStatuses(rpc, agedOut);
```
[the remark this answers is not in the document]
pashpashkin: This duplicates the concern resolved above. createRpc(env) uses DEFAULT_RPC_REQUEST_TIMEOUT_MS = 30_000 from packages/sdp-rpc/src/solana.ts, where withRequestTimeout aborts stalled requests via AbortController; this call has no retries. Failures are caught and finalization proceeds, so this is bounded latency

pashpashkin:
@greptileai review please

multipletwigs:
Today's ambiguity is a two-cause OR SDP can't split apart:

`Kora fails` = `network fail (SDP↔Kora)` OR `Kora-internal fail (Kora↔validator)`

Splitting sign (Kora) from send (SDP-local) shrinks it to one term SDP owns:

`RPC fail` = `SDP↔validator, response lost after possible forward`

`signAsFeePayer` (kora.adapter.ts:154, port.ts:60) already does sign-only via Kora, and `vault-execution.service.ts` already does this exact split for Earn, sign via Kora, submit locally, treat a throw as unknown-but-reconcilable via the known signature. 

Will address in another pr

multipletwigs (approved):

PR #1434, opened 2026-08-20
fix(payments): extend outcome parking to batches and recurring collections

pashpashkin:
## What changed

The submission-outcome fence from the boundary PR now covers recurring collections and batch chunks.

- An ambiguous recurring collection stays `processing`, keeps
  `provider_data.submission_outcome = 'unknown'`, returns
  `409 transfer_submission_outcome_unknown`, and is excluded from retry and stale recovery.
- An ambiguous batch chunk and its recipients stay `processing`. If its signature cannot be
  journalled, the chunk parks with `provider_data.submitted_signature` instead of cascading
  `failed`.
- Transfers, chunks, and collections now share the same submission verdict, durable marker, and
  operations runbook; transfers and chunks also share signature recording.

## Why

An ambiguous recurring submit could be failed and rescheduled, charging the same cycle twice. An
ambiguous batch chunk could be failed for every recipient and then paid again when the batch was
re-run.

## Impact

- `POST /v1/payments/recurring-payments/{id}/collect` may return `409` while the attempt and transfer
  remain `processing`. The cycle is not rescheduled; do not retry it.
- `POST /v1/payments/transfer-batches` may return `200` with a chunk and its recipients still
  `processing`. A non-null reconciliation `error` no longer implies a failed status.
- Parked rows require operator reconciliation but do not directly block unrelated collections or
  batches.

## Pre/post release actions

- **Pre:** none. No database schema change, migration, or feature flag.
- **Post:** monitor and resolve parked transfers, chunks, collections, and unlinked attempts using
  `docs/ops/release-operations.md`.
- **Rollback:** resolve parked rows before deploying a pre-fence image. Old recovery can mark them
  `failed` and reschedule a recurring cycle, creating a duplicate-charge risk. Record unresolved ids
  if an emergency rollback cannot wait.

## Result Validation

- Final stacked head: 267 API test files, 3412 passed, 14 skipped, 0 failed; API typecheck and
  changed-file Biome checks pass.
- Coverage verifies recurring parking without recollection, recovery past an inert settled marker,
  batch chunk and recipient parking, lost-signature parking, and the shared submission chokepoint.

Only ambiguous failure paths for recurring collection and batch execution change. Recurring
activation, update, cancel, and resume submit as before, although lifecycle changes blocked by a
parked cycle now return the reconciliation `409`. Issuance, workflows, and private channels are
unchanged.

## Beyond spec

1. A batch chunk whose signature write is lost now parks with `submitted_signature` instead of
   failing the chunk and its recipients.
2. Parked batch recipients carry the reconciliation message in `error` while remaining `processing`;
   the public schema now states that non-null `error` does not imply failure.
3. Parked collection attempts are excluded from stale recovery to prevent recollection and queue
   starvation. A retained marker is inert once its transfer leaves `processing`.
4. Parking back-links an unlinked recurring attempt to its transfer; the runbook includes a sweep if
   that write also fails.
5. The shared recorder log no longer uses the transfer-specific `createTransfer:` prefix.

## Question for the API owner

A parked current cycle blocks cancellation until reconciliation. Blocking update and resume prevents
orphaning the attempt, but cancel does not move the schedule. Should cancellation be allowed while a
charge is unresolved? No further collection occurs while the cycle is parked.

### Residual risks

- If both marker writes fail during a DB outage, recovery may later time out the unmarked row.
- If the recurring attempt-to-transfer backlink also fails, stale recovery can reschedule the cycle;
  this is logged and covered by the runbook sweep.
- Parked rows have no dedicated alert, and parked recurring cycles leave the cron queues quietly.

## Known gaps

See https://github.com/solana-foundation/solana-developer-platform/pull/1433 — this PR consumes that contract and shares its gap.

### DIFF
--- a/apps/sdp-api/src/openapi/paths/payments.ts
+++ b/apps/sdp-api/src/openapi/paths/payments.ts
@@ -314,7 +314,7 @@ export function registerPaymentsPaths(registry: OpenAPIRegistry) {
     summary: "Create transfer batch",
     operationId: "createPaymentTransferBatch",
     description:
-      "Executes a custody-signed outbound transfer batch to counterparty crypto-wallet accounts, chunks recipients into Solana transactions, and returns the batch, recipient, and transfer records. Supply an Idempotency-Key to retry safely: an identical resolved request returns the original batch without another on-chain submission, while reusing the key for a different request returns 409.",
+      "Executes a custody-signed outbound transfer batch to counterparty crypto-wallet accounts, chunks recipients into Solana transactions, and returns the batch, recipient, and transfer records. Supply an Idempotency-Key to retry safely: an identical resolved request returns the original batch without another on-chain submission, while reusing the key for a different request returns 409. A chunk whose provider outcome is unknown comes back inside a 200 as a processing transfer without a signature and with a reconciliation notice in error, its recipients left processing — it may already be on chain, so do NOT re-send that batch.",
     security: [{ apiKeyAuth: [] }],
     request: {
       headers: projectScopeWithIdempotencyHeaders,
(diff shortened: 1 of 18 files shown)


pashpashkin:
@greptileai review please

multipletwigs (dismissed):
should minimize send surface to just sdp-api, which will be addressed in a followup ticket

multipletwigs:
On the cancel question: allow it.

Cancelling doesn't touch the schedule, so it can't orphan anything the way update/resume would. And it can't change what already happened to that charge — the submission already went out; cancelling now doesn't make it more or less landed. Whatever it resolves to comes out in reconciliation regardless of subscription status, so the customer finds out then, not from whether cancel was blocked.

Blocking it doesn't add safety, just makes someone wait on an unrelated ambiguous RPC response to cancel their own subscription. Let cancel through while parked.

PR #1437, opened 2026-08-20
fix(i18n): retranslate stale catalog entries instead of wedging the release flow

keeganthomp:
Release Flow's `prepare-release-pr` job has failed on ~90% of merges to `main` since 2026-08-04 (149 of 162 total failures across 645 runs). This unblocks it and removes the failure class.

## Root cause

`04f585e5` (#1065) changed en `channelBalanceAmount` from `{amount} USDC` to `{amount} {symbol}` at 14:08 UTC on Aug 4. The Release Flow run for that very commit is the first failure. `fr` and `pt` kept the old value.

`collectMissingTranslations` only queued **absent** keys, so it never re-sent them. `validateCatalogs` rejected **present** ones whose placeholders drifted. Detector blind, validator fatal, and `setCatalogValue` refused to overwrite regardless. Every run generated ~2955 correct translations over 33 minutes, then discarded all of them over 2 keys it could not touch.

## Changes

| Change | Why |
| --- | --- |
| `fr`/`pt` `channelBalanceAmount` | one-line repair, unblocks now |
| `translationDefects()` | collector and validator share one predicate |
| `setCatalogValue(…, { overwrite })` | stale entries repair instead of throwing |
| `translateMissingEntries` | a failed batch defers; it no longer aborts the run |
| `translate-release-strings` job | own `continue-on-error` job, reports on the release PR |

## Reviewer notes

- **The invariant is the fix:** the validator cannot reject what the collector will not retranslate. Test `the validator rejects nothing the collector fails to queue` asserts it directly.
- Terminology now checks per source key rather than the whole target catalog, so orphaned keys (deleted from en) are ignored. Deliberate: nothing can retranslate a key with no source string, so that was a second latent wedge.
- Batch tolerance is a separate fix from the placeholder bug. It addresses the 15 of 87 post-Aug-13 jobs that died on Eve 503s before validation ever ran.
- **Failure reporting:** the job is `continue-on-error`, so any throw before the summary comment used to be a green run with no comment at all. The catch now posts a `failed` comment, and skips it when a summary already went out so `partial` is not clobbered. The runbook states plainly that the catalog gate no longer runs on the push to `main`.
- **Not fixed here:** `release-flow.mjs:394` force-recreates `sdp/release-main`, so a translation commit survives only if the release PR merges before the next push to `main`. Affects drainage, not failure. Follow-up.
- First green run translates the full 2957-key backlog, roughly 30 minutes. Later runs are small.

## Verification

85/85 script tests (3 rewritten for the defer contract, 11 new). biome 18 warnings, 0 errors (14 at baseline; the 4 new ones are `noUndeclaredEnvVars` on CI-only `GITHUB_*` reads, which do not belong in turbo.json `globalEnv`). actionlint and action-pins clean. Simulated a full run against the real catalogs: 2959 entries applied, validator clean. Catalog-change policy classifies this `translation-only`.


resourcefulmind (dismissed):
I looked at this because it gates the release flow. No logic defect.

I wanted to mention though, after this change the repo's only caller of
validateCatalogs sits inside a continue-on-error job. That is a change to the release gate rather
than just to this PR, and it belongs in the runbook. Happy to be corrected on this if it's not the case.

Also related, if an English key changes shape the apply throws before the summary comment is posted, so the
run stays green and the release PR carries no comment at all. A failure comment from the catch would
cover it.

keeganthomp:
Both correct, thanks. Fixed in 6551ce46.

**The gate.** You're right, and it's worse than "only caller": `validateCatalogs` has no caller outside this job at all, so a catalog defect used to fail Release Flow on the push to `main` and now can't. The runbook says that outright now instead of implying the release PR still gates it for free.

**The silent throw.** Confirmed, and it reproduces today. If an en string leaf becomes a nested object, the collector queues the new key as non-stale and `applyTranslations` throws `Cannot add wallet.balance.label: balance is not an object`. Same silence for `createTranslationCommit` failures and the introduced-drift throw: every one of them lands before `updateReleasePrComment`, so the run is green and the release PR carries nothing.

The catch now posts a `failed` comment with the error and a link to the (green) run, under the same marker so the next run replaces it. One wrinkle worth flagging: the deferred-batch throw already posts its `partial` summary before throwing, so a blind failure comment would have overwritten the more useful one. That's a `summaryReported` guard, extracted as `failureReport()` so the decision is testable rather than buried in the catch.

The script had no test seam (it ran `main()` on import), so I added one. The opt-out is inverted on purpose: anything short of an explicit `TRANSLATION_SCRIPT_IMPORT_ONLY=1` still runs, since a sync that silently skips itself is exactly the failure class this PR is about. 4 new tests, 85/85.

GuiBibeau (approved):
Thanks for unblocking

resourcefulmind (approved):

PR #1442, opened 2026-08-21
fix(rpc): destroy the stored secret when a BYOK connection is deactivated

G1de0n:
## Summary

Deactivating a BYOK RPC connection flipped the connection row and left the customer's API key exactly where it was — the Secret Manager version stayed readable indefinitely. The customer has withdrawn the credential from our platform; from their side it is revoked, and from ours it is still stored. That is the gap this closes.

## Changes

`deactivateRpcConnection` now reads the credential before flipping the row, and once deactivation succeeds destroys the stored secret version. It mirrors what the custody credential path already does on deactivation, including the shape of the failure handling:

- Best effort. A failed destroy does not fail the request — the connection is already deactivated and the relay has stopped using it — but it logs `rpc_connection_secret_orphan_risk` with the credential id, backend and request id so an operator can clean up.
- Scoped to `gcp_secret_manager`, the only backend that supports destroying a version. The `encrypted_db` backend rejects `destroyVersion` outright; ciphertext cleanup there is a separate problem and is untouched here.
- The credential read moves before the state change so the version ref is captured while the row still points at it.

No change to reactivation, which remains refused by design — rotation means creating a new connection.

## Testing

New `src/services/rpc-connection-deactivation.test.ts` seeds a connection whose credential records the GCP backend, deactivates it through the service, and asserts `destroyVersion` was called with that version ref; a second case pins that a repeat deactivation still conflicts. `pnpm --filter @sdp/api exec vitest run src/services` — 948 passed. Typecheck and lint clean.

Linear: HOO-976


apps/sdp-api/src/services/rpc-connection.service.ts
```diff
@@ -377,24 +378,54 @@ export async function deactivateRpcConnection(
...
     organizationId: auth.organizationId,
     connectionId,
     scopeKeys,
   });
-  if (!deactivated) {
-    throw conflict("The RPC connection is already deactivated");
-  }
 
-  const credential = await store.findConnectionSecret({
+  const deactivated = await store.deactivateConnection({
```
G1de0n: Not taking this one — the ordering is load-bearing rather than incidental.

The read must happen before the state change because it captures the secret version ref of the credential we are about to destroy. The two queries do hit different tables today (the read selects from provider_credentials, the update writes rpc_connections), so Promise.all would return the same values right now — but that is a property of the current implementation, not something the call site guarantees. Running them concurrently would leave the read-then-mutate intent invisible in the code, and the day deactivation starts touching the credential row the race appears silently.

The cost is one database round-trip on an admin endpoint that runs rarely, so there is nothing measurable to win here.

pashpashkin (approved):

PR #1443, opened 2026-08-21
fix(custody): HOO-1005 Deepsec: harden provider credentials and redirects

eldarik:
Closes [HOO-1005](https://linear.app/solana-fndn/issue/HOO-1005/deepsec-harden-custody-provider-credentials-and-redirects) (DeepSec audit series)

Fixes the four confirmed DeepSec custody-provider findings across DFNS, Coinbase CDP, and Utila. All of them are credential-exposure or token-lifetime issues in the provider clients — no route, schema, or DB change.

## What was wrong

1. **DFNS followed redirects to any origin, carrying the bearer token.** `dfnsRequestRaw` correctly used `redirect: "manual"`, but then manually re-issued a redirected `POST` as a `GET` to whatever `Location` said — with `Authorization: Bearer <DFNS_AUTH_TOKEN>` and the `X-DFNS-*` headers attached. A `302` to an attacker origin handed over a live credential.
2. **DFNS and Utila errors interpolated raw upstream bodies, URLs, and credential ids into messages.** `truncateForError` put 200–300 characters of the upstream response into `SigningError.message`, alongside `url=` and — on user-action rejection — both the configured credential id *and* every credential id allowed on the account. Provider error bodies routinely echo the request, including the `Authorization` header.
3. **The keychain adapters' `sign()` catch blocks re-published provider SDK errors verbatim** onto `SignResult.error` with no redaction. Not reachable from a live route today (`SigningService.sign()` has no non-test caller), but it is a landmine and it contradicts the redaction contract the rest of the codebase enforces.
4. **Coinbase CDP wallet-auth JWTs had no `exp`.** A leaked `X-Wallet-Auth` token stayed replayable forever, and the bearer token was missing the `aud` CDP documents for it.

### Upstream error summarization (`packages/sdp-custody/src/redaction.ts`)
New `summarizeUpstreamErrorBody(rawBody, httpStatus?)` reduces any upstream error body to a single machine-readable code that is safe to log and to return. Only a whitelisted set of code fields is read (`errorCode`, `error_code`, `errorType`, `code`, `status`, `type`, `reason`), at the top level or under an `error` envelope; the value must be identifier-shaped (`/^[A-Za-z0-9_.:-]{1,64}$/`), so anything with whitespace — i.e. all prose — is rejected. Codes that merely restate the HTTP status are skipped, so Google-style `{ error: { code: 400, status: "INVALID_ARGUMENT" } }` reports the descriptive half. Everything else, including unparsable bodies and arrays, collapses to `unavailable`.

`isSensitiveCredentialKey` also now matches `credentialid`, covering `credentialId` / `allowedCredentialIds` in any future `context`/`details` payload — a provider credential id identifies the signing key it unlocks and must not reach a log sink.

### DFNS redirect containment (`packages/sdp-custody/src/dfns/client.ts`)
- Redirect handling moves into a dedicated `followDfnsRedirect`, which resolves `Location` against the request URL and compares **`URL.origin` objects**, not string prefixes — so protocol-relative `//evil.example` and an unparsable `Location` both fail closed rather than slipping through.
- Cross-origin redirect → `NETWORK_ERROR` naming only the refused origin, and the follow-up request is never made. Same-origin `POST` → `GET` follow is preserved unchanged, since that pattern looks intentional on DFNS's side.
- A redirected **GET** now throws a distinct `unsupported redirect` instead of falling through the generic `!response.ok` branch. Same error class, same code — see Deviations.

### DFNS / Utila error messages (`dfns/client.ts`, `provisioning/index.ts`)
- API errors, non-JSON responses, and redirect follow-up failures now report `status=` / `contentType=` / `code=<whitelisted>` and nothing else. The raw body, `url=`, and `location=` are gone.
- User-action signing rejection no longer names any credential id — just how many the account allows.
- Utila `CreateWallet` failures report `code=<whitelisted>` instead of 300 characters of body.

### Keychain adapter redaction (`keychain/base-keychain.adapter.ts`, `keychain/keychain-utila.adapter.ts`)
Both `sign()` catch blocks run the message through `redactCredentialString` before putting it on `SignResult.error`.

### Coinbase CDP JWT claims (`packages/sdp-custody/src/provisioning/coinbase.ts`)
Claim values and TTLs are named constants. The wallet-auth token gains `nbf` and `exp = now + 60`; the bearer token gains the documented `aud: ["cdp_service"]`, keeping `iss: "cdp"` and `exp = now + 120`.

## Deviations worth reading before review

1. **No `SigningError.context` field.** Rather than relocating the body and credential id onto a new `context` property, both are dropped entirely in favour of the whitelisted code. The acceptance bullet forbids raw bodies reaching **logs** too, and routing the body to `context` would just re-apply the same regex scrubber. This also keeps the `SigningError` constructor unchanged.
2. **`iss`/`aud` deliberately NOT added to the wallet-auth JWT.** CDP's auth docs list the wallet-auth required claims as `iat`/`nbf`/`jti`/`uris`/`reqHash` with a 1-minute validity, and document `iss: "cdp"` / `aud: ["cdp_service"]` for the **bearer** token only; `@coinbase/cdp-sdk@1.54.0`'s `generateWalletJwt` sets no `iss`, `aud`, or `exp`. Asserting a claim CDP validates against a strict set would reject every JWT and take down **all account creation**. So acceptance bullet 3 lands as "short expiry everywhere, issuer/audience on the token CDP documents them for". Only a live CDP account creation can settle whether more is safe — see QA below.
3. **Finding 2c (Coinbase raw body in its error message) deferred.** `isCoinbaseCdpAlreadyExistsError()` classifies 409s by sniffing `already_exists` out of that message, and that predicate gates the lookup-by-name fallback that makes re-provisioning idempotent. Changing the message without rewriting the predicate would make re-provisioning hard-fail. The message is already run through `redactCredentialString`. Whoever picks it up must change both together.
4. **Tests live in `apps/sdp-api`, not `packages/sdp-custody`.** The package has no test runner wired in; these specs stub `globalThis.fetch` following `provisioning.test.ts`'s existing convention and run in CI today via the API suite. Wiring a runner into `sdp-custody` stays open.

## Second commit: `@types/node` for `sdp-custody`

`packages/sdp-custody` declared no `devDependencies`, so `tsc --noEmit` fails there with 20 errors for globals it uses everywhere (`node:buffer`, `node:crypto`, `fetch`, `Response`, `btoa`, `AbortSignal`, `TextEncoder`). This is pre-existing — checking out the base commit's `src/` and re-running gives the identical 20 errors — but CI only typechecks packages a PR changes, and `sdp-custody` is otherwise never typechecked, so this PR is what first trips it. Pinned to `26.1.2`, the version all nine sibling packages already declare. The lockfile gains only the importer entry, and `pnpm install --frozen-lockfile` accepts it without re-resolving anything else.

## QA

**Automated (all green).** Full API suite **253 files / 3156 passed / 14 skipped**; workspace `pnpm typecheck` **19/19** (verified uncached); `pnpm check:module-boundaries`; biome clean on every changed file. New coverage: `dfns-client.test.ts` (10), `keychain-adapter-redaction.test.ts` (2), plus CDP claim assertions and two Utila error variants in `provisioning.test.ts` and summarizer cases in `redaction.test.ts`.

**Local fake-provider harness — the only place this behavior is *positively* observable.** `getSafeSigningErrorMessage` rewrites `NETWORK_ERROR` into a fixed generic sentence, and the `SigningError` branch of `app.onError` returns *before* the logging/Sentry block, so a deployed run can only show the **absence** of leakage, never the new message shape. So these were driven against two real `node:http` servers on separate origins, calling `@sdp/custody` directly with `fetch` unstubbed:

| Case | Result |
| --- | --- |
| absolute cross-origin 302 | `DFNS API returned a cross-origin redirect (POST /wallets): status=302 redirectOrigin=http://127.0.0.1:8792; refusing to forward credentials` |
| protocol-relative 307 host swap | same refusal, `status=307` |
| same-origin 302 POST→GET | followed as before → `wallet id=wa_created_1` (legitimate flow intact) |
| GET receiving a 302 | `DFNS API returned an unsupported redirect (GET /wallets/wa_get_redirect): status=302` |
| 403 JSON with `Bearer <token>` in `message` | `Utila CreateWallet failed (403): code=PERMISSION_DENIED` |
| 502 non-JSON HTML containing a token | `Utila CreateWallet failed (502): code=unavailable` |
| **evil origin tally** | **0 requests recorded, 0 credential headers** — the bearer token never left the provider origin |

Every message was swept for the planted DFNS auth token, the planted credential id, the planted Utila service-account token, and the substrings `Bearer `, `-----BEGIN`, `<html>`, `is not permitted`, `authorization:` — all absent. All six errors carry `code = NETWORK_ERROR`.

**Still outstanding — needs a reviewer's attention.** Live **Coinbase CDP account creation** against the real sandbox is the one check nothing local can substitute for: account creation is what actually sends `X-Wallet-Auth`, so it is the only way to prove CDP accepts the new `exp` and the bearer's new `aud` rather than rejecting the JWT outright. A 401/403 from CDP there means Deviation #2 must be revisited. Everything else in the hosted pass can only assert absence of leakage, per the masking above.

## Acceptance criteria

- [x] DFNS never forwards `Authorization` across origins; redirects are same-origin only — proven with a zero-request evil origin, not just message text
- [x] DFNS and Utila errors cannot expose credential ids, secrets, or raw upstream bodies to logs or API clients
- [ ] Coinbase wallet-auth JWTs have a short expiry and expected audience/issuer — **expiry done everywhere; `iss`/`aud` set on the bearer token only, deliberately (Deviation #2), pending live CDP verification**
- [x] Tests cover redirects, redaction, expiry, and upstream error variants


multipletwigs:
hey @eldarik i merged the earlier redaction package - could we move all the redaction logic to the package instead from sdp-custody?

eldarik:
@multipletwigs fixed

multipletwigs (approved):

PR #1447, opened 2026-08-21
feat(api): HOO-1000 enforce tenant isolation at the database with RLS identities

arseniy-nikitochkin:
## Summary

Database-level defense-in-depth for tenant boundaries (HOO-1000, a sub-issue of HOO-976 , SDP-005 residual risk): forced row-level security on every tenant-owned table, explicit scoped database identities for requests and system workers, DB-enforced uniqueness for counterparty provider lookup keys, and a narrow, independently audited operational bypass.

## How it works

- **Migration `0063_tenant_isolation_rls.sql`** — `ENABLE` + `FORCE ROW LEVEL SECURITY` with policies on: 54 tables carrying `organization_id` (plus `organizations` on `id`), 15 child tables via parent-delegating `EXISTS` policies, the polymorphic sponsorship-budget tables (tenant read-only, privileged write), and a tightened `audit_logs` INSERT policy. Policies key off transaction-local GUCs (`app.tenant_isolation_identity`, `..._organization_id`, `..._actor`). **No identity ⇒ deny** (reads empty, writes rejected).
- **Identity stamping** (`src/db/identity.ts`, `src/db/client.ts`) — an AsyncLocalStorage identity is stamped onto every statement/transaction with `set_config(..., is_local = true)`, so pooled connection reuse cannot leak identities and no `getDb()` call site changes.
- **Request identity** — auth middlewares (API key / Clerk / session / replay) resolve credentials under a named `system` identity, then narrow the request to the caller's organization. **Worker identity** — webhooks, `/pay`, login routes, public token metadata, invitation-accept, cron ticks, the managed job, migration runner, seeds, and ops scripts each declare a named system identity at their entry point. Grant sites are locked down by registry tests in `src/lib/tenant-boundary.test.ts`.
- **Operational bypass** — `runWithOperatorDatabaseAccess` (`src/db/operator-access.ts`) grants the `operator` identity only after writing actor + reason to the append-only, externally anchored audit ledger; a static test keeps it out of application code.
- **Migration `0062_counterparty_provider_lookup_integrity.sql`** — makes the *effective* provider lookup key (`COALESCE(denormalized column, provider_data JSON path)`) unique for active counterparties (BVNK + Mural), closing the dual-write window where one webhook reference could match two tenants. Scoped repository paths keep writing provider data; racing claims fail with a clean unique violation.

## Acceptance mapping

- *RLS or equivalent enforced boundary covers tenant records* → migration 0063 + coverage ratchet (`src/db/migrations/tenant-isolation-coverage.test.ts`) that fails when a new table ships without an isolation decision.
- *Request and system-worker identities have explicit scoped access* → middleware narrowing + named system identities at every entry point, registry-tested.
- *Counterparty provider lookup columns: DB-enforced tenant uniqueness, writable through scoped repository paths* → migration 0062 + `mutateProviderData` tests under tenant identity.
- *Cross-tenant reads/writes and concurrent provider-data updates fail safely in DB-backed tests* → `src/db/tenant-isolation.test.ts` (raw SQL: denied/zero-row/unique-violation) and `src/routes/tenant-isolation-http.test.ts` (tenant A's key gets 404 on tenant B's records through the full stack). Tests connect through a new `NOSUPERUSER NOBYPASSRLS` `sdp_runtime` role — superusers bypass RLS, so the old test connection would have proven nothing.
- *Operational bypass is narrow and independently audited* → single audited entry point, import-restricted, ledger-recorded (hash-chained + Redis-anchored).

## Deployment notes

- Enforcement is real only when the runtime `DATABASE_URL` role is `NOSUPERUSER`/`NOBYPASSRLS` — the same contract `docs/ops/audit-ledger.md` already mandates; `pnpm --filter @sdp/api audit:ledger verify` reports the posture. `FORCE` covers non-superuser owner credentials.
- 0062 will fail to apply if production data already holds a cross-tenant duplicate effective reference — that failure is the finding, not a bug; resolve the duplicate and re-run.
- Ops guide: `docs/ops/tenant-isolation.md`.
- Per-statement identity stamping adds a transaction wrap (~2–3 extra round trips) per out-of-transaction statement; explicit transactions gain one `set_config` round trip.


apps/sdp-api/src/db/migrations/postgres/0081_tenant_isolation_rls.sql
```diff
@@ -0,0 +1,370 @@
...
+      'CREATE POLICY sdp_tenant_isolation_delete ON %I FOR DELETE'
+      || ' USING (sdp_tenant_isolation_is_privileged())',
+      scoped_table
+    );
+  END LOOP;
+END $$;
+
+-- ---------------------------------------------------------------------------
+-- audit_logs already has forced RLS from 0047 as an append-only ledger.
+-- SELECT stays USING (true): the hash-chain trigger must read the previous
```
GuiBibeau: This SELECT policy still uses `USING (true)`, so a tenant or a request with no database identity can read audit rows from every organization. That breaks the tenant isolation and fail-closed guarantees in this PR. Can the hash-chain trigger get separate privileged access while normal reads stay scoped to the current organization?

apps/sdp-api/src/db/identity.ts
```diff
@@ -0,0 +1,167 @@
...
+    throw new DatabaseIdentityError(`${label} is required for a database identity`);
+  }
+  return trimmed;
+}
+
+export function currentDatabaseIdentity(): DatabaseIdentity | undefined {
+  return identityStorage.getStore() ?? defaultIdentity;
+}
+
+export function runWithDatabaseIdentity<T>(identity: DatabaseIdentity, fn: () => T): T {
```
GuiBibeau, line 66: This public helper accepts `system` and `operator` identities directly. Any caller can skip `runWithOperatorDatabaseAccess` and get cross-tenant access without writing an audit record. Can this low-level helper stay private, with only the approved tenant, system, and audited operator entry points exported?

docs/ops/tenant-isolation.md
```diff
@@ -0,0 +1,110 @@
+# Database-enforced tenant isolation
+
+Migration `0079_tenant_isolation_rls.sql` puts forced row-level security on
```
GuiBibeau: This guide is out of date with the final code. The RLS migration is `0081`, the lookup migration is `0080`, BVNK no longer uses `bvnk_customer_reference`, and identity stamping now uses `SET LOCAL` instead of `set_config`. Can we update the guide before merging?

apps/sdp-api/src/db/migrations/postgres/0080_counterparty_provider_lookup_integrity.sql
```diff
@@ -0,0 +1,70 @@
...
+  IF mural_conflicts IS NOT NULL THEN
+    RAISE EXCEPTION USING
+      MESSAGE = format(
+        'Duplicate active counterparty provider references must be resolved before this migration can enforce uniqueness. mural: %s. See docs/ops/tenant-isolation.md.',
+        mural_conflicts
+      );
+  END IF;
+END $$;
+
+CREATE UNIQUE INDEX IF NOT EXISTS idx_counterparties_mural_organization_id_effective_active
```
GuiBibeau, line 55: The PR description says both BVNK and Mural provider references get database-enforced uniqueness. This migration only adds the Mural index, while BVNK now uses `counterparty_provider_accounts` with a non-unique provider reference index. Can we add the BVNK constraint here or update the PR if that requirement changed?

apps/sdp-api/src/cron/runner.ts
```diff
@@ -130,10 +131,33 @@ export function startCron(deps: CronDeps): CronHandle | null {
 
   const tasks: ScheduledTask[] = [];
 
+  // Every tick runs under a named system database identity: reconciliation is
```
GuiBibeau, line 134: The revoked API key cache tick below does not use this helper. Without a system database identity, its `api_keys` queries return no rows, so the repair job silently does nothing in in-process deployments. Can that tick use `scheduleSystemTask` too?

apps/sdp-api/src/db/identity.ts
G1de0n: Fixed in 1d4029a48. `runWithDatabaseIdentity` no longer leaves the `@/db` barrel — the exported surface is the kind-specific wrappers plus the audited `runWithOperatorDatabaseAccess`. A new tenant-boundary test pins the raw runner to its four approved files, so a future import of it anywhere else fails CI.

apps/sdp-api/src/cron/runner.ts
G1de0n: Fixed in 1d4029a48. The revoked API key cache tick now goes through `scheduleSystemTask` with `cron:revoked-api-key-cache`, matching every other reconciler; the in-process job path already ran it under `runWithSystemDatabaseIdentity` inside its monitored tick.

apps/sdp-api/src/db/migrations/postgres/0081_tenant_isolation_rls.sql
G1de0n: Fixed in 1d4029a48. The SELECT policy now gates on pg_trigger_depth() > 0 OR sdp_tenant_isolation_allows(organization_id): the hash-chain trigger's prior-row read happens inside the BEFORE INSERT trigger where the depth is non-zero, so the chain never forks, while ordinary reads are scoped like every other tenant table (privileged sees all, a tenant sees its own org, org-less system rows stay privileged-only). The org-less audit_ledger_anchors got the same treatment — trigger reads pass, otherwise privileged-only.

docs/ops/tenant-isolation.md
G1de0n: Updated in 1d4029a48: the guide now names 0080/0081, describes SET LOCAL stamping, and documents the split provider-reference model (mural on the counterparty row with the JSON fallback, BVNK through counterparty_provider_accounts customer links).

apps/sdp-api/src/db/migrations/postgres/0080_counterparty_provider_lookup_integrity.sql
G1de0n: Added in 1d4029a48: BVNK customer links get the same database-enforced uniqueness — a pre-flight names conflicting active links (same stop-with-ids contract as the mural check), then (provider, provider_customer_reference) is unique among active kind=customer_link rows, so a system-scoped webhook lookup can never match two organizations.

apps/sdp-api/src/db/migrations/postgres/0080_counterparty_provider_lookup_integrity.sql
```diff
@@ -0,0 +1,107 @@
...
+-- conflicting ids rather than silently reassigning a provider relationship.
+DO $$
+DECLARE
+  reference_conflicts TEXT;
+BEGIN
+  SELECT string_agg(format('%s/%s -> [%s]', provider, provider_customer_reference, ids), '; ')
+  INTO reference_conflicts
+  FROM (
+    SELECT provider, provider_customer_reference, string_agg(id, ', ' ORDER BY id) AS ids
+    FROM counterparty_provider_accounts
```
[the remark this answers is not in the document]
G1de0n: Fixed in 2e660c1b4, at the state's root rather than by loosening the check: an active link on an archived counterparty was itself the anomaly (unreachable by the runtime lookup, yet occupying the reference). archiveCounterparty now cascades to its provider-account links in the same statement, and 0080 aligns historical rows first — links whose parent is not active are archived before the pre-flight runs. The pre-flight and the unique index therefore cover exactly the runtime lookup's candidate set, and an archived tenant's reference is immediately re-claimable.

GuiBibeau (dismissed):

resourcefulmind (approved):

PR #1448, opened 2026-08-21
feat(earn): quote-derived slippage floors for vault deposits, default…

cutbow7:
…ed and disclosed

The vault deposit dialog could not submit a Veda deposit at all: Veda requires an explicit `minSharesOut` (its SDK refuses an implicit tolerance) and the dialog, built for Kamino where a sandbox floor is optional, collected none — every confirm died on the typed refusal.

**The floor comes from a live quote, never from the amount.** A new capability, `EarnVaultDepositQuoteProvider` / `supportsVaultDepositQuote`, answers "what would this deposit mint right now" from the provider's own accounting; `@sdp/veda` implements it over the SDK's `previewDeposit` — a pure read behind the same kit firewall and proof-then-deadline boundary as every chain call, deliberately NOT gated by `assertVedaVaultUsable` (a quote moves nothing, and a read consuming a money-in gate is the ADR 0002 pattern). Blocking conditions come back as data (`blockingIssues`) in the vault's own words, not as thrown errors. `POST /v1/earn/vault-deposit-previews` serves it behind the deposit's own money-in gates — surfacing, entitlement, admission, environment — with no wallet, no policy gate, no idempotency key, and answers 501 for a provider that cannot quote (measured against the real Kamino client in tests, not staged).

The dashboard derives `minSharesOut = quotedShares × (1 − toleranceBps/10⁴)` in BigInt at the share mint's own scale, floored, one-atom minimum. Deriving from the deposit AMOUNT was rejected: it is only correct while the share rate happens to be 1:1, and the day yield accrues into the rate it starts refusing honestly-priced deposits. The tolerance now covers exactly what it can — the rate moving between the quote and the transaction landing. An unavailable quote DISABLES the deposit rather than falling back to amount arithmetic; a quote is debounced behind typing and aborted on change.

**Defaulted and disclosed rather than demanded.**
`EARN_PROVIDER_DEPOSIT_SLIPPAGE_FLOOR` (`@sdp/types`) declares per provider that the dashboard derives a floor, and the default tolerance — 10 bps for Veda. Kamino's entry is `null` and its dialog is byte-for-byte unchanged. The control hides behind a quiet disclosure ("Slippage tolerance: 0.1%"); the summary always shows "Expected shares" and "Minimum shares received", because the floor is part of what confirm signs.

**A blown floor gets the caller's words and the control that fixes it.** Simulation failing on the vault's `SlippageExceeded` used to surface as the raw simulation string. The API now recognizes the program's NAMED error in the simulation verdict — never the bare custom-error number, which is every Anchor program's first error code — and answers 400 with
`details.reason: "slippage_exceeded"`. The dialog maps that reason to its own copy, auto-opens the slippage section, and RE-QUOTES, so the retry's floor comes from the rate that refused rather than the one that preceded it.

The derived floor joins the idempotency-key fingerprint because the API's own fingerprint includes it: a retry with a different floor is a CHANGED request that must mint a fresh key, not a replay the server would refuse.


apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-tracking.ts
```diff
@@ -52,8 +52,23 @@ export function vaultDepositRequestFingerprint(input: {
...
   amount: string;
+  /**
+   * The derived share floor, or `null` when the provider takes no derived
+   * floor. In the fingerprint because the API's own idempotency fingerprint
+   * includes it: replaying a held key with a DIFFERENT floor is a changed
+   * request the server refuses, so the client must mint a fresh key for it —
+   * exactly the property that lets "raise the tolerance and retry" work after
+   * a slippage refusal.
+   */
+  minSharesOut: string | null;
```
keeganthomp: The derived floor moves with the live quote, so this fingerprint is no longer reproducible from user input, and that is what the store's cross-reload guarantee rests on. Reload mid-flight, re-enter the same amount, and the fresh quote returns a slightly different `sharesOut` (a `vestingYield` vault re-interpolates its rate on every read), so `claim()` misses the stored entry and mints a NEW key: a second approval request for one intent after a 202, or after a 5xx the second on-chain deposit `answerRetiresIdempotencyKey` is written to prevent (`earn-idempotency-key-store.ts:318-343`).

Suggest fingerprinting the user's `toleranceBps` rather than the derived floor. It is stable across re-quotes, and it still changes on "raise the tolerance and retry", which is the property you actually want here. The alternative is to persist the floor next to the claimed key and reuse it instead of re-deriving.

apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx
```diff
@@ -498,6 +760,14 @@ export function EarnVaultDepositModal({
   const { wallets, error: walletsError, isLoading: walletsLoading } = useEarnFundingWallets();
   const [walletId, setWalletId] = useState<string | null>(null);
   const [amountInput, setAmountInput] = useState("");
+  // Declared per provider in @sdp/types: non-null means this provider REQUIRES
+  // an explicit share floor and its shares are pegged 1:1, so the floor may be
+  // derived from the deposit amount. Null renders no slippage control at all.
```
keeganthomp: This says the opposite of what the PR does: `@sdp/types` explicitly documents "Never from the deposit amount", and nothing here is pegged 1:1. Reads like a leftover from the amount-derived draft.

```suggestion
  // Declared per provider in @sdp/types: non-null means this provider REQUIRES
  // an explicit share floor, which the dashboard derives from a LIVE quote and
  // never from the deposit amount. Null renders no slippage control at all.
```

apps/sdp-api/src/routes/earn/index.ts
```diff
@@ -89,6 +91,16 @@ earn.post(
   }),
   createEarnVaultDeposit
 );
+// The deposit QUOTE: a read carrying the deposit's own money-in gates (it
+// exists only to open a new position) but no policy gate, no wallet and no
+// idempotency key — it moves nothing. POST because the parameters are a body,
+// exactly like the custodial withdrawal-preview.
+earn.post(
```
keeganthomp, line 152: `routes/earn/CLAUDE.md` carries an exhaustive route map (every one of the other 15 routes has an entry with its gate order) and this one is missing from it.

Worth a line in "Gate asymmetry" too: it currently files "previews" under money-out/live-reads taking `assertEarnProviderConfigured` ONLY, so as written this route looks like a violation of it rather than the deliberate money-in-shaped read it is.

apps/sdp-api/src/routes/earn/handlers/vault.ts
```diff
@@ -90,6 +95,88 @@ import { parseParams, parseQuery } from "./shared";
...
+  let quote: EarnVaultDepositQuote;
+  try {
+    quote = await client.quoteVaultDeposit(earnRuntime(c), {
+      providerReference: strategy.provider_reference,
+      amount: body.amount,
+    });
+  } catch (error) {
+    // An unusable amount is the CALLER's, in the provider's own words —
+    // same mapping the deposit build applies. Everything else keeps bubbling.
+    if (error instanceof SdpVedaError && error.code === "INVALID_AMOUNT") {
```
keeganthomp: `vault-deposit.service.ts:144` maps these on the shared `code` shape across `SdpKaminoError | SdpVedaError` on purpose, "so a new vault-direct provider inherits the mapping by using the same vocabulary", and this is the first `@sdp/veda` import to reach the routes layer. As written, the next quote-capable provider 500s on a caller-fixable amount here instead of 400ing. Reusing that same code-shape match (exported, or lifted somewhere both can see) keeps the two paths from drifting.

apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx
```diff
@@ -76,6 +78,248 @@ function atomsToDecimalString(atoms: bigint, decimals: number): string {
...
+        onClick={onToggle}
+        type="button"
+      >
+        <ChevronDownIcon
+          aria-hidden="true"
+          className={cn("size-3.5 transition-transform", open && "rotate-180")}
+        />
+        {t("DashboardEarn.deposit.vaultSlippageToggle", { percent })}
+      </button>
+      {open ? (
```
keeganthomp: nit: type an invalid tolerance, then collapse the section, and confirm sits disabled with the reason hidden. The collapsed toggle's placeholder percent is the only hint.

```suggestion
      {open || invalid ? (
```

keeganthomp (changes requested):
Nice change, and deriving the floor from a live quote instead of the amount is clearly the right call. One blocker: putting the quote-derived floor into the CLIENT idempotency fingerprint makes that fingerprint non-reproducible from user input, so a reload-and-resubmit mints a fresh key instead of replaying the held one (apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-tracking.ts:63).

The rest is non-blocking: a new comment that asserts the opposite of the invariant this PR establishes, the missing route-map entry in `routes/earn/CLAUDE.md`, and a provider-specific error catch in a provider-agnostic handler. Separately, outside the diff: the production-floor comment at [handlers/vault.ts#L331](https://github.com/solana-foundation/solana-developer-platform/blob/fd7a547442b10347d016973a9f39288b9569c2b5/apps/sdp-api/src/routes/earn/handlers/vault.ts#L331) still says "The dashboard does not yet quote one (that needs a live rate with a displayed tolerance and an expiry)", which this PR falsifies for the first two thirds (the expiry is still missing).

keeganthomp:
@greptile-apps re-review

keeganthomp (approved):

PR #1449, opened 2026-08-21
feat(earn): Veda instant withdrawals, with quote-derived exit floors

cutbow7:
The instant redemption from ADR 0003's "instant lands first, and alone": burn shares, receive the vault asset, one transaction — carried by the movement model `POST /v1/earn/vault-withdrawals` already runs for Kamino (PRO-1702). Veda becomes the second `supportsVaultWithdraw` provider; the QUEUED exit stays deliberately unimplemented (its lifecycle is settled by a solver Veda operates and does not fit the movement model — ADR 0003 §4 waits on the product answer). Implementing only the instant half is not auto-selecting a route: the caller asked for an immediate redemption and gets exactly that or a typed refusal; SDP never silently substitutes the queue.

**The exit build inherits NO money-in gate.** `buildVedaWithdrawPlan` deliberately skips `assertVedaVaultUsable` — that check demands the withdrawal QUEUE and exists to stop money going IN; demanding it on the way OUT is the inversion ADR 0002 forbids. The vault's own refusals (`RESTRICTED_REDEMPTION` when a withdraw authority is set, `SHARE_LOCKED` inside the post-deposit lock window) surface as a new `WITHDRAW_REFUSED` code with the SDK's own sentence, mapped by the shared `refusedBuildMessage` — extracted from the deposit service into `vault-refusals.ts` so the two directions cannot drift.

**The exit floor mirrors the deposit floor, end to end.** Veda's SDK refuses an implicit tolerance on the way out exactly as on the way in, so `EarnVaultWithdrawInput` gains `minAmountOut`, refused-if-absent by Veda's builder with a typed INVALID_AMOUNT, cross-checked against the plan's encoded `accepted.minAmountOut` in both directions of disagreement, and joined to the withdrawal idempotency fingerprint (a retry with a different floor is a changed request that must mint a fresh key). The floor derives from a live quote: `quoteVaultWithdrawal` (`supportsVaultWithdrawQuote`, over the SDK's `previewWithdraw`) served by `POST /v1/earn/vault-withdrawal-previews` — which carries EXIT gates only: position scoping and the read-side wallet binding, both 404, capability 501, and nothing money-in-shaped. The deposit preview keeps its deliberately DIFFERENT gate set; the asymmetry is the point.

**One copy of the slippage machinery.** The dashboard's floor math, tolerance parsing, refusal detection, debounced quote hook and disclosure section move to `earn-vault-slippage.tsx`, consumed by both vault modals — two copies of a funds-protection rule is how one drifts (the `earn-idempotency-key-store` precedent). The withdraw modal now quotes the live payout, shows "Expected amount" and "Minimum amount received", hides the tolerance behind the same disclosure (10 bps default, `EARN_PROVIDER_WITHDRAW_SLIPPAGE_FLOOR`), disables the exit when the quote is unavailable rather than guessing, and answers a blown floor with its own copy, the control open, and a FRESH quote. Kamino's dialog and contract are byte-for-byte unchanged (`null` policy, no floor).


apps/sdp-api/src/routes/earn/index.ts
```diff
@@ -137,6 +139,16 @@ earn.post(
   }),
   createEarnVaultWithdrawal
 );
+// The exit QUOTE: a read with EXIT gates only — position scoping and the
+// read-side wallet binding (both 404), capability (501), and deliberately
+// nothing money-in-shaped (ADR 0002 exit safety): no surfacing, no
+// entitlement, no admission, no environment capability.
+earn.post(
+  "/vault-withdrawal-previews",
+  requirePermissions("earn:read"),
```
keeganthomp: Every sibling position/wallet read carries `wallets:read` (`GET /vault-positions`, `GET /vault-deposits`, `GET /vault-withdrawals`), and `POST /vault-withdrawals` itself takes it. Dropping it here is load-bearing because of the no-op documented at the top of this file: for a key with no wallet bindings `getAllowedApiKeyWalletIdsForPermissions` returns `null`, so `listReadableEarnVaultWallets` hands back every org wallet and the handler's binding check passes for any position id in the org. Net effect: an `earn:read`-only key learns a position's live redeemable value here while `GET /vault-positions` answers it 403. `wallets:read` is not a money-in gate, so ADR 0002 exit safety does not argue for dropping it.

```suggestion
  requirePermissions("earn:read", "wallets:read"),
```

Worth a spec too: the preview tests cover missing-position 404 and Kamino 501, but not the wallet-binding 404 this gate backs.

packages/sdp-veda/src/sdk.ts
```diff
@@ -358,6 +361,140 @@ export async function previewVedaDeposit(
...
+  } catch (cause) {
+    throw mapVedaSdkError(
+      cause,
+      `Veda could not quote a withdrawal for vault ${input.vault}`,
+      "WITHDRAW_REFUSED"
+    );
+  }
+
+  const assetDecimals =
+    typeof quote.assetDecimals === "number" ? quote.assetDecimals : asset.decimals;
```
keeganthomp: This prefers the SDK's `assetDecimals` over `asset.decimals`, which you already read from the mint account a few lines up, and `typeof === "number"` lets anything through (including past Solana's 9 cap that `mintDecimals` rejects). It matters because `assetsOut` is formatted at this scale and the dashboard derives `minAmountOut` at the same scale, while `buildVedaWithdrawPlan` quantizes that floor against `asset.decimals`: a disagreement rescales the floor by 10^n, which is exactly the "reads as protection everywhere and imposes none on chain" case the package's CLAUDE.md names. The deposit twin at least routes `quote.shareDecimals` through `shareMintDecimals`, which validates and throws.

```suggestion
  // The mint account's own decimals decide the scale: the floor derived from
  // `assetsOut` is quantized against `asset.decimals` in the build, so trusting
  // a different number here would silently rescale it.
  const assetDecimals = asset.decimals;
```

Non-blocking, but cheap to close.

apps/sdp-api/src/services/earn/vault-withdraw.service.ts
```diff
@@ -61,7 +63,7 @@ export interface VaultWithdrawalExecutionOptions {
 
 function requireAcceptedWithdrawalPlan(
   plan: EarnVaultTransactionPlan,
-  input: Pick<VaultWithdrawalInput, "tokenMint" | "shareMint" | "shares">
+  input: Pick<VaultWithdrawalInput, "tokenMint" | "shareMint" | "shares" | "minAmountOut">
 ): void {
```
keeganthomp, line 78: The deposit's twin returns `{ minSharesOut }` so the service can persist it as `acceptedMinSharesOut` (`min_shares_out` on `earn_movements`); this one returns `void`, and `createSignedVaultWithdrawalIntent` never sees the floor, so a landed instant exit leaves no queryable record of what floor was actually encoded. `min_shares_out` is documented as SHARE units, so it cannot be reused for a token-unit floor, which makes this a migration rather than a one-liner. Deliberate deferral, or worth a follow-up? Non-blocking either way, since the signed transaction bytes still carry it.

apps/sdp-api/src/routes/earn/handlers/vault.ts
```diff
@@ -1410,6 +1416,72 @@ export async function findEarnVaultWithdrawalIdempotentKeyReplay(
  * custody wallet that holds it. Build, simulate, sign, record, then broadcast;
  * the reconciliation sweep finishes an ambiguous send.
  */
+/**
```
keeganthomp, line 1292: nit: the `POST /v1/earn/vault-withdrawals` docblock immediately above belongs to `createEarnVaultWithdrawal`, so it now reads as this function's docs and `createEarnVaultWithdrawal` (line 1485) is left with none. Moving those four lines down to sit on 1485 fixes both.

apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-deposit-modal.tsx
```diff
@@ -818,11 +635,14 @@ export function EarnVaultDepositModal({
...
     slippagePolicy !== null && amountValidation.kind === "valid"
       ? amountValidation.canonicalAmount
-      : null,
+      : null;
+  const quote = useDebouncedVaultQuote<EarnVaultDepositPreview>(
+    quoteKey,
+    (signal) =>
+      fetchEarnVaultDepositPreview({ strategyId: strategy.id, amount: quoteKey ?? "" }, signal),
     quoteRefreshKey
   );
```
keeganthomp: nit: the old `useVaultDepositQuote` keyed its effect on `[strategyId, canonicalAmount, refreshKey]`; the new hook keys on `[key, refreshKey]` and this call site passes only the amount, so a `strategy` swap at the same amount would keep serving the previous strategy's quote. Not reachable today (the modal blocks interaction underneath), but the hook's own doc says `key` is the serialized quote input, so worth matching:

```suggestion
  const quoteAmount =
    slippagePolicy !== null && amountValidation.kind === "valid"
      ? amountValidation.canonicalAmount
      : null;
  const quote = useDebouncedVaultQuote<EarnVaultDepositPreview>(
    quoteAmount === null ? null : JSON.stringify([strategy.id, quoteAmount]),
    (signal) =>
      fetchEarnVaultDepositPreview({ strategyId: strategy.id, amount: quoteAmount ?? "" }, signal),
    quoteRefreshKey
  );
```

keeganthomp (changes requested):
Solid work, and the exit-safety reasoning holds up: the four-quadrant `minAmountOut` cross-check, the shared `refusedBuildMessage`, and pulling the floor into both idempotency fingerprints all look right, and the slippage extraction is faithful.

One blocker: `POST /vault-withdrawal-previews` registers with `earn:read` alone (apps/sdp-api/src/routes/earn/index.ts:148) while every sibling position read carries `wallets:read`, and the handler's binding check is a documented no-op for a key with no wallet bindings, so an `earn:read`-only key can read any org position's live payout through the new route.

The rest is non-blocking: the Veda exit quote trusting the SDK's `assetDecimals` over the mint-account read it already has (packages/sdp-veda/src/sdk.ts:486), the withdrawal floor not being persisted the way the deposit's is, and two small refactor leftovers.

keeganthomp (dismissed):

keeganthomp (approved):

PR #1452, opened 2026-08-22
chore(solana): upgrade @solana/kit to 6.10.0

multipletwigs:
- Upgrade `@solana/kit` (and the addresses/codecs/instructions/keys/programs/rpc/signers/sysvars/transactions family) from a stale mixed 6.8–6.10 baseline to a unified 6.10.0 in `pnpm-workspace.yaml`, the actual catalog pnpm reads
- Delete a dead `overrides` block in `package.json` pinning `@solana/kit`/`@solana/rpc-spec` to 6.5.0 — pnpm only reads `pnpm.overrides`, not a bare top-level `overrides` key, so it never applied
- Delete `packages/kit-augment`, a shim backfilling the `ExtendedClient` type that kit 6.9+ already ships natively; drop its two import sites and its `check-module-boundaries.mjs` entries, and regenerate `docs/architecture/module-map.md`
- Fix a real type break in `packages/sdp-solana/src/token-2022.ts` and `packages/sdp-issuance/src/mosaic/service.ts`: both cross-version-bridge Mosaic SDK's kit-v5-typed RPC interface into our client via a hand-rolled `Rpc<SolanaRpcApi> & MosaicSdkRpc` intersection; swapped both to the repo's existing `SolanaRpcSdkBridge<MosaicSdkRpc>` type (same thing `createRpcForSdk` already returns) — no behavior change, just accurate typing


multipletwigs:
superseeded by dependabots!

PR #1454, opened 2026-08-23
fix(payments): persist signed transfers before broadcast

pashpashkin:
## What changed

- Replaced single-transfer `signAndSend` with a durable sign → persist → broadcast sequence.
- Added signed transaction, block-height expiry, and submission-start fields to `payment_transfers`.
- Retries transient RPC failures using the same signed bytes and verifies the returned signature.
- Keeps ambiguous outcomes `processing` for reconciliation; definite failures still become `failed`.
- Extends pending-transfer reconciliation with block-height and archival-history checks.
- Coordinates transfer persistence with Kora sponsorship accounting.

## Why

Previously, an RPC timeout could happen after broadcast but before the signature was recorded. The transfer could then be marked `failed` even though it might still land on-chain, making a replacement payment unsafe.

## Impact

**Before**

1. Create a `processing` transfer.
2. Sign and broadcast through `signAndSend`.
3. Persist the signature only after submission returns.
4. A timeout or aged cache miss could incorrectly fail an already-broadcast transfer.

**After**

1. Fully sign and derive the deterministic signature.
2. Persist the signed bytes, signature, and block-height expiry.
3. Persist `submission_started_at`.
4. Broadcast the exact persisted transaction.
5. Settle as confirmed/finalized, failed only on a definite verdict, or keep `processing` for reconciliation.

```text
processing
  → sign and derive signature
  → persist signed bytes, signature, and expiry
  → mark submission started
  → broadcast the persisted bytes
  ├─ confirmed/finalized → settle
  ├─ definite failure → failed
  └─ ambiguous outcome → stay processing → reconciliation
```

## Pre/post release actions

**Pre-release**

1. Apply `0066_payment_transfer_submission_outbox.sql` before deploying the API and cron. No backfill is required.
2. Deploy the API and pending-transfer cron from the same revision.

**Post-release**

1. Run low-value SOL and SPL transfer smokes.
2. Wait several reconciliation ticks and monitor repeated `sdp_api_payment_submission_unresolved` events by transfer and signature.

**Rollback**

Do not start a pre-change pending-transfer cron while this query returns rows:

```sql
SELECT id, type, signature, submission_started_at, updated_at
FROM payment_transfers
WHERE status = 'processing'
  AND signed_transaction IS NOT NULL
  AND submission_started_at IS NOT NULL;
```

Keep the new cron and reconcile or fix forward.

## Result Validation

- `pnpm -C apps/sdp-api exec vitest run src/services/payments/signed-submission.test.ts src/routes/payments.transfers.test.ts src/services/jobs/track-pending-transfers.test.ts src/services/sponsorship-budget.service.node.test.ts src/services/sponsorship.service.node.test.ts` — 131 passed
- `pnpm --filter @sdp/api typecheck` — passed
- `pnpm --filter @sdp/api lint` — passed
- `pnpm check:api-playground` — passed

Yes, SDP-owned single SOL, SPL, and MagicBlock private transfers.

- The successful response shape is unchanged.
- An ambiguous broadcast or confirmation now returns `200` with the transfer still `processing` and its known signature.
- Definite preflight and on-chain failures still fail normally.
- Batch and recurring flows are handled by the stacked PR.

## Known gaps

- A process crash after `submission_started_at` but before the first RPC call can leave the transfer `processing` for manual reconciliation. Safe automatic rebroadcast requires an additional durable Kora accounting boundary.
- Structured unresolved events are logs, not configured alerts. Manual monitoring is acceptable for devnet; alerting is needed before mainnet.

## Notes

- `signed_transaction` is internal and is not returned by the public API.
- It contains no private key material, but remains broadcast-capable until its blockhash expires.
- Existing rows retain `NULL` outbox fields and continue through the legacy reconciliation path.
- The consumers PR is stacked on this branch.
- Supersedes #1433. This version replaces the previous submission-outcome boundary with durable signed-transaction persistence before broadcast.


pashpashkin:
@greptileai review pls

apps/sdp-api/src/services/jobs/track-pending-transfers.ts
```diff
@@ -289,63 +433,194 @@ async function syncProcessingTransfersOnChain(
...
     const transfer = processingWithSig[i];
     const status = statuses[i] ?? null;
+    if (status && !(await applyOnChainVerdict(env, repo, transfer, status, nowIso))) {
+      await keepStartedSubmissionForReconciliation(
+        repo,
+        transfer,
+        nowIso,
+        unresolvedReasonForStatus(status)
+      );
+    }
```
G1de0n, line 445: This loop calls keepStartedSubmissionForReconciliation for every cache-hit transfer whose verdict didn't apply, including legacy rows with no signed submission (signed_transaction IS NULL). That's broken for legacy transfers stuck at processed-only confirmation, two ways.

updated_at gets rotated on every tick. Once the signature drops out of the recent-status cache, STUCK_PROCESSING_AFTER_MS restarts counting from that rotation, so expiry on the legacy row gets pushed back a whole extra window. Before this change it just skipped a processed-only status and left the row alone.

It's also spamming sdp_api_payment_submission_unresolved warns with last_valid_block_height: null and submission_started_at: null, which is misleading since these rows never had outbox state.

Suggestion: guard the call with hasSignedSubmission(transfer) so only outbox-backed rows get rotated/reported, and legacy rows keep the old skip behavior.

pashpashkin: @G1de0n 
Good catch - fixed. Legacy transfers now keep the previous behavior when RPC reports them as `processed`, `updated_at` is not changed, and no unresolved-submission warning is emitted.

I kept the existing behavior for actual database write failures. The failed row is moved to the back of the polling queue so it cannot block later transfers. The regression test verifies that legacy `processed` transfers keep their timestamp unchanged and emit no warning.

G1de0n (approved):

PR #1455, opened 2026-08-23
fix(payments): extend durable submission to batches and recurring

pashpashkin:
## What changed

- Routes batch chunks and recurring collections through the durable signed-submission flow.
- Keeps ambiguous submission outcomes `processing`; definite failures settle the associated bookkeeping as `failed`.
- Atomically creates and links recurring collection attempts with their transfers.
- Verifies exact on-chain recurring collection evidence before advancing the next due period.
- Adds an SDP-generated random UUID as an on-chain Memo instruction to every batch chunk, ensuring otherwise identical chunks produce distinct signatures.
- Allows cancellation to stop future collections while an already-processing collection settles independently.
- Emits structured events for unresolved submissions and recurring proof mismatches.

## Why

An RPC or confirmation error does not prove that a transaction was never broadcast. Treating it as a definite failure could expose a payment as retryable even though funds may have moved.

Recurring attempt and transfer records also need to remain consistent, and identical batch chunks must not share the same transaction signature.

## Impact

**Before**

1. Batch and recurring transactions used `signAndSend`.
2. Bookkeeping was completed after submission.
3. An ambiguous exception could incorrectly fail the payment.
4. Recurring cancellation waited for an existing collection to resolve.

**After**

1. Create and atomically link the tracking records.
2. Persist the signed payload, signature, and expiry before broadcast.
3. Keep uncertain outcomes `processing` until reconciliation.
4. Advance the recurring due period only after exact on-chain proof.
5. Stop future collections on cancellation without changing the outcome of one already in progress.

```text
batch/recurring processing
  batch chunk / recurring collection
  → build transaction
  → add on-chain UUID Memo (batch only)
  → sign and persist bytes + signature
  → broadcast the persisted bytes
  ├─ confirmed → settle bookkeeping
  ├─ definite failure → failed
  └─ ambiguous outcome → stay processing → reconciliation
```

## Pre/post release actions

**Pre-release**

1. Merge the signed-transfer submission PR first, then retarget this PR to `main`.
2. Require green CI on the resulting stack.

**Post-release**

1. Submit a low-value multi-chunk batch with identical amounts and verify distinct signatures and final recipient states.
2. Exercise recurring collect and cancel while a collection is processing, then wait several reconciliation ticks.
3. Confirm the API and payment jobs use the same revision.
4. Monitor `sdp_api_payment_submission_unresolved` and require zero `sdp_api_recurring_payment_collection_proof_mismatch` events.

**Rollback**

Prefer fix-forward. Do not roll the recurring collection job back while a processing attempt belongs to a `canceling` or `canceled` recurring payment; the previous job does not recover those attempts.

## Result Validation

- `pnpm -C apps/sdp-api exec vitest run src/services/payments/signed-submission.test.ts src/routes/payments.transfers.test.ts src/services/jobs/track-pending-transfers.test.ts src/routes/payments/transfer-batches.test.ts src/routes/payments.recurring.test.ts` — 120 passed
- `pnpm --filter @sdp/api typecheck` — passed
- `pnpm --filter @sdp/api lint` — passed
- `pnpm check:api-playground` — passed
- Full API suite — 3602 passed; 15 tests failed from parallel interference in two Redis suites, which passed 52/52 when rerun serially

Yes, batch submission and recurring collection/cancellation behavior.

- Successful response shapes remain unchanged.
- An uncertain recurring submission can return `200` with its transfer still `processing`.
- An uncertain batch chunk remains `processing` instead of being falsely failed.
- A recurring proof mismatch returns `409`, emits a structured event, and leaves the records unresolved for investigation.
- Cancellation no longer waits for an already-processing collection.
- Every batch chunk now contains a publicly visible on-chain Memo instruction.

## Known gaps

- The crash window after the submission-start marker but before the first RPC call is inherited from the preceding PR; automatic rebroadcast is intentionally not implemented yet.
- Structured events are logs rather than configured alerts.

## Notes

- The batch Memo contains only an SDP-generated random UUID—no customer data—and is not exposed as the transfer's API `memo`.
- Kora must allow the Solana Memo program; the batch integration smoke validates this.
- This PR contains no migration and depends on the preceding signed-outbox PR.
- Supersedes #1434 and depends on the preceding signed-transfer submission PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -503,10 +503,10 @@ jobs:
             test_files: src/tests/payments-wallet-scope.test.ts src/tests/api-keys-flow.test.ts src/tests/api-keys-rotation.test.ts src/tests/custody-local.test.ts
             rpc_provider: alchemy
             custody_provider: local
-          - name: Surfpool / Token Flows (Kora)
+          - name: Surfpool / Token and Payment Flows (Kora)
             provider: surfpool
             kora_mode: docker
-            test_files: src/tests/deploy.test.ts src/tests/mint.test.ts src/tests/burn.test.ts src/tests/freeze.test.ts
+            test_files: src/tests/deploy.test.ts src/tests/mint.test.ts src/tests/burn.test.ts src/tests/freeze.test.ts src/tests/transfer-batches.test.ts
             rpc_provider: alchemy
             custody_provider: local
     env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -599,6 +599,8 @@ jobs:
         env:
           DOPPLER_TOKEN: ${{ secrets.DOPPLER_TOKEN_CI }}
           DOPPLER_CONFIG: dev_ci
+          DOPPLER_PRESERVE_ENV: TEST_DATABASE_URL,REDIS_URL,SOLANA_NETWORK,KORA_SURFPOOL_MODE,COUNTERPARTY_PII_ENCRYPTION_KEY
+          COUNTERPARTY_PII_ENCRYPTION_KEY: BwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwc=
           SDP_INTEGRATION_CUSTODY_PROVIDER: ${{ matrix.custody_provider }}
         run: |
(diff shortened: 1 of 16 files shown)


pashpashkin:
@greptileai review pls

G1de0n (dismissed):

G1de0n (approved):

PR #1470, opened 2026-08-24
feat(earn): add persistent button integration builder

keeganthomp:
- Persists one Earn button configuration per organization and project, including strategy, style, accent color, and a stable public token.
- Rechecks catalogue, environment, provider, access, and deposit availability before saving.
- Adds an unsigned engineering handoff with a server-only integration example and no tenant data or API key.
- Replaces the oversized previews with polished iOS/Web frames, live controls, curated accents, and custom color selection.

**before**: The builder displayed disabled controls, lost changes on reload, and produced no engineering handoff.

**after**:

1. The dashboard loads or saves the scoped configuration.
2. Admission checks reject unavailable strategies with a named reason.
3. The stable public link renders the configured button and server integration without dashboard authentication.

**Verification**

- Typechecks passed for types, API, and web.
- 68 API tests and 13 web tests passed.
- Module boundaries, i18n, Biome, production build, and browser QA passed.


keeganthomp:
Screenshot 2026-08-24 at 3 15 18 PM
Screenshot 2026-08-24 at 3 15 35 PM
Screenshot 2026-08-24 at 3 15 43 PM

apps/sdp-web/src/app/dashboard/markets/earn/earn-button-builder.tsx
```diff
@@ -27,179 +40,582 @@ import { ListEmptyState } from "@/components/ui/list-empty-state";
...
+  configurationLoad,
   earnHref,
   providerAccess,
   strategyId,
 }: {
+  configurationLoad: EarnButtonConfigurationLoad;
   earnHref: string;
   providerAccess: EarnProviderAccess | null;
   strategyId?: string;
 }) {
```
keeganthomp: Addressed in [`558734dd406a`](https://github.com/solana-foundation/solana-developer-platform/commit/558734dd406ae6c8fda046d775ffcedda541811a): Grouped coordinated builder fields and transitions in one reducer; 14 focused web tests and web typecheck pass.

apps/sdp-web/src/app/dashboard/markets/earn/earn-button-builder.tsx
```diff
@@ -27,179 +40,582 @@ import { ListEmptyState } from "@/components/ui/list-empty-state";
...
+  const [shareOrigin, setShareOrigin] = useState("");
+
+  useEffect(() => setShareOrigin(window.location.origin), []);
+  useEffect(() => {
+    const nextConfiguration =
+      configurationLoad.kind === "ready" ? configurationLoad.configuration : null;
+    setSavedConfiguration(nextConfiguration);
+    setStyle(nextConfiguration?.style ?? "ink");
+    setAccentColor(nextConfiguration?.accentColor ?? DEFAULT_EARN_BUTTON_ACCENT_COLOR);
+    setIsSaving(false);
```
keeganthomp: Addressed in [`558734dd406a`](https://github.com/solana-foundation/solana-developer-platform/commit/558734dd406ae6c8fda046d775ffcedda541811a): Removed prop-sync state adjustment and reset the session by configuration identity; the project-scope reset test passes.

apps/sdp-web/src/app/dashboard/markets/earn/earn-button-builder.tsx
```diff
@@ -27,179 +40,631 @@ import { ListEmptyState } from "@/components/ui/list-empty-state";
...
+        {t("DashboardMarkets.earnProgram.previewRate", {
+          apy: formatProviderApy(strategy.currentApy, locale),
+        })}
+      </p>
     </div>
   );
 }
 
-function IosButtonPreview({ strategy, style }: { strategy: EarnStrategy; style: EarnButtonStyle }) {
+function IosButtonPreview({
```
keeganthomp: Addressed in [`31a8e0b2713a`](https://github.com/solana-foundation/solana-developer-platform/commit/31a8e0b2713a155edced27a08cd6aa15c9972826): Moved each builder subcomponent into its own focused file; 14 web tests, typecheck, module boundaries, and the production build pass.

apps/sdp-web/src/app/dashboard/markets/earn/earn-button-builder.tsx
```diff
@@ -27,179 +40,631 @@ import { ListEmptyState } from "@/components/ui/list-empty-state";
...
+            <div aria-hidden="true" className="mx-auto h-1 w-24 rounded-full bg-primary/80" />
           </div>
         </div>
       </div>
     </figure>
   );
 }
 
-function WebButtonPreview({ strategy, style }: { strategy: EarnStrategy; style: EarnButtonStyle }) {
+function WebButtonPreview({
```
keeganthomp: Addressed in [`31a8e0b2713a`](https://github.com/solana-foundation/solana-developer-platform/commit/31a8e0b2713a155edced27a08cd6aa15c9972826): Moved each builder subcomponent into its own focused file; 14 web tests, typecheck, module boundaries, and the production build pass.

apps/sdp-web/src/app/dashboard/markets/earn/earn-button-builder.tsx
```diff
@@ -27,179 +40,631 @@ import { ListEmptyState } from "@/components/ui/list-empty-state";
...
         </div>
       </div>
     </figure>
   );
 }
 
-export function EarnButtonBuilder({
+type PreviewDevice = "ios" | "web";
+
+function DevicePreview({
```
keeganthomp: Addressed in [`31a8e0b2713a`](https://github.com/solana-foundation/solana-developer-platform/commit/31a8e0b2713a155edced27a08cd6aa15c9972826): Moved each builder subcomponent into its own focused file; 14 web tests, typecheck, module boundaries, and the production build pass.

apps/sdp-web/src/app/dashboard/markets/earn/earn-button-builder.tsx
```diff
@@ -27,179 +40,631 @@ import { ListEmptyState } from "@/components/ui/list-empty-state";
...
+          <IosButtonPreview accentColor={accentColor} strategy={strategy} style={style} />
+        ) : (
+          <WebButtonPreview accentColor={accentColor} strategy={strategy} style={style} />
+        )}
+      </div>
+    </section>
+  );
+}
+
+function AppearanceStyleControls({
```
keeganthomp: Addressed in [`31a8e0b2713a`](https://github.com/solana-foundation/solana-developer-platform/commit/31a8e0b2713a155edced27a08cd6aa15c9972826): Moved each builder subcomponent into its own focused file; 14 web tests, typecheck, module boundaries, and the production build pass.

apps/sdp-web/src/app/dashboard/markets/earn/earn-button-builder.tsx
```diff
@@ -27,179 +40,631 @@ import { ListEmptyState } from "@/components/ui/list-empty-state";
...
+              {accentColor}
+            </span>
+          </div>
+        </fieldset>
+      ) : null}
+    </div>
+  );
+}
+
+function EngineeringHandoff({
```
keeganthomp: Addressed in [`31a8e0b2713a`](https://github.com/solana-foundation/solana-developer-platform/commit/31a8e0b2713a155edced27a08cd6aa15c9972826): Moved each builder subcomponent into its own focused file; 14 web tests, typecheck, module boundaries, and the production build pass.

apps/sdp-web/src/app/dashboard/markets/earn/earn-button-builder.tsx
```diff
@@ -27,179 +40,631 @@ import { ListEmptyState } from "@/components/ui/list-empty-state";
...
+      ) : (
+        <p className="mt-4 text-xs font-medium text-secondary">
+          {t("DashboardMarkets.earnProgram.unsavedHandoff")}
+        </p>
+      )}
+    </section>
+  );
+}
+
+function BuilderFooter({
```
keeganthomp: Addressed in [`31a8e0b2713a`](https://github.com/solana-foundation/solana-developer-platform/commit/31a8e0b2713a155edced27a08cd6aa15c9972826): Moved each builder subcomponent into its own focused file; 14 web tests, typecheck, module boundaries, and the production build pass.

apps/sdp-api/src/routes/earn/index.ts
```diff
@@ -55,9 +61,27 @@ async function requireEarnFeature(c: Context<{ Bindings: Env }>, next: Next) {
 }
 
 earn.use("*", requireEarnFeature);
+
+// Public, read-only engineering handoff. Registered before auth intentionally:
+// possession of the unguessable token grants access to strategy/style only,
+// never tenant metadata or an API key.
+earn.get("/button-configurations/public/:publicToken", getPublicEarnButtonConfiguration);
```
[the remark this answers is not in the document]
keeganthomp: Addressed in [`4243e5817b3a`](https://github.com/solana-foundation/solana-developer-platform/commit/4243e5817b3a3c4fb0edceeb1d190b637a330d8d): Registered all three Earn button configuration routes in the internal generated OpenAPI contract, including the unauthenticated handoff response and operational error statuses, while keeping Earn outside the supported public API catalogue. OpenAPI generation, API typecheck/build, public playground drift, and explicit internal/public document assertions pass.

apps/sdp-api/src/openapi/paths/earn.ts
```diff
@@ -0,0 +1,80 @@
+import type { OpenAPIRegistry } from "@asteasolutions/zod-to-openapi";
+
+import {
+  earnButtonConfigurationPublicParamsSchema,
+  earnButtonConfigurationSchema,
+} from "@/routes/earn/schemas";
+import { errorResponseSchema } from "../schemas";
```
keeganthomp: Addressed in [`f10b3a7145c9`](https://github.com/solana-foundation/solana-developer-platform/commit/f10b3a7145c9d11ea7ba644b38e1e1aaa08962fb): Replaced the OpenAPI barrel import with the direct base-schema import. Biome and API typecheck pass.

apps/sdp-api/src/openapi/paths/earn.ts
```diff
@@ -0,0 +1,80 @@
...
+    security: [{ apiKeyAuth: [] }],
+    request: {
+      headers: projectScopeHeaders,
+    },
+    responses: {
+      200: {
+        description: "Earn button configuration",
+        content: jsonContent(earnButtonConfigurationResponse),
+      },
+      ...errorResponses(errorResponseSchema, [401, 403, 404, 500]),
```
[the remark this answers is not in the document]
keeganthomp: Addressed in [`3a53b2afa2c2`](https://github.com/solana-foundation/solana-developer-platform/commit/3a53b2afa2c2766c0bd89165781acd2bab6caeca): Added the reachable 400 project-scope response to the authenticated current-configuration GET and pinned it in the OpenAPI regression test. Biome, API typecheck, and generated-document assertions pass.

apps/sdp-api/src/openapi/paths/earn.ts
```diff
@@ -0,0 +1,80 @@
...
+    security: [{ apiKeyAuth: [] }],
+    request: {
+      headers: projectScopeHeaders,
+    },
+    responses: {
+      200: {
+        description: "Earn button configuration",
+        content: jsonContent(earnButtonConfigurationResponse),
+      },
+      ...errorResponses(errorResponseSchema, [400, 401, 403, 404, 500]),
```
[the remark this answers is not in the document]
keeganthomp: Done in b6d62c8c: added 429 to both authenticated button-configuration operations and pinned it in spec.test.ts.

apps/sdp-api/src/openapi/paths/earn.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  registry.registerPath({
+    method: "get",
+    path: "/v1/earn/button-configurations/current",
+    tags: ["Earn"],
+    summary: "Get the current Earn button configuration",
+    operationId: "getEarnButtonConfiguration",
+    description:
+      "Gets the saved Earn button configuration for the active organization and project.",
+    security: [{ apiKeyAuth: [] }],
```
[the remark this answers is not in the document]
keeganthomp: Addressed in [`5e42bbecfa23`](https://github.com/solana-foundation/solana-developer-platform/commit/5e42bbecfa23c08c09ee7014b9e9c12fbd518d84): Documented API key, Clerk JWT, and session cookie alternatives for both Earn configuration operations; 16 OpenAPI regression tests, API typecheck, and API build pass.

cutbow7 (changes requested):
## Review: `codex/earn-button-builder` → `main`

Solid feature overall — the tenant-scoping tests, the project guard on the PUT proxy, and the public-response minimization are all careful work. But there are several real problems, two of them on the new **unauthenticated** surface.

### Correctness

**1. The public handoff endpoint bypasses every catalogue visibility/admission gate** — `apps/sdp-api/src/routes/earn/handlers/button-configurations.ts:113`

`getPublicEarnButtonConfiguration` resolves the strategy with a bare `repo.getStrategyById(row.strategy_id)`. Every other catalogue read goes through `requireEarnStrategy`, which applies `isHiddenStrategy` (`handlers/strategies.ts:172`) — and `routes/earn/CLAUDE.md` is explicit that "a detail route that drifts from the list route leaks a row by id." Consequences:

- A strategy later added to `HIDDEN_VAULTS` / matching `HIDDEN_STRATEGY_TERMS` keeps leaking its `name` and `provider` through the one route that requires no auth at all.
- A **paused** strategy (the deliberate depeg stop switch) still renders a polished `/earn/integrate/[token]` page whose `depositIntoEarnVault` snippet 400s on every call — the exact "polished dead end" the PUT handler's own comment says the gates exist to prevent.
- A **delisted** strategy (row deleted by the sync — migration 0068 intentionally has no FK) yields `strategyName: null`, and the page falls back to rendering the raw `earn_strategy_<uuid>` as the strategy name while still shipping the snippet.

The read path needs at least the hidden/status checks, and the page needs an honest "this configuration is stale" state.

**2. The public page hard-500s on any non-404, and the endpoint sits in the shared anonymous rate bucket** — `apps/sdp-web/src/app/earn/integrate/[token]/earn-integration-handoff-data.ts:30`

The page fetch runs server-side with no auth, so every view lands in the anonymous limiter (`ANONYMOUS_MAX_REQUESTS = 20`/min, keyed by `getClientIp(c) ?? "unknown"` — and all views proxy through the same web-server egress). A handoff link posted in a partner's team channel hits 429 after ~20 opens in a minute, `loadPublicEarnButtonConfiguration` throws, and Next serves a hard 500 to everyone — and the bucket is shared with the live `/pay/:token` page. The existing precedent, `pay/[token]/page.tsx:76`, degrades any `!response.ok` to `notFound()` instead. The branch's own unit test ("preserves an operational 429 response as an error") locks the crash in.

**3. A configuration-load failure bricks the entire builder** — `earn-button-configuration.server.ts:18`, consumed at `earn-button-builder.tsx:203`

Only `SdpApiResponseError` 404 maps to "nothing saved"; everything else becomes `{kind: "error"}`, and `builderEmptyState` short-circuits on it **before** looking at `?strategy=`. Notably `createSdpApiClient()` throws plain `Error("Selected project required")` (`sdp-api.ts:288`) when no project is selected — that user gets "Button configuration unavailable — Refresh before making changes", which refreshing cannot fix, on a page that on `main` rendered the previews and snippet fine (neither depends on a saved configuration). A transient API 500 causes the same full-page dead end.

**4. An in-flight save clobbers newer edits and reports them as saved** — `earn-button-builder.tsx:96`

Only the Save button is disabled while `isSaving`; the appearance controls stay live. Click Save on "ink", switch to "accent"/purple while the PUT is in flight → `saveSucceeded` resets `style`/`accentColor` from the response, `hasUnsavedChanges` flips false, and the footer says "Configuration saved. The handoff link is current." for a configuration that isn't what the preview showed. Either disable the controls during the save or don't overwrite local selection from the response.

**5. Blocking index build on `projects` inside the transactional migration path** — `0068_earn_button_configurations.sql:17`

`CREATE UNIQUE INDEX ... ON projects(id, organization_id)` (non-concurrent) blocks all writes to `projects` for the duration during deploy. Every other migration indexing a pre-existing hot table (0030, 0031, 0033, 0037, 0039, 0041, 0043, 0045) uses `-- sdp:migration-mode: non-transactional` + `CREATE INDEX CONCURRENTLY`. The index also exists only to legalize the composite FK — every other project-scoped table uses plain `FOREIGN KEY (project_id) REFERENCES projects(id)` with `organization_id` as a column, which needs no new index (both reads already filter on `organization_id`).

**6. An empty `?strategy=` param hides the saved configuration** — `earn-button-builder.tsx:196`

`strategyId ?? savedConfiguration?.strategyId` — `page.tsx` passes `""` through (`typeof "" === "string"`), `??` keeps it, and the falsy check downstream routes to "Choose a strategy first" while a saved strategy, style, and live handoff link exist. Use `||`, or normalize empty to `undefined` in `page.tsx`.

**7. Case-sensitive preset matching vs. lowercase hex the API accepts** — `earn-button-appearance-controls.tsx:24`

`EARN_BUTTON_ACCENT_COLOR_PATTERN` is `/^#[0-9A-F]{6}$/i` and the DB CHECK accepts either case, but `isPresetColor` compares `===` against uppercase presets. A config saved via the API with `#9945ff` renders the Purple swatch unselected (custom swatch gets the ring), and clicking Purple flips `hasUnsavedChanges` for a visually identical color. Normalize on load, not just in the color-input handler.

### Design / cleanup

**8. The money-in admission chain is now copy-pasted into a third handler** — `button-configurations.ts:60–92`

The PUT re-implements, inline and in order, the same six-gate sequence `POST /vault-deposits` runs in `handlers/vault.ts`, and `handlers/admission.ts` documents exactly why that's dangerous ("a second copy is a second thing that can drift toward permissive"). When PRO-1703 opens production (or a gate is added) in one place only, the builder saves configs deposits refuse — or vice versa. Extract one `assertVaultDepositAdmissible(c, strategyId)` both handlers call.

**9. Validation shapes duplicated instead of shared** — `openapi/schemas/earn.ts:11` and three siblings

The OpenAPI schema re-inlines `/^#[0-9A-Fa-f]{6}$/` instead of importing `EARN_BUTTON_ACCENT_COLOR_PATTERN` (it already imports from `@sdp/types` on line 1), and the public-token shape `.length(24).regex(/^[A-Za-z0-9_-]+$/)` is hand-copied in the route schema, the OpenAPI schema, and the web client while the generator (`customAlphabet(..., 24)`) lives in the repository. This PR proved the right pattern with the accent constant — a `EARN_BUTTON_PUBLIC_TOKEN_LENGTH`/`_PATTERN` in `@sdp/types` closes the drift.

**10. Dead `GET` proxy route** — `app/api/dashboard/markets/earn/button-configuration/route.ts:6`

Nothing calls it — the builder loads config server-side via `createSdpApiClient()`, and the only client call is the PUT. Its unit test exercises only itself, and unlike the PUT it lacks the `expectedProjectId` guard, so a future adopter inherits the stale-project-cookie hazard the PUT was carefully guarded against. Drop it (or guard it and use it).

### Minor notes (no action required to merge)

- `earn-button-builder.tsx:111` guards style-option exhaustiveness with a module-load `throw` (a runtime crash of the whole route segment) where `satisfies Record<EarnButtonStyle, ...>` gives the same guarantee at compile time — and would let the `EARN_BUTTON_STYLES` re-export shim in `earn-button-preview.tsx` be deleted.
- Migrations 0068+0069 land in the same PR; nothing is deployed, so the accent column (and its `DROP CONSTRAINT IF EXISTS` dance for a state that can't exist) could be folded into 0068's `CREATE TABLE`.
- `EarnButtonConfigurationLoad` is declared twice (`earn-button-configuration.server.ts` exports it; the builder re-declares it) — a type-only import is erased and won't drag `server-only` into the client bundle.
- The module's own CLAUDE.md says "Selection state is `border-primary bg-fill-subtle` across the whole module" and "truncates with a `title` carrying the full string" — the new segmented pills/swatch rings and the title-less truncating handoff link both deviate. Relatedly, `dashboard/issuance/create/segmented-control.tsx` already exists (with `useId()` radio names and focus-visible styling) and could replace both hand-rolled segmented controls.
- The handoff page resolves `apiBaseUrl` for its own fetch but the generated snippet still hardcodes `DEFAULT_SDP_API_URL` — worth threading through, given the feature is sandbox-only.
- Save errors interpolate raw English strings (`"Selected project required"`, API messages) into the localized `saveError` template — mixed-language output in non-`en` locales.
- The retargeted builder test dropped the last coverage of the "Choose a strategy first" branch (no `strategyId`, no saved config).

---

Findings **1, 2, and 3** are the ones I'd block on; the rest are fix-before-merge or follow-up material.

keeganthomp:
All ten findings addressed in 34e41565f.

| # | Fix |
| --- | --- |
| 1 | Public read now applies `isHiddenStrategy` + status; response gains `strategyAvailable`, name/provider withheld when false, and the handoff page renders a stale notice instead of the snippet. Route tests cover hidden, paused, and delisted rows. |
| 2 | Any non-OK public read degrades to `notFound()` (pay-page precedent); the test that pinned the 429 crash now pins the degrade. |
| 3 | A failed config load only dead-ends when no strategy is selected either; with `?strategy=` the builder renders with an inline warning. |
| 4 | The save response no longer resets local style/accent, so an edit made mid-save survives and `hasUnsavedChanges` keeps the footer honest. |
| 5 | 0068 rewritten with a plain `project_id` FK, no unique index on `projects`; 0069 folded in since nothing is deployed. |
| 6 | Empty `?strategy=` normalized away in page.tsx, and the builder uses `\|\|`. |
| 7 | Saved accent colors are uppercased on entry, so `#9945ff` selects the Purple swatch. |
| 8 | Chain extracted to `assertVaultDepositAdmissible` in handlers/admission.ts; both the deposit route and the builder PUT call it (strategy resolution stays per-caller: bare-by-id for money, `requireEarnStrategy` for new configs). |
| 9 | `EARN_BUTTON_PUBLIC_TOKEN_LENGTH`/`_PATTERN` added to @sdp/types and consumed by the generator, route params, OpenAPI, and the web client; OpenAPI now imports the accent pattern too. |
| 10 | GET proxy route dropped. |

From the minor notes: took the `satisfies` exhaustiveness check (the `EARN_BUTTON_STYLES` re-export shim is gone), the type-only `EarnButtonConfigurationLoad` import, the snippet now embeds the resolved `apiBaseUrl` on the handoff page, the handoff link carries a `title`, and the "Choose a strategy first" test is restored.

Left as follow-ups: reusing `segmented-control.tsx` / the selection-state convention (visual refactor), and localizing save errors (needs an error-mapping layer rather than string interpolation).

cutbow7 (approved):

PR #1485, opened 2026-08-25
feat(earn): catalogue mainnet strategies in sandbox behind a cluster toggle (PRO-1742)

keeganthomp:
Lets sandbox browse the curated mainnet Earn catalogue so shelf curation is reviewable outside prod ([PRO-1742](https://linear.app/solana-fndn/issue/PRO-1742/catalogue-mainnet-strategies-in-sandbox-behind-a-cluster-toggle)). The sync mirrors production's accepted mainnet snapshots into non-production environments as a browse-only sub-shelf, `GET /v1/earn/strategies` defaults to the environment's own cluster with an explicit `?cluster=` opt-in, and the Treasury strategies card gets a sandbox-only Devnet/Mainnet toggle whose non-fundable rows badge "{cluster} only" with Deposit disabled.

## Reviewer notes

- **Delists are cluster-scoped lanes in non-prod** (env-wide in prod, unchanged): each fetch converges only the sub-shelf it read, so a mirror-source failure can never tear down the devnet shelf, and vice versa.
- **Deploy window:** old instances' hourly sync will delist freshly mirrored rows (env-wide delist, devnet keep set) until the roll completes, then the next new-code pass restores them. Browse-only rows, so harmless.
- **Mirror collisions:** the upsert key is (provider, reference, environment) with no cluster, so a reference both sources list would flip clusters every pass. The environment's own shelf wins; the mirror row is dropped with a warn.
- **Wire behavior change:** default lists now exclude wrong-cluster rows (previously listed with `fundable: false`). The detail read deliberately still serves an addressed row whatever its cluster.
- **Curation re-keyed environment to cluster** (`HIDDEN_VAULTS`/`CURATED_VAULTS`, both empty today) so the sandbox mirror inherits exactly the curation production applies.
- **The deposit refusal needed no new code:** `assertStrategyDepositable` already 400s naming the cluster; the pre-existing route test covers it.
- `ui-copy-baseline.json` is a wholesale `check:i18n:baseline` regen; it also prunes stale entries main had accumulated.

## Verification

161 API tests across the five earn suites and 1339 sdp-web unit tests pass locally (the web suite is not in CI). Live browser pass on the local stack: devnet default, toggle reveals the real 23-vault mainnet shelf badged "Mainnet only" with Deposit disabled, allocation summary unaffected.

CLAUDE.md updates ride along (routes/earn, sdp-earn, web earn module) plus the stale "sync refuses mainnet outside production" docstrings.


apps/sdp-api/src/cron/earn-catalogue-sync.ts
```diff
@@ -96,18 +104,166 @@ export async function syncEarnCatalogue(env: Env): Promise<void> {
...
+      continue;
+    }
+    mirrorRows.push(snapshot);
+  }
+
+  await writeCatalogueLane(repo, client, {
+    environment,
+    snapshots: mirrorRows,
+    delistScope: "mainnet-beta",
+  });
```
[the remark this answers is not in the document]
keeganthomp: Real, and one step sharper than described: the stored row only survives as mainnet-beta because the own-lane upsert failed, which is exactly the write failure the keep set exists to protect. Fixed in edfbe4b2: collided references now ride in the mirror's keep set (never its upsert set), so the mirror delist can't turn a failed own-lane write into a delisting. Pinned by a new sync test.

apps/sdp-api/src/cron/earn-catalogue-sync.ts
```diff
@@ -96,18 +104,166 @@ export async function syncEarnCatalogue(env: Env): Promise<void> {
   // environment — same contract as the route-layer earnRuntime().
   const providerEnv = env;
 
+  for (const client of Object.values(EARN_PROVIDER_CLIENTS)) {
+    await syncProviderCatalogue(repo, client, providerEnv);
```
keeganthomp, line 108: Deliberately sequential: providers sync one at a time so failure handling and log context stay isolated per provider, same as before this PR. An hourly job whose pass completes in seconds gains nothing from fan-out, so not applying.

apps/sdp-api/src/cron/earn-catalogue-sync.ts
```diff
@@ -96,18 +104,166 @@ export async function syncEarnCatalogue(env: Env): Promise<void> {
...
+    });
+  }
+
   for (const environment of SYNCED_ENVIRONMENTS) {
-    for (const client of Object.values(EARN_PROVIDER_CLIENTS)) {
-      await syncProviderCatalogue(repo, client, { env: providerEnv, environment });
+    if (environment === "production") {
+      continue;
     }
+    await syncNonProductionEnvironment(repo, client, env, environment, production);
```
keeganthomp, line 141: Deliberately sequential: this loop has exactly one non-production member today, and each environment's pass consumes the shared production fetch already in hand. Nothing to parallelize, so not applying.

apps/sdp-api/src/cron/earn-catalogue-sync.ts
```diff
@@ -96,18 +104,173 @@ export async function syncEarnCatalogue(env: Env): Promise<void> {
   // environment — same contract as the route-layer earnRuntime().
   const providerEnv = env;
 
+  for (const client of Object.values(EARN_PROVIDER_CLIENTS)) {
+    await syncProviderCatalogue(repo, client, providerEnv);
```
keeganthomp, line 108: Re-post of the earlier warning on this line; the sequential provider loop is deliberate (per-provider failure isolation, hourly job, seconds-long pass). See the prior thread for the full rationale.

apps/sdp-api/src/cron/earn-catalogue-sync.ts
```diff
@@ -96,18 +104,173 @@ export async function syncEarnCatalogue(env: Env): Promise<void> {
...
+    });
+  }
+
   for (const environment of SYNCED_ENVIRONMENTS) {
-    for (const client of Object.values(EARN_PROVIDER_CLIENTS)) {
-      await syncProviderCatalogue(repo, client, { env: providerEnv, environment });
+    if (environment === "production") {
+      continue;
     }
+    await syncNonProductionEnvironment(repo, client, env, environment, production);
```
keeganthomp, line 141: Re-post of the earlier warning on this line; the loop has one non-production member and consumes the shared production fetch, so there is nothing to parallelize. See the prior thread for the full rationale.

apps/sdp-api/src/cron/earn-catalogue-sync.ts
```diff
@@ -96,18 +104,173 @@ export async function syncEarnCatalogue(env: Env): Promise<void> {
...
+      continue;
+    }
+    ownRows.push(snapshot);
+  }
+
+  await writeCatalogueLane(repo, client, {
+    environment,
+    snapshots: ownRows,
+    delistScope: environmentCluster,
+  });
```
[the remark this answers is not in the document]
keeganthomp: The sequence is accurate, but the implied fix inverts the file's safety asymmetry: protecting the devnet row from its own truthful delist would keep a vault the provider DELISTED sitting fundable: true for an hour, while the flagged outcome is an hour's absence of a browse-only mirror row that self-heals on the next pass. Absence is the cheaper failure, so this stays as designed; 8b88fc79 records the decision in the lane doc comment so it reads as chosen, not missed.

apps/sdp-api/src/cron/earn-catalogue-sync.ts
keeganthomp, line 108: Re-post on push; the provider loop stays deliberately sequential (per-provider failure isolation, hourly job, seconds-long pass). Rationale on the first thread for this line.

apps/sdp-api/src/cron/earn-catalogue-sync.ts
keeganthomp, line 141: Re-post on push; one non-production environment and a shared production fetch leave nothing to parallelize here. Rationale on the first thread for this line.

apps/sdp-api/src/cron/earn-catalogue-sync.ts
```diff
@@ -96,18 +104,182 @@ export async function syncEarnCatalogue(env: Env): Promise<void> {
   // environment — same contract as the route-layer earnRuntime().
   const providerEnv = env;
 
+  for (const client of Object.values(EARN_PROVIDER_CLIENTS)) {
+    await syncProviderCatalogue(repo, client, providerEnv);
```
keeganthomp, line 108: Fixed the recurrence in 82ba17df1: react-doctor's workflow now scopes to the React apps (project: sdp-web,sdp-docs), so React rules stop walking @sdp/api cron code. The sequential awaits here stay deliberate, rationale on the earlier threads.

apps/sdp-api/src/cron/earn-catalogue-sync.ts
```diff
@@ -96,18 +104,182 @@ export async function syncEarnCatalogue(env: Env): Promise<void> {
...
+    });
+  }
+
   for (const environment of SYNCED_ENVIRONMENTS) {
-    for (const client of Object.values(EARN_PROVIDER_CLIENTS)) {
-      await syncProviderCatalogue(repo, client, { env: providerEnv, environment });
+    if (environment === "production") {
+      continue;
     }
+    await syncNonProductionEnvironment(repo, client, env, environment, production);
```
keeganthomp, line 141: Same fix, 82ba17df1: the workflow is scoped to the React apps now, so this stops re-posting. Loop stays deliberately sequential (one non-production environment, shared production fetch).

cutbow7 (changes requested):
## Review: PRO-1742 — catalogue mainnet strategies in sandbox behind a cluster toggle

Overall this is a carefully built change — the lane/keep-set design, the collision handling, the NULL-`host_cluster` COALESCE rule, and the test coverage are all thorough, and several things I chased (bind ordering in the repository, the `testnet` rejection, cluster label/env maps, the collided-reference keep-set corner) check out. The findings below are the gaps that survived adversarial verification, roughly by severity.

### Correctness — the mirror has states with no convergence path

**1. Orphaned sandbox mainnet rows have no delist path when the production fetch stops succeeding** — `apps/sdp-api/src/cron/earn-catalogue-sync.ts:217`

The sandbox own lane delists only `devnet` (`delistScope: environmentCluster`), and the only mainnet-scoped delist sits *after* `if (!production.ok) return`. So once a provider's production fetch fails or becomes a steady-state skip (e.g. production credentials revoked after rows were mirrored), nothing can ever delete a sandbox row with `host_cluster = 'mainnet-beta'` — `deleteUnlistedStrategies` has exactly one call site, and no other cron or migration deletes from `earn_strategies`. The stale rows are served indefinitely under `?cluster=mainnet-beta` as a "production catalogue" production no longer vouches for, and the comment "there is no mainnet shelf to mirror and nothing goes stale" is false for this state. (Pre-0057 legacy rows are safe — the backfill plus COALESCE covers them; the stuck class is explicit `mainnet-beta` rows.)

**2. The mirror lane is gated on a successful sandbox own-fetch** — `apps/sdp-api/src/cron/earn-catalogue-sync.ts:181`

`if (!own.ok) return` skips the mirror too. Two consequences:
- A **mainnet-only provider** (sandbox fetch = steady-state `PROVIDER_NOT_CONFIGURED`/`NOT_IMPLEMENTED`) is *never* mirrored — permanent absence, not "one hour of staleness", for arguably the provider class the mirror exists for.
- A provider whose sandbox source starts failing **persistently** freezes its previously mirrored mainnet rows forever (no re-upsert, no delist).

The coupling is deliberate (the collision rule needs the fresh own reference set), which justifies skipping on *transient* failure — but the own-fetch path doesn't distinguish `steadyState` the way the production-fetch path right below does, and the doc comment "one hour of staleness, never a wrong write" doesn't hold for either persistent case.

**3. A legitimately empty mainnet shelf never converges, and warns forever** — `apps/sdp-api/src/cron/earn-catalogue-sync.ts:448`

If production's accepted mainnet shelf shrinks to zero rows for a provider, the mirror lane calls `writeCatalogueLane` with empty `snapshots` and `keepWithoutUpsert`, so the `listedProviderReferences.length === 0` guard skips the delist **every pass** — previously mirrored sandbox rows stay active forever. Each hourly pass also logs `"skipped delist pass on an unreliable catalogue pass"` (`listed_count: 0, upsert_failed: false`) for what was a perfectly reliable empty read. The guard is pre-existing, but the forever-stale mirror shelf is new surface; the mirror lane could distinguish "reliable empty read" from "unreliable pass".

**4. Mirrored rows are never metrics-refreshed, and `earn-metrics-refresh.ts`'s comment is now false** — `apps/sdp-api/src/cron/earn-metrics-refresh.ts:62`

The 5-minute refresh writes by `(provider, reference, environment)`: Kamino returns `[]` outside production, and the production pass writes only `environment='production'`, so sandbox mirror rows are refreshed by neither pass — hourly-stale APY/TVL beside 5-minute-fresh production twins, and pinned-at-null APY for any provider whose catalogue snapshot omits the optional `currentApy`. The untouched comment — "A mainnet-only provider is catalogued into both, and its rows in both want the same fresh figures" — describes the pre-PRO-1742 architecture. If hourly staleness is the accepted trade-off, the comment should say so.

**5. (Minor) A production operator pause doesn't reach the mirror** — `apps/sdp-api/src/cron/earn-catalogue-sync.ts:392`

The mirror is a second write of the provider *snapshot* (`status: "active"`), never a read of production's row, so an emergency-stopped vault stays visibly active in the sandbox review surface while production hides it. Curation-by-hiding does propagate (the cluster-keyed `HIDDEN_VAULTS`/`CURATED_VAULTS` is a nice touch) — operator status is the one fidelity gap. Safety holds (`fundable: false` throughout), so this is a fidelity note, not a blocker.

### Design note

**6. The collision machinery is compensating for the upsert key lacking `host_cluster` — and Ground collides today** — `apps/sdp-api/src/cron/earn-catalogue-sync.ts:249`

The whole own-wins/skip/`keepWithoutUpsert` apparatus exists because `ON CONFLICT (provider, provider_reference, environment)` can't hold two clusters' rows for one reference. That's fine for Kamino (cluster-specific addresses, zero overlap), but Ground's `providerReference` is a provider-local slug (`source.id`, e.g. `morpho-gauntlet-usdc`) with an environment-derived `hostCluster` — any slug listed in both its catalogues collides on **every** pass, so Ground's mirror shelf would silently under-report (hourly warns only) once Ground is surfaced. Extending the unique key isn't free either (`updateStrategyMetrics` updates by the bare triple and would hit both rows), so I'm not saying "just change the key" — but this deserves an explicit decision recorded somewhere, since the current shape quietly caps the mirror at address-keyed providers.

### Frontend

**7. The cluster-toggle SWR wiring double-fetches** — `apps/sdp-web/src/app/dashboard/markets/treasury-solutions/treasury-solutions-workspace.tsx:1120`

Three compounding issues:
- `onClusterChange={setCatalogueCluster}` stores the raw segment value, so toggling Mainnet → Devnet leaves the key at `["dashboard-earn-strategies", "devnet"]` — permanently distinct from `"environment-default"` though both fetch the identical shelf. The comment that "on the default shelf both hooks share one SWR key, so no second fetch happens" is false for the rest of the session after one round-trip. Normalizing to `undefined` when the value equals `CLUSTER_BY_SDP_ENVIRONMENT[sdpEnvironment]` fixes it.
- `onRefresh` calls both `refreshStrategies()` and `refreshCatalogue()`; pre-toggle these are two un-deduped `mutate()`s of the *same* key → the up-to-20-page pagination runs twice per Refresh click.
- `useEarnStrategies` passes no `keepPreviousData`, so each key flip tears the table to skeletons while the full pagination reruns; the repo already does this right in `activity-tab.tsx` and `wallet-card-balance-value.tsx`.

**8. Availability→badge mapping is now duplicated across the two workspaces** — `treasury-solutions-workspace.tsx:363` / `earn-program-workspace.tsx:94`

`DepositAvailabilityBadge` and `availabilityMessageKey` + inline ternary implement the same mapping (same `cluster_unavailable` interpolation, same four fallback reasons, `"{cluster} only"` duplicated in two i18n namespaces), and the earn-program map's exclusion of `cluster_unavailable` is enforced only by a comment — its `default:` case silently swallows the value. The next availability variant has to be threaded through a switch, a 5-level ternary, and two JSON namespaces; missing one reproduces exactly the "bare Unavailable" collapse this PR's tests assert against. A shared badge component next to `earnVaultDepositAvailability` (parameterized by namespace) would cover both.

### Docs & style

**9. Four authoritative docs still state the invariant this PR reverses.** The CLAUDE.md files were updated, but the documents they designate as the invariant record were not, and `git grep PRO-1742 -- docs/` is empty:
- `docs/decisions/0002-earn-provider-pluggability.md:568` — "The catalogue sync refuses to STORE a mainnet instrument outside production"
- `docs/contributing/earn-pluggability-playbook.md:244-247` — "non-production may never STORE a mainnet instrument… that is the intended outcome, not a gap to work around"
- `packages/sdp-earn/README.md:19` (+ its sync section) — "catalogued per environment's own cluster", one lane per environment
- `docs/architecture/earn-v1-data-flow.md:97` — "the sync refuses to STORE a `mainnet-beta` instrument"

A future provider author following the CLAUDE.md pointers will conclude sandbox mainnet rows are a bug awaiting delist. An ADR addendum + doc sweep should land with this PR.

**10. (Nit) `deleteUnlistedStrategies` builds its cluster filter off-pattern** — `earn.repository.postgres.ts:250`

Twin ternaries (`clusterClause` string + separate `clusterBindings` array that must agree positionally, spliced into the SQL) — while `listStrategies` 25 lines below expresses the identical COALESCE filter with the file's `conditions.push()`/`bindings.push()` pattern. Bindings are correct as written; the risk is the next edit, in the one query where a mis-bound predicate deletes customer-visible rows.

### Verified non-issues (so you don't re-chase them)

- The `"Mainnet only"` badge taking precedence over paused/surfacing states: deliberate, tested, and the non-surfaced/non-`vault_direct` arms are unreachable through the list (SQL filters to `SURFACED_EARN_PROVIDERS`; only Kamino is surfaced, and it's `vault_direct`).
- The foreign-cluster drop logging warn instead of the old guard's error: consistent with the file's convention (operational failures = error, provider drift = warn — the declared-support drop is warn too).
- `catalogueCluster` surviving a sandbox→production→sandbox switch: unreachable — the environment is derived from the selected project, and the project switch changes the `SWRConfig` `key`, remounting the whole subtree.
- Repository bind ordering, the `testnet` 400, NULL-`host_cluster` handling in both new predicates, and the collided-reference keep-set corner all check out.

---

The strongest theme is #1–#3: the mirror's "accepted staleness" is documented as one hour, but three realistic states (revoked production credentials, a failing sandbox source, an emptied mainnet shelf) make it *unbounded* with no convergence or alerting path. A small "sandbox mainnet rows with no surviving truth source get delisted after N passes" rule — or at least an error-level signal — would close all three.

keeganthomp:
All ten addressed in aa0f126f4.

**1–3 (unbounded mirror staleness).** Real, and the sharpest theme, so the mirror lane now distinguishes reliable answers from outages in both directions:

- A steady-state production skip (stub, credentials absent or revoked) converges the mirror to EMPTY, giving orphaned sandbox mainnet rows the delist path they lacked. A successful fetch whose accepted mainnet shelf is empty converges the same way, which also ends the hourly false "unreliable pass" warn.
- Mechanism: `deleteUnlistedStrategies` accepts `allowEmptyKeepSet`, valid only cluster-scoped (the repo throws otherwise). Fundable own shelves keep the absolute empty-keep-set refusal; the asymmetry flips only for browse-only, hourly re-mirrored rows.
- A steady-state own-fetch skip no longer blocks the mirror (it proceeds with an empty own reference set), so a mainnet-only provider is mirrored at all. A transient own failure still skips both lanes, as you argued it should.

New tests cover all four states (steady-state production skip, emptied mainnet shelf, mainnet-only provider, transient own failure) plus repo tests for the authorized-empty delist and the missing-scope throw.

**4 (mirror metrics).** Accepted trade-off, now stated at the flagged comment: mirror rows refresh hourly via the mirror upsert (`currentApy` and `riskMetadata` ride the snapshot), and crossing environments in the 5-minute pass would need the same collision protection as the mirror lane. The "catalogued into both" sentence is gone.

**5 (operator pause).** Documented as a known fidelity gap (lane docstring + ADR addendum), not fixed, per your non-blocker framing.

**6 (collision cap).** Recorded at the collision rule and in the new ADR 0002 addendum: keep the bare triple, since extending the key double-writes every bare-triple consumer (`updateStrategyMetrics` above all); revisit if a slug-keyed provider ever needs a faithful mirror.

**7 (SWR wiring).** All three: the toggle handler normalizes the environment's own cluster back to `undefined`, Refresh only mutates the catalogue key once it has left the shared default, and `useEarnStrategies` gains `keepPreviousData` (the card's skeletons now gate on "no data yet"). Two new workspace tests pin the normalization and the one-revalidation click.

**8 (badge duplication).** One `EarnDepositAvailabilityBadge` in `earn-market-presentation.tsx`, parameterized by an exhaustive `Record<EarnVaultDepositAvailability, MessageKey>` label map per namespace, so the next variant is a compile error in both maps instead of a bare "Unavailable".

**9 (docs).** ADR 0002 gains a PRO-1742 addendum (with a superseded pointer at the 2026-08-14 rule); the playbook, sdp-earn README, and earn-v1-data-flow now describe the two-lane model, the opt-in, and the convergence rules.

**10 (delist filter).** Rebuilt on the file's `conditions.push()`/`bindings.push()` pattern.

cutbow7 (approved):

PR #1486, opened 2026-08-26
feat(counterparties): collect PII JIT in ramp requirements

multipletwigs:
We should just collect PII data as part of the JIT Requirements flow and counterparty is just for recording ramps/onchain transactions + address bookeeping.


multipletwigs:
closing to split into smaller PRs.

PR #1490, opened 2026-08-26
chore(deps): upgrade @solana/kit 7.1.1

snowmead:
## Problem

The workspace is already a Kit 7 workspace running on a Kit 6 resolution. On `main`, twelve dependencies declare `@solana/kit: ^7.0.0` and are resolved against 6.10.0 — visible only because `strictPeerDependencies: false` swallows the warnings:

```
@solana/kora@0.3.0-beta.3                     ^7.0.0    fee sponsorship, live
@solana/subscriptions@0.5.0                   ^7.0.0    recurring payments, live
@solana/surfpool@1.5.0                        ^7.0.0    api-integration harness
@solana/kit-plugin-{rpc,signer,instruction-plan}        pulled by Kora
@solana-program/{system,token,token-2022,record,zk-elgamal-proof}
```

Two of those sit on live money paths. Kora 0.3.0-beta.3 pulls `kit-plugin-instruction-plan@0.13.0`, whose 7.0.0 default caps planned transaction messages at 16 top-level instructions; running its Kit 7 planner against a Kit 6 host is a behavioral question nothing had answered. `@solana/pay@1.0.26` is the only direct dependency still asking for Kit 6.

The `# Trial upgrade to satisfy upstream Solana keychain v1 peer requirements` comment and the `strictPeerDependencies` relaxation are residue from that drift.

## What changed

1. **Catalog to 7.1.1** — `@solana/kit` and the nine `@solana/*` subpackages in `pnpm-workspace.yaml`. `@solana-program/*` and `@solana/mosaic-sdk` are untouched; they already sit on the versions that peer `^7.0.0`.
2. **`@solana/pay` covered by scoped rules** — `peerDependencyRules.allowedVersions` now names `@solana/pay>@solana/kit` plus the three sibling peers a strict resolve surfaced (`@solana/keys`, `rpc-subscriptions-api`, `rpc-subscriptions-spec`). It resolves onto 7.1.1 and the lockfile carries one resolution for it.
3. **Root `package.json` npm-style blocks deleted** — it carried duplicates pnpm never reads: an `overrides` block pinning 6.5.0 and a `workspaces.catalog` pinning 6.8.0, both stale against the 6.10.0 the lockfile actually resolved. `check-pinned-dependency-versions.mjs` validates their format but not their value, so the drift passed CI unnoticed. An earlier revision of this PR synced them to 7.1.1; QA pointed out a synced copy just drifts again, so both blocks are gone instead. The check validates `workspaces.catalog` only when present, and nothing else reads either block. `workspaces.packages` stays.
4. **`@sdp/kit-augment` deleted** (second commit) — it existed because kit 6.8 shipped `extendClient` without exporting the `ExtendedClient` type the generated Codama plugin references. Kit 6.10 already exported it, and the shim on `main` was already a bare `export {}`, so this removal stands alone rather than being required by Kit 7; it rides along here as cleanup. Removes the package, the side-effect import from both SPC client barrels, the module-boundary entry, and regenerates the module map.

No source changes anywhere else. All 153 files importing `@solana/kit` compile untouched.

## One kit major for the workspace, not for the tree

`@solana/token-acl-gate-sdk@0.3.0` hard-pins `@solana/kit: 6.10.0` as a regular dependency and is a direct dep of `sdp-api` and `sdp-issuance`, so it keeps its own nested 6.10.0 copy after this PR:

```
apps/sdp-api          -> 7.1.1
packages/sdp-issuance -> 7.1.1
token-acl-gate-sdk    -> 6.10.0  (its own copy)
```

The one place the two majors meet is `packages/sdp-issuance/src/mosaic/service.ts:667`: kit 7 addresses go into the kit 6 `findWalletEntryPda` and the PDA comes back into a kit 7 RPC call, on the allowlist check used during concurrent mints. QA derived the same PDA both ways across three address pairs, so address encoding and PDA derivation are stable across the major.

## Why the 6 to 7 diff is this small

Every 7.0.0 changelog for the packages this repo imports — `transactions`, `signers`, `addresses`, `codecs`, `rpc`, `rpc-subscriptions`, `instructions`, `keys`, `programs`, `sysvars` — is patch changes only. The major was carried by two things:

- The `ReactiveStreamStore` redesign in `@solana/subscribable`. Nothing here uses it; `createSolanaRpcSubscriptions` (`packages/sdp-rpc/src/solana.ts:192`) is unaffected.
- Dropping kit's own `getMinimumBalanceForRentExemption` helper. `packages/sdp-rpc/src/solana.ts:473` already calls the RPC method the changelog points at, and every other call site routes through it.

## Why not 8.0.0

`@solana/kit` latest is 8.0.0, but `@solana/kora` has no Kit 8 release and the Kit 8 plugin line peers `^8.0.0`, so Kora would have to move first. It should be a short follow-up when it does: this repo already uses the Kit 8 replacements (`getTransactionSizeLimit`, `isTransactionWithinSizeLimit` in `transfer-batches/transaction.ts`) and none of the six APIs Kit 8 removes.

## Why `strictPeerDependencies` stays false

Tried it. A from-scratch resolve under `true` fails on peer ranges this repo does not control: dozens of `typescript@^5.0.0: found 6.0.3` across the Solana packages, Kora's `kit-plugin-*` peers that nothing declares, and klend-sdk's `kit ^3.0` subtree. None are Kit 7 specific. The stale "keychain v1 trial" comment is replaced with what is actually blocking, and flipping it belongs in its own PR.

## Validation

On the branch rebased onto `main` (ca2a84c8):

- Lockfile rebuilt the way QA validated it: `main`'s lockfile taken wholesale, then re-resolved from this branch's manifests. The #1515 `bigint-buffer` vendoring survives — zero `1.1.5` references, `link:packages/bigint-buffer` throughout, and a second `pnpm install --lockfile-only` is a no-op.
- `pnpm typecheck` — 20/20 after the catalog move, 19/19 after the shim removal; each commit typechecks on its own. Counts moved from the original 19/18 because `main` gained workspace packages since.
- `pnpm check:pinned-dependencies`, `pnpm check:module-boundaries` — pass.
- `biome` — clean on every changed file.
- `pnpm install --frozen-lockfile` — exit 0.
- Unit — 16/16 workspace test tasks, 6,235 passed / 18 skipped / 0 failed (`@sdp/api` 4,013 passed / 14 skipped, `sdp-web` 1,564 passed).
- Integration, Kora on Surfpool, and the issuance E2E were run by QA on this branch merged into `main` ([review](https://github.com/solana-foundation/solana-developer-platform/pull/1490#issuecomment-5455795679)): sdp-api 3,850 passed / 0 failed, kora on surfpool 9/10 (the reservation-ceiling failure reproduces on unmodified `main`: the local Surfpool Kora seeds a different `fee_payer_policy` than deployed devnet, so `feePayerMayTransferLamports` is false there regardless of Kit version), mosaic-token-acl 3/3, mosaic-abl 5/5, transfer-batches 1/1, issuance E2E 18 across all five shards, plus a manual pass: wallet created and funded, transfer driven to finalized, allowlist entry added and removed.

## Notes

`pnpm dedupe` collapses a leftover `@solana/codecs-*@7.0.0` set but also bumps `@types/hast` 3.0.4 to 3.0.5, so it is deliberately not run here.

Rings landed first (#1500) and this branch is rebased on top of it. Once this merges, the follow-up to drop the Rings SDK's Kit-neutral scaffolding — the `./testing` subpath, `kit-cross-major.test.ts`, and the string-DTO port — and put `@sdp/helius-rings-sdk` on the catalog is unblocked.


TarekkMA (dismissed):

GuiBibeau:
@snowmead doing a manual QA run of some happy paths. We are also increasing number of product alerts right now to be able to merge this at ease.

resourcefulmind:
Did the manual QA on this. Pulled it locally and merged main into it first, so I was testing what will actually ship rather than the branch on its own.

No defect found, happy for it to go in. Three things worth knowing.

**The leftover kit 6.10.0 isn't orphaned.** `@solana/token-acl-gate-sdk@0.3.0` pins it as a hard dependency and it's a direct dep of both `sdp-api` and `sdp-issuance`, so after install you get:

```
apps/sdp-api          -> 7.1.1
packages/sdp-issuance -> 7.1.1
token-acl-gate-sdk    -> 6.10.0  (its own copy)
```

So this splits the resolution. One kit major today, two afterwards. It is not a problem in itself, but the PR body says the opposite and someone will read that later.

**I tested the one place the two majors meet and it's fine.** `mosaic/service.ts:667` passes kit 7 addresses into a kit 6 `findWalletEntryPda` and the result back into a kit 7 RPC call, on the allowlist check used during concurrent mints. I derived the same PDA both ways across three address pairs and got identical results, so address encoding and PDA derivation didn't change across the major. Worth a line in the body rather than a fix.

**The rebase needs a bit of care.** The lockfile is the only conflict, but this branch predates #1515, so it still carries `bigint-buffer@1.1.5` while main has the vendored workspace package, and both sit under the subtree kit 7 re-resolves. I did the merge locally, took main's lockfile and reinstalled, and it comes out right: kit 7.1.1 with `link:packages/bigint-buffer` and zero references to 1.1.5. Just don't regenerate the lockfile from this branch's manifests or that CVE fix goes backwards.

On the merged tree: typecheck 19/19, sdp-api 3,850 passed and 0 failed, kora on surfpool 9/10 (the reservation-ceiling one, which fails on main too), mosaic-token-acl 3/3, mosaic-abl 5/5, transfer-batches 1/1, and the full issuance E2E at 18 across all five shards including allowlist. Then drove it by hand: created and funded a wallet, sent a transfer through to finalized, added and removed an allowlist entry. Nothing odd.

My additional notes would be: `kit-augment` was already an inert `export {}` on main referring to kit 6.10, so its removal isn't really caused by this upgrade and would bisect more cleanly on its own. And the root `package.json` blocks being reconciled are ones pnpm never reads (`overrides` rather than `pnpm.overrides`, and `workspaces.catalog`), so hand-syncing them will just drift again.

Cc @GuiBibeau

TarekkMA (approved):

PR #1492, opened 2026-08-26
feat(payments): log and capture swallowed ramp provider estimate errors

niks3089:
## What

`estimateAcrossProviders` in `apps/sdp-api/src/routes/payments/handlers/ramps.ts` catches every provider error into `{ status: "error" }` inside an HTTP 200 with no telemetry — a ramp provider can fail 100% with no signal (this is how one provider errored in a live demo unnoticed). This keeps the 200 response contract unchanged and makes the catch site observable:

- **Structured log event** `sdp_api_ramp_provider_error` via the existing `logEvent` money-path convention (`provider`, `organization_id`, `error_name`/`error_code`, credential-redacted `error_message`). The `sdp_api_` prefix follows every existing event in the codebase (the proposal's shorthand was `ramp_provider_error`).
- **Sentry capture** at the catch site, gated by `isSentryEnabled` like the global handler, tagged with provider + org. `redactErrorForCapture` moved verbatim from `app.ts` to `lib/errors.ts` (exported) so both call sites share one redactor without an app→routes import cycle.

A follow-up PR in `sdp-infra` adds the single Grafana alert rule on this event's rate (one rule only — over-alerting is an explicit stakeholder concern).

Part of the SDP Improvement Proposal, phase 1 (item: make swallowed ramp provider errors observable / G1).

## Verified

- `pnpm --filter @sdp/api typecheck`: pass.
- `pnpm --filter @sdp/api lint` (biome): clean.
- New unit test `ramps.estimate-fanout.test.ts` (2 tests): provider throw → per-provider error entry preserved **and** log event emitted with the right payload; success path emits nothing. Ran locally with the suite's container global-setup stubbed out (no Docker on this machine — testcontainers Postgres/Redis can't start); CI runs the full suite with services.


niks3089:
@greptile review

multipletwigs (approved):

PR #1491, opened 2026-08-26
test(web): wire sdp-web unit tests into CI by renaming test:unit to test

niks3089:
## What

`apps/sdp-web` had a `test:unit` script but no `test` script, so CI's `turbo run test` (via `scripts/run-workspace-tests.mjs`) skipped the workspace entirely — its 161 unit-test files (1,264 tests) never ran in CI. This renames the script key to `test`, matching every other workspace (no workspace delegates; all use the runner directly), and updates the one doc that described the old behavior.

No skip-list was needed: the full suite passes as-is despite being dark.

Part of the SDP Improvement Proposal, phase 1 (item: re-wire the dashboard's dead unit tests into CI).

## Verified

- `pnpm exec turbo run test --dry-run` with CI's filter (`!@sdp/api-integration`) now includes `sdp-web#test`.
- `pnpm exec turbo run test --filter=sdp-web`: **161 files / 1,264 tests pass** (10s).
- `pnpm --filter sdp-web typecheck`: pass.
- `pnpm --filter sdp-web lint`: crashes locally on clean `main` too (`eslint-plugin-react` × ESLint 10 under Node 26, `contextOrFilename.getFilename is not a function`) — pre-existing environment issue, unrelated to this diff (which touches only `package.json` + a CLAUDE.md); CI runs Node 24.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1494, opened 2026-08-26
ci(smoke): gate release promotions on the dev smoke, plus a 30-minute heartbeat

niks3089:
## What

The `gcp-read-only` Playwright project (pinned to https://api-dev.solana.com, gated by `PLAYWRIGHT_USE_EXTERNAL_API=1`) existed but nothing ran it (proposal gap G2). This wires it in as a **prod promotion gate**: dev is the staging bake, and prod deploys only after the smoke passes against it.

- **New reusable workflow** `.github/workflows/sdp-dev-smoke.yml` (`workflow_call` + `schedule: */30` + `workflow_dispatch`): builds sdp-web against the dev API and runs the smoke via a new `test:e2e:gcp-smoke` script. Setup mirrors ci.yml's `browser_e2e` job minus the local Postgres/Redis/Surfpool stack; Clerk + fixture credentials come from Doppler `dev_ci`.
- **Prod gate**: `deploy-sdp-api-gcp-prod.yml` gains a `smoke` job and the `deploy` job takes `needs: smoke` with `needs.smoke.result == 'success'` folded into its existing `if` — a red smoke against dev blocks the release-triggered deploy AND manual `workflow_dispatch` redeploys. `release-please.yml` passes `secrets: inherit` so the nested smoke can reach Doppler.
- **The dev deploy workflow is untouched** — dev keeps deploying on every main push (that's what makes it the bake environment), and the 30-minute cron with Slack-on-failure catches dev breakage between releases.

- **Deploy notifications**: both sdp-api deploy workflows now post started/success/failure to Slack via the repo's existing \`scripts/notify-slack.sh\` (service, version, actor, run link — same shape as the other services in #deployments). The script was orphaned since \`deploy-kora.yml\` was deleted; SDP deploys were silent. Fails soft + \`continue-on-error\`, so a Slack hiccup can't affect a deploy. Note: requires \`SLACK_DEPLOY_WEBHOOK_URL\` in Doppler \`dev_ci\` — if it lives elsewhere, the script logs a warning and skips.
- **Explicit secrets**: \`secrets: inherit\` replaced with declared \`DOPPLER_TOKEN_CI\` at every hop (release-please → prod deploy → smoke) — the repo's \`release-production-workflows.test.mjs\` policy test forbids inherit in the release chain.

No new test code — wiring only.

Part of the SDP Improvement Proposal, phase 1.

## Verified

- `actionlint` clean on all four touched workflows; all action refs full-SHA pins copied from ci.yml.
- `pnpm --filter sdp-web run test:e2e:gcp-smoke --list` resolves 3 smoke tests + auth-setup.
- sdp-web unit suite (1,366 tests) passes after the `package.json` conflict resolution with #1491's script rename.
- Not verified here: a live smoke run (needs Doppler `dev_ci` secrets this machine doesn't have) — trigger the new workflow via `workflow_dispatch` once merged, before the next release relies on it.

**Review updates:** manual `workflow_dispatch` redeploys (the rollback path) skip the smoke gate — only release-triggered promotions are gated; cron halved to */30 per review.


niks3089:
@greptile review

G1de0n (changes requested):
Heads up: the smoke gate also fires on manual workflow_dispatch runs of deploy-sdp-api-gcp-prod.yml. That's our rollback path.

deploy has needs.smoke.result == 'success' folded into its if. Dev redeploys on every push (that's what makes it the bake environment), so an unrelated main push breaking dev turns smoke red and blocks a prod rollback via workflow_dispatch + image_sha until dev's fixed. Rolling back prod can't depend on dev being healthy.

Cheap fix: add a skip_smoke boolean input (default false) to workflow_dispatch, gate on needs.smoke.result == 'success' || inputs.skip_smoke. Or just scope the smoke gate to the release-triggered path and leave manual dispatch ungated.

Two non-blockers while I'm in here: */15 cron builds all of sdp-web ~96x/day, */30 halves that with basically no signal loss. Also worth keeping in mind, the gate validates whatever's currently running on dev, not the release image being promoted. Fine as a bake check, not a verification of the promoted artifact.

niks3089:
@G1de0n all three addressed:

1. **Rollback path ungated** — the smoke job now skips on `workflow_dispatch` (`github.event_name != 'workflow_dispatch'`), and the deploy condition accepts `skipped` only on the dispatch path. Went with your second option rather than `skip_smoke`: manual dispatch can only redeploy an existing image SHA — an artifact that already passed the gate when it was released — so gating it added no safety, and an incident is the wrong time to ask a human to remember an input flag. Release-triggered deploys remain hard-gated.

2. **Cron halved to `*/30`** — agreed, ~48 full builds/day instead of ~96 for the same signal.

3. **Bake check vs promoted-artifact verification** — agreed and deliberate for this PR. Exact-SHA parity (deploy the release SHA to dev, smoke that, then promote) is scoped in the verification umbrella (PRO-1760 → canary work in PRO-1763). Related note from today: demos run on the prod deployment (sandbox projects/devnet), so the heartbeat probing only dev is itself a gap — also tracked in PRO-1763.

G1de0n (approved):

PR #1499, opened 2026-08-27
refactor(counterparties): drop pii encryption code paths and shorten ids to cpty_

multipletwigs:
Part of the JIT PII work (previous slices: #1497, #1498). Counterparty PII moves to just-in-time collection, so the never-cutover encryption layer is fully retired here — code paths, cipher, env surface, and schema.

**Account repository**

- reads/writes plaintext columns only: cipher usage, migration-phase reads, advisory locks, and the decrypt-first/plaintext-fallback read are deleted
- writes are single-statement via `RETURNING *`: create = one INSERT; update = `SELECT … FOR UPDATE` + `UPDATE … RETURNING *` in one transaction
- `counterparty-pii.repository.ts` deleted — its two surviving helpers folded into their single consumers (`providerLookupReferences` → counterparty repo, `cryptoAccountLookup` → account repo)

**Cipher + env surface**

- pii-cipher service deleted outright (KMS + local AES-GCM implementations and tests)
- `COUNTERPARTY_PII_*` removed everywhere: env.d.ts, sdp-env-config fields + secret-generation tests, self-hosted `.env.example`, ci.yml, self-hosted smoke workflow

**Counterparty ids**

- new ids are `cpty_<uuid>` (was `counterparty_<uuid>`); generator and `SDP_COUNTERPARTY_ID_PATTERN` now colocated in `@sdp/payments/src/counterparty.ts` next to `CounterpartyRow`
- the old prefix is retired entirely (review feedback): `SDP_COUNTERPARTY_ID_PATTERN` is anchored to `^cpty_<uuid>$` rather than prefix-agnostic, so the BVNK `externalReference` mangle rejects a `counterparty_` id loudly instead of silently mapping both prefixes to the same `cp_<32hex>` reference — one canonical prefix, no dual-format support to reason about later
- pre-mainnet this costs nothing: any lingering devnet row with an old-prefix id fails loudly at the BVNK boundary rather than being quietly grandfathered in
- all test fixtures moved to `cpty_` ids, plus a regression test that a retired-prefix id is rejected
- OpenAPI id examples → `cpty_example` (including 8 stale `cp_example` ones — `cp_` is the BVNK external-reference prefix, not the id prefix); postman collection + playground catalog regenerated

**Schema**

- migration `0070` drops `pii_encrypted`, `provider_data_encrypted`, `sensitive_data_encrypted`, and the `counterparty_pii_migration_state` table
- shipping the DDL in the same PR is safe here: with this PR nothing reads or writes any of these — and the encrypted columns were never authoritative (every environment stayed in `dual_write`, plaintext always populated), so the deploy-window/rollback concern from #1498 doesn't apply in practice with the payments module dark
- deploy window, explicitly: prod runs the migrate job before the new revision, and in that window the old revision's account reads would 500 (`recordCounterpartyPiiFallbackRead` targets the dropped state table). Confirmed acceptable: payments is dark in prod — no counterparty-account traffic exists to hit that path, and rollback-to-previous-image is equally moot for the same reason

**before** — account writes serviced the dormant encryption rollout: advisory lock → phase read → KMS/local encrypt → dual plaintext+ciphertext write → re-SELECT; reads decrypted ciphertext first, fell back to plaintext, incremented telemetry

**after** — plain relational writes: INSERT/UPDATE plaintext with `RETURNING *`, row-level `FOR UPDATE` only; reads map plaintext columns; the ciphertext columns and state table no longer exist

**Upcoming changes**

- `provider_data` is the last nested-JSON surface on counterparties: reverse-lookup values (`bvnk.customer.customerReference`, `mural.organization.id`) are extracted with hand-rolled structural traversal (`nestedString`, `readRecord`-style checks). Direction (PRO-1752): denormalize provider links into a dedicated `counterparty_provider_accounts` table — DB-enforced columns with zod runtime validation at the boundary — retiring the traversal helpers and the `provider_data` expression indexes.

**Verification**

- `npx tsc --noEmit` — clean in `sdp-payments`, `sdp-api`, and `sdp-env-config`
- `npx vitest run` — BVNK provider-data + counterparties routes + webhooks: 80 passed; counterparties + account + payment-requests suites re-run against the post-0070 schema (vitest global setup applies all migrations to a fresh DB): 64 passed
- `npm test` in `packages/sdp-env-config` — 48 passed


eldarik:
**Approve the code, hold on the migration** 🟡 — the refactor is clean (net −632, writes down to one statement each, the `repository-factory` test-cipher cast gone), and catching the regex problem is the best thing here: with only the generator changed, `buildBvnkCustomerExternalReference`'s `^counterparty_<uuid>$` pattern would have thrown on every new `cpty_` id and broken BVNK customer creation. Nice catch. My concerns are all in `0070`, not the TypeScript.

🔴 **`0070` drops the account ciphertext in the same release as the first code that stops reading it, with nothing verifying the premise.** For `counterparties` the plaintext-only read is already live from #1498. For `counterparty_accounts` it isn't — the currently-deployed revision still decrypts `sensitive_data_encrypted` first (it was bound unconditionally on every write), so this PR is the first code to depend on account plaintext *and* it deletes the fallback in the same deploy. `0036` explicitly made `details`/`provider_account_data` nullable to allow ciphertext-only rows, so "plaintext was always populated" is an assumption about every environment including self-hosted ones we can't inspect. Could the migration assert it instead of asserting it in a comment? ~8 lines:
```sql
DO $$ BEGIN
  IF EXISTS (SELECT 1 FROM counterparty_accounts
              WHERE details IS NULL OR provider_account_data IS NULL) THEN
    RAISE EXCEPTION 'ciphertext-only account rows exist; refusing to drop sensitive_data_encrypted';
  END IF;
  IF EXISTS (SELECT 1 FROM counterparty_pii_migration_state WHERE phase <> 'dual_write') THEN
    RAISE EXCEPTION 'counterparty PII migration advanced past dual_write';
  END IF;
END $$;
```

🔴 **Prod runs the migrate job before the new revision, so the old revision serves against the new schema.** `deploy-sdp-api-gcp-prod.yml` executes `sdp-prod-api-public-migrate` at L124–126 and waits, then deploys the service at L261/L302. In that window the old code's every counterparty-account *read* falls to the plaintext branch and calls `recordCounterpartyPiiFallbackRead` → `UPDATE counterparty_pii_migration_state` → relation does not exist. Hard 500 on the whole surface, not a degradation — and the workflow's rollback step (L278–280, previous image) can't restore dropped columns, so a post-release rollback leaves it broken until roll-forward. I think "payments is dark in prod" makes the real impact zero, but that's the only thing standing between this and an outage, so it's worth confirming in the PR body rather than a parenthetical. The alternative is one extra release (code now, DROP next).

🟡 **`0070` should also restore what `0036` loosened.** `0036` did `ALTER COLUMN details DROP NOT NULL, DROP DEFAULT` (same for `provider_account_data` / `counterparties.provider_data`) *only* to permit ciphertext-only rows. That loosening now outlives the rollout while the mappers moved the other way — `row.details as CounterpartyAccountDetails` with the old `?? {}` gone, and `listBatchRecipients` dereferencing `account.details.address`. Worth backfilling `'{}'` and re-adding `SET DEFAULT '{}'` / `SET NOT NULL` on those three columns while you're in there (leave `identity`/`email` nullable — JIT collection needs those).

🟡 **`COUNTERPARTY_PII_*` isn't gone everywhere.** `scripts/secret-keys.mjs` L24–25 still lists both keys in `API_LOCAL_ENV_KEYS` — "every env key the SDP API reads" — which feeds `DOCKER_ENV_KEYS` → the self-hosted docker env file. Nothing catches it: `check-env-contract.mjs` only flags keys missing from the list, never extras.

🟡 **`cp_<32hex>` isn't reversible any more, but the JSDoc still says it is.** Both prefixes now produce the same reference, so you can't tell which one to prepend going back. Nothing reverses it today (BVNK webhooks resolve via the stored `bvnk_customer_reference` column, not by parsing `cp_`), so no bug — but the "letting handlers reconstruct the SDP counterparty id and load by primary key" sentence is now false and will mislead whoever implements that path.

⚪ Minor: counterparty-id examples are now three shapes in the public docs — `cpty_example` (the 5 you changed), `cp_example` (`payments.ts` 831/901/1051/1096/1145/1162, `counterparties.ts` 37/313), and `counterparty_example` for fields with no explicit example, since `generate-playground-catalog.ts:138` derives the placeholder from the property name — that's the 7 remaining hits in the regenerated catalog, all in `expectedResponse`. Also: the `UPDATE` in `updateCounterpartyAccount` is `WHERE id = ?` only (safe — the `FOR UPDATE` select above it is fully scoped and holds the lock — but `archiveCounterpartyAccount` right below keeps the tenancy predicate in its `UPDATE`, so it reads as an inconsistency); and `.env.example` lost the blank line between the SPC block and the "Public URLs" section.

multipletwigs:
Point by point:

🔴 **Migration assert** — checked the pre-drop write path before deciding: `network`/`address` were bound unconditionally in **every** phase; only `label`/`details`/`provider_account_data` were `dual_write`-gated. Crypto wallet accounts are the only kind ever written (bank details are JIT pass-through by design), so a hypothetical `encrypted_only` row loses at most a cosmetic label — and no environment ever ran the flip. Skipping the assert as not load-bearing.

🔴 **Deploy window** — confirmed and now stated explicitly in the PR body: payments is dark in prod, so there is no counterparty-account traffic to 500 between migrate job and revision swap; rollback-to-previous-image is moot for the same reason.

🟡 **`NOT NULL`/`DEFAULT` restore** — deliberate: `provider_account_data` stays nullable (writes always bind `{}`, the row type stays truthful), and the `provider_data` surface is being denormalized into a dedicated provider-accounts table in the next slice — constraints ride with that redesign rather than churning here.

🟡 **`COUNTERPARTY_PII_*` in `secret-keys.mjs`** — good catch, fixed in 16b5235.

🟡 **`cp_` reversibility JSDoc** — fixed in d44eab5: `SDP_COUNTERPARTY_ID_PATTERN` is now anchored to `^cpty_<uuid>$` (old prefix retired entirely), so `cp_<32hex>` is unambiguously reversible again and the JSDoc sentence is accurate. Added a test that a retired-prefix id is rejected.

⚪ **Minors** — `cp_example` → `cpty_example` (all 8) in d44eab5, postman + playground regenerated; the remaining `counterparty_example` hits are synthesized by `generate-playground-catalog.ts:138` from the property name (`planId` → `plan_example` likewise), left as-is; `.env.example` blank line restored in 16b5235; update-vs-archive `WHERE` asymmetry — agreed it's safe (the scoped `FOR UPDATE` holds the lock), leaving as-is.

pashpashkin (approved):

PR #1500, opened 2026-08-27
feat(helius): implement Rings identity provisioning, shielded reads, and shield deposits

TarekkMA:
## Summary

Turns the Helius Rings module from a port with placeholder gateways into a working
shielded-wallet flow. A new `@sdp/helius-rings-sdk` package runs the Zolana SDK in
process behind the existing `RingsGatewayPort`, so the routes, service, and state
machine added in #1438 keep their shape while the adapter behind them becomes real.

**What now works:** identity provisioning against the user registry, on-chain
identity reads, Photon-backed shielded balance sync, shield (deposit) operations
end to end, and per-upstream health. `transfer` and `withdraw` remain out of scope —
the composer offers `shield` only.

**Key decisions worth a reviewer's attention:**

- **One seam between `@solana/kit` majors.** The SDK is pinned to kit 7 and
  `sdp-api` is on 6. Two majors' branded types match structurally, so a leaked
  `Address` typechecks and then misbehaves at runtime. Only plain strings and
  `@sdp/helius-rings` types cross, `services/helius-rings/gateway.ts` is the only
  file in `apps/` allowed to import the SDK, and `check-module-boundaries.mjs`
  enforces the package's dependency list.
- **Keys are derived, never stored — and the deterministic KA is interim.** Shielded
  material comes from `HELIUS_RINGS_DETERMINISTIC_KA_SEED` per
  `(organization, project, wallet, owner)` and is destroyed after each use. A gateway
  is fixed to one tenant at construction, so a wallet id can't derive material under
  another organization's path. The seed is permanent once any wallet is provisioned —
  re-keying orphans the notes encrypted to the old keys, and SDP refuses it.
  **This is a stand-in until Rings integrates with the Helius key authority**, the
  same component `apps/sdp-helius-gateway` already speaks to; the seed, the derivation
  path, and this whole in-process key handling go away with that integration.
- **A failed broadcast doesn't fail the operation.** Marking `submit_failed`
  retryable would let an operator retry a transaction that may already have landed,
  double-depositing. It now moves to `indexing` with a `broadcast: "unconfirmed"`
  event so Photon decides what actually happened, matching the existing
  crash-recovery path.
- **No credentials in responses.** The RPC URL carries a Helius API key and upstream
  errors routinely quote the endpoint they failed on, so adapter failures and health
  details use fixed text that names the stage and the actionable fix.

**Also:** stale `ready_to_sign` rows are swept to `signer_failed` so a process dying
mid-sign can't strand an operation; `indexedAt` now reports Photon's `blockTime`
rather than SDP's poll clock; and the dashboard consolidates shielded balance into
the wallets table with identity detail behind a dialog.

New config: `HELIUS_RINGS_RPC_URL`, `HELIUS_RINGS_INDEXER_URL`,
`HELIUS_RINGS_PROVER_URL`, `HELIUS_RINGS_DETERMINISTIC_KA_SEED`, and
`HELIUS_RINGS_ALLOW_INSECURE_HTTP` (the public devnet indexer and prover are plain
http, so reaching them is an explicit opt-in rather than something inferred from the
URL scheme). All are registered as secrets. A half-configured deployment answers
health red naming the missing variables instead of 500-ing every request.

## Test plan

- [x] `@sdp/helius-rings-sdk` — 123 tests across 11 files
- [x] `sdp-api` Rings suites — 191 tests across 12 files
- [x] `sdp-web` Rings suites — 55 tests across 5 files
- [x] Typecheck and lint clean on the SDK package and `sdp-api`
- [ ] Devnet walkthrough: provision a wallet, confirm the registry record, shield a
      small amount, confirm the balance and the operation timeline
- [ ] Health endpoint with an upstream deliberately unset — expect red naming the
      variable, not a 500
- [ ] Confirm no Helius API key appears in any error or health response


apps/sdp-web/src/app/dashboard/helius-rings/operation-composer.tsx
```diff
@@ -280,11 +273,13 @@ function ComposeStep({
             onValueChange={(walletId) => onPatch({ walletId })}
             placeholder={t("DashboardHeliusRings.composer.walletPlaceholder")}
           >
-            {wallets.map((wallet) => (
-              <SelectItem key={wallet.id} value={wallet.id}>
-                {wallet.name}
-              </SelectItem>
-            ))}
+            {wallets
```
TarekkMA: Leaving this as a two-pass filter on a two-item allowlist. Combining it would not change behavior, and React Doctor scored this as a warning rather than an error.

packages/sdp-helius-rings-sdk/src/health.test.ts
```diff
@@ -0,0 +1,220 @@
...
+  });
+}
+
+/** Routes by URL so one fake serves both the indexer and the prover. */
+function fetchStub(
+  handlers: Readonly<{ indexer?: () => Promise<Response>; prover?: () => Promise<Response> }>
+): typeof globalThis.fetch {
+  return ((input: string | URL) => {
+    const url = String(input);
+    if (url.startsWith(INDEXER_URL)) {
```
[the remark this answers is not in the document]
TarekkMA: Fixed: the health-test fetch stub now routes by `URL.host` instead of a `startsWith` prefix, so a crafted hostname cannot ride the indexer URL.

packages/sdp-helius-rings-sdk/src/health.test.ts
```diff
@@ -0,0 +1,220 @@
...
+      return (handlers.indexer ?? (() => Promise.resolve(jsonResponse({ result: "ok" }))))();
+    }
+    if (url.startsWith(PROVER_URL)) {
```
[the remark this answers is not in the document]
TarekkMA: Fixed: the health-test fetch stub now routes by `URL.host` instead of a `startsWith` prefix, so a crafted hostname cannot ride the prover URL.

packages/sdp-helius-rings-sdk/src/health.test.ts
```diff
@@ -0,0 +1,220 @@
...
+    const health = await probeRingsHealth(input());
+
+    expect(health).toEqual({ rpc: "green", photon: "green", prover: "green", gateway: "green" });
+    expect(health.detail).toBeUndefined();
+  });
+
+  it("asks Photon for its health over JSON-RPC, since the client has no such method", async () => {
+    const requests: Array<{ method: string; body: unknown }> = [];
+    const capture = ((url: string | URL, init?: RequestInit) => {
+      if (String(url).startsWith(INDEXER_URL)) {
```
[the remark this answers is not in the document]
TarekkMA: Fixed: the Photon request capture now matches `URL.host` rather than a string prefix.

packages/sdp-helius-rings-sdk/src/health.test.ts
```diff
@@ -0,0 +1,220 @@
...
+  it("probes the prover at GET /health", async () => {
+    const seen: Array<{ url: string; method: string }> = [];
+    const capture = ((url: string | URL, init?: RequestInit) => {
+      seen.push({ url: String(url), method: init?.method ?? "GET" });
+      return Promise.resolve(jsonResponse({ result: "ok" }));
+    }) as typeof globalThis.fetch;
+
+    await probeRingsHealth(input({ fetch: capture }));
+
+    const prover = seen.find((request) => request.url.startsWith(PROVER_URL));
```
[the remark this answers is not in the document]
TarekkMA: Fixed: the prover `/health` lookup now matches `URL.host` rather than a string prefix.

packages/sdp-helius-rings-sdk/src/shield.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+const UINT64_MAX = 0xffff_ffff_ffff_ffffn;
+
+export async function buildShieldTransaction(
+  deps: ShieldDeps,
+  input: ShieldInput
+): Promise<string> {
+  const owner = parseAddress(input.owner, "owner");
+  const asset = parseAddress(protocolMint(input.mint), "mint");
```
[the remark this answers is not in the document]
TarekkMA: Fixed: `prepareOperation` now requires an active `helius_rings_asset_allowlist` row before reserving, so an unsupported mint never gets an operation row.

packages/sdp-helius-rings-sdk/src/sync.ts
```diff
@@ -0,0 +1,93 @@
...
+ * label the figure as base units.
+ */
+function toAssetBalance(reportedMint: string, amount: bigint): AssetBalance {
+  const mint = sdpMint(reportedMint);
+  const isNative = mint === SDP_NATIVE_MINT;
+
+  return {
+    mint,
+    symbol: isNative ? NATIVE_MINT_SYMBOL : "UNKNOWN",
+    decimals: isNative ? NATIVE_MINT_DECIMALS : null,
```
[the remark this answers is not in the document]
TarekkMA: Fixed in the API `syncWallet` projection: after Photon returns, allowlisted mints (including seeded USDC) get their symbol and decimals from the platform table. The SDK still only knows native SOL.

apps/sdp-api/src/services/jobs/poll-rings-indexing.ts
```diff
@@ -85,8 +105,22 @@ export async function pollRingsIndexing(
     // indexing transition never committed. The timeout is deliberately not
     // applied to it — the budget measures how long Photon has been asked, and a
     // resumed operation has not been asked yet.
-    if (operation.state !== "indexing" && operation.state !== "submitted") continue;
+    if (
+      operation.state !== "indexing" &&
+      operation.state !== "submitted" &&
+      operation.state !== "ready_to_sign"
+    ) {
+      continue;
```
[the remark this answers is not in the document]
TarekkMA: Fixed: `listInFlightOperations` now queries only `ready_to_sign`, `submitted`, and `indexing`, so waiting `approval_required` rows cannot fill the sweep limit.

apps/sdp-api/src/services/jobs/poll-rings-indexing.ts
```diff
@@ -39,9 +58,9 @@ export interface PollRingsIndexingDependencies {
 }
 
 export function isRingsIndexingPollEnabled(
-  env: Pick<Env, "HELIUS_RINGS_ENABLED" | "HELIUS_RINGS_ADAPTER">
+  env: Pick<Env, "HELIUS_RINGS_ENABLED"> & RingsUpstreamEnv
 ): boolean {
-  return isHeliusRingsEnabled(env) && env.HELIUS_RINGS_ADAPTER === "http";
+  return isHeliusRingsEnabled(env) && ringsUpstreamsConfigured(env);
```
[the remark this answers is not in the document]
TarekkMA: Fixed: the job now runs whenever Rings is enabled. Timeouts still fire without upstreams; Photon `executeOperation` is gated on the configured URLs.

apps/sdp-web/src/app/dashboard/helius-rings/wallet-identity-check.tsx
```diff
@@ -0,0 +1,195 @@
...
+  );
+}
+
+function Address({ label, value }: { label: string; value: string }) {
+  return (
+    <div className="flex flex-col gap-0.5">
+      <dt className="text-xs text-secondary">{label}</dt>
+      <dd className="font-mono text-xs text-primary" title={value}>
+        {shortenShieldedAddress(value)}
+      </dd>
```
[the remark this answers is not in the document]
TarekkMA: Fixed: the identity dialog now renders the full shielded address with wrapping instead of a shortened `title`.

infra/self-hosted/.env.example
```diff
@@ -70,11 +77,19 @@ SELF_HOSTED_STORED_CONNECTION_SETUP_ENABLED=false
 PRIVATE_CHANNELS_ENABLED=false
 
 # Helius Rings is opt-in and devnet-only. Gates the shielded-wallet API
-# routes and the dashboard workspace; production wiring stays off until the
-# live gateway integration lands.
+# routes and the dashboard workspace. Off by default: it reaches external
+# infrastructure and moves no value, so nothing turns it on for you.
```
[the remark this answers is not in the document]
TarekkMA: Fixed: the self-hosted comment now states that enabling Rings can move tokens from the backing custody wallet into the shielded pool.

packages/sdp-env-config/src/fields.ts
```diff
@@ -726,16 +726,45 @@ export const FIELDS: EnvField[] = [
     help: "Gates the devnet-only Helius Rings shielded wallet API routes.",
   },
   {
-    key: "HELIUS_RINGS_ADAPTER",
+    key: "HELIUS_RINGS_RPC_URL",
+    section: "advanced",
+    kind: "url",
+    label: "Helius Rings Solana RPC URL",
+    help: "Helius devnet RPC the Rings SDK reads and submits through, with the API key already applied. Required once Rings is enabled.",
```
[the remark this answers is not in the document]
TarekkMA: Fixed: `HELIUS_RINGS_RPC_URL` is a password field so the configurator masks the API key in the URL.

apps/sdp-api/src/lib/feature-flags.ts
```diff
@@ -30,6 +30,18 @@ export function isHeliusRingsEnabled(env: Pick<Env, "HELIUS_RINGS_ENABLED">): bo
   return isTruthyFlag(env.HELIUS_RINGS_ENABLED);
 }
 
+/**
+ * Permits plain-http Rings upstreams. The public devnet indexer and prover are
+ * http on a real host, so without this the adapter cannot reach them — but in
+ * plaintext an indexer response reveals which notes an identity owns, which is
+ * why it is an explicit choice rather than something read off the URL scheme.
+ */
+export function isRingsInsecureHttpAllowed(
```
nbaztec, line 37: I don't believe this is necessary for a beta service. Any RPCs using http are already aware that they operate on insecure transport layer. We don't need to enforce this behavior for now, and can assume the user uses HTTPS.

apps/sdp-api/src/routes/helius-rings/handlers.ts
```diff
@@ -87,6 +88,74 @@ export async function getRingsWallet(c: AppContext) {
...
+    degraded: synced.degraded,
+    observedAt: synced.observedAt,
+  });
+}
+
+/**
+ * GET /wallets/:walletId/identity — what the Rings registry publishes for this
+ * wallet's owner, and whether it is the identity this tenant derives.
+ *
+ * A GET behind the read permission, unlike `/sync`. A sync advances the
```
nbaztec: This mentions permissions but there's no such guarantee upstream uses those permissions. The comment does not fit here.

apps/sdp-api/src/services/helius-rings/service.ts
```diff
@@ -539,15 +686,38 @@ export class HeliusRingsService {
...
+      // So the failure is not taken. Failing `submit_failed` would mark the
+      // operation retryable, and a retry files a *fresh* operation under a new
+      // client nonce — a different intent key, a newly built transaction, a
+      // second broadcast. Whenever the first one did land that shields the
+      // amount twice. The identical ambiguity after a crash is already resolved
+      // the other way by `executeOperation`, and this takes the same answer:
+      // carry the signature into `indexing` and let Photon settle what actually
+      // happened. A transaction that truly never left ends at
+      // `indexing_timeout`, retryable only after Photon has had the full budget
+      // to disagree.
```
nbaztec: Comment too verbose

apps/sdp-api/src/services/helius-rings/service.ts
```diff
@@ -557,21 +727,31 @@ export class HeliusRingsService {
...
+          ? { code: "invalid_input" as const, message: error.message, retryable: false }
+          : error instanceof HeliusRingsError && error.code === "conflict"
+            ? { code: "invalid_input" as const, message: error.message, retryable: false }
+            : error instanceof HeliusRingsError && error.code === "gateway_unavailable"
+              ? { code: "gateway_unavailable" as const, message: error.message, retryable: true }
+              : {
+                  code: "gateway_unavailable" as const,
+                  message: error instanceof Error ? error.message : "rings gateway failed",
+                  retryable: true,
+                };
```
nbaztec: Prefer to make this a lookup table/switch

apps/sdp-api/src/services/jobs/poll-rings-indexing.ts
```diff
@@ -11,11 +11,18 @@
  *     a hit completes the operation, a miss leaves it untouched (idempotent).
  *  3. `indexing` older than the timeout budget → `failed:indexing_timeout`
  *     (retryable) — the sweep never leaves an operation in limbo forever.
+ *  4. `ready_to_sign` older than its budget → `failed:signer_failed`
+ *     (retryable). Unlike the three above this one has no transaction to
+ *     reconcile: the pipeline broadcasts only after the transition out of
+ *     `ready_to_sign` commits, so an operation that died in this window never
+ *     reached an RPC. The row is stale state rather than a lost payment, which
+ *     is what makes failing it outright safe — and retryable honest, since
+ *     rebuilding it cannot duplicate anything.
```
nbaztec: can we make these explainers succint.

apps/sdp-web/src/app/dashboard/helius-rings/operation-composer.tsx
TarekkMA: Same as the previous React Doctor warning on this line: this is a two-item allowlist, so a second pass is not a real cost. Leaving the chain as-is.

apps/sdp-api/src/types/env.d.ts
```diff
@@ -229,9 +229,23 @@ export interface Env {
...
+  // Permits plain-http Rings upstreams. The public devnet indexer and prover
+  // are http on a real host, and over plaintext an indexer response reveals
+  // which notes an identity owns — so this is opt-in per environment rather
+  // than inferred from the URL.
+  HELIUS_RINGS_ALLOW_INSECURE_HTTP?: string;
+
+  // Base64 32-byte master seed the deterministic key authority derives every
+  // shielded identity from. Permanent once any wallet is provisioned: SDP
+  // refuses to re-key a registered identity, so changing this orphans its notes.
+  HELIUS_RINGS_DETERMINISTIC_KA_SEED?: string;
```
nbaztec: We don't need a seed in env for a mock in-memory KA if it's temporary (unless good reasons to do so). Just print a warning that in-memory KA is in use and use some `const` seed.

apps/sdp-api/src/types/env.d.ts
```diff
@@ -229,9 +229,23 @@ export interface Env {
...
+  // missing, so a misconfiguration is diagnosable rather than silent.
+  HELIUS_RINGS_RPC_URL?: string;
+  HELIUS_RINGS_INDEXER_URL?: string;
+  HELIUS_RINGS_PROVER_URL?: string;
+
+  // Permits plain-http Rings upstreams. The public devnet indexer and prover
+  // are http on a real host, and over plaintext an indexer response reveals
+  // which notes an identity owns — so this is opt-in per environment rather
+  // than inferred from the URL.
+  HELIUS_RINGS_ALLOW_INSECURE_HTTP?: string;
```
nbaztec, line 241: We shouldn't concern ourselves with http/https in this version

apps/sdp-api/src/types/env.d.ts
```diff
@@ -229,9 +229,23 @@ export interface Env {
...
 
-  // Rings gateway selector. Only "http" activates the live adapter and the
-  // indexing-poll job; anything else keeps NotImplementedRingsGateway.
-  HELIUS_RINGS_ADAPTER?: string;
+  // Rings upstreams, all required once Rings is enabled. Absence does not
+  // disable the gateway: it reports every component red naming what is
+  // missing, so a misconfiguration is diagnosable rather than silent.
+  HELIUS_RINGS_RPC_URL?: string;
+  HELIUS_RINGS_INDEXER_URL?: string;
+  HELIUS_RINGS_PROVER_URL?: string;
```
nbaztec, line 237: This is good for now, we can think later if we should put it in DB config (like SPC) at a later stage

packages/sdp-helius-rings-sdk/src/test/shielded-identity-fixtures.ts
```diff
@@ -0,0 +1,126 @@
+import { USER_REGISTRY_PROGRAM_ID } from "@heliuslabs/zolana";
```
nbaztec, line 6: Please name this file something other than `support.ts` to identify what it does.

packages/sdp-helius-rings-sdk/src/identity.ts
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * The honest caveat: this is a pure read *with respect to the chain* — one
+ * account read, no transaction built, nothing signed, no fee — but it is not a
+ * read with respect to the seed. Deciding "ours" versus "foreign" means having
+ * the derived keys to compare against, so the material is derived in process
+ * exactly as `syncRingsWallet` already derives it. It is precisely as
+ * seed-dependent as a sync, and no less.
+ *
+ * It exists because an operator who hits a provisioning conflict otherwise has
+ * no way to see what is actually published short of decoding the PDA by hand.
```
nbaztec: Comment too verbose

packages/sdp-helius-rings-sdk/src/material.ts
```diff
@@ -0,0 +1,222 @@
...
+
+/** A nullifier secret is 31 bytes, which is always inside the BN254 field. */
+export const NULLIFIER_KEY_BYTE_LENGTH = 31;
+
+/**
+ * Identifies the wallet a {@link ShieldedMaterialSource} should produce material
+ * for. The tenant triple is how SDP addresses a wallet; how a source turns that
+ * into key material is the source's own business.
+ */
+export interface MaterialRequest {
```
nbaztec, line 21: What's a `Material` here?

TarekkMA: it's the key material, this request is passed to the KA to fetch it

apps/sdp-api/src/types/env.d.ts
TarekkMA: yes didn't want to add that in this PR

apps/sdp-api/src/routes/helius-rings/handlers.ts
TarekkMA: will change the language so that it doesn't state what is not there

apps/sdp-api/src/lib/feature-flags.ts
TarekkMA: `allowInsecureHttp` is a parameter passed to `createZolanaClient` so I couldn't just hardcode it to yes, that's why I think the env var should be kept

apps/sdp-api/src/types/env.d.ts
TarekkMA: I would rather keep the seed private in the secrets so we don't expose any of the devnet wallets

apps/sdp-api/src/services/jobs/poll-rings-indexing.ts
```diff
@@ -59,38 +54,36 @@ export async function pollRingsIndexing(
...
-    // polls in one call, which is the only thing that rescues a broadcast whose
-    // indexing transition never committed. The timeout is deliberately not
-    // applied to it — the budget measures how long Photon has been asked, and a
-    // resumed operation has not been asked yet.
-    if (operation.state !== "indexing" && operation.state !== "submitted") continue;
     try {
+      // Aged out rather than polled: nothing was ever broadcast for this row.
+      if (operation.state === "ready_to_sign") {
+        if (Date.parse(operation.updated_at) < signingCutoff) {
+          await failSigningTimeout(repository, operation);
```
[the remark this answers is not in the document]
TarekkMA: Thanks. the mechanism is right, but I don't think the trigger is reachable.

Custody signing doesn't block indefinitely on an approval. The Fireblocks signer polls a bounded number of times and then throws: DEFAULT_MAX_POLL_ATTEMPTS = 60 at DEFAULT_POLL_INTERVAL_MS = 1000 in @solana/keychain-fireblocks, so the call gives up after ~60s. Utila is the same, DFNS is ~30s. Against a ten-minute RINGS_SIGNING_TIMEOUT_MS, the signer has already failed nine minutes before the sweep fires. The only way in is an org raising maxPollAttempts above 600, which our custody config accepts because it validates only "positive integer" — worth tightening, but separate from this sweep.

And if it did happen: the broadcast is gated on the compare-and-swap out of ready_to_sign, so nothing is broadcast twice, the discarded signature is recorded as a signature.discarded event, and the failure is retryable: true. A second approval would be needed, but no funds move twice and nothing is lost silently.

nbaztec (approved):

PR #1506, opened 2026-08-27
ci(smoke): canary with prod leg, Loki-reported for consecutive-failure alerting

niks3089:
## What

Turns the scheduled smoke into a proper canary per review direction: failures alert through the canary path (Loki → Grafana → Slack), not deployment notifications, and prod gets probed too.

- **Loki reporting**: every leg pushes a `sdp_smoke_run` summary line (outcome, `failed_any`, run url) to Grafana Cloud Loki via new `scripts/report-smoke-to-loki.sh`, stream-labeled `service=sdp-smoke, target=dev|prod` — the tokens-repo canary pattern. The companion sdp-infra rule alerts on **2+ failures in 70m per target** (two consecutive 30-minute runs), so one transient never pages, and Grafana posts `[RESOLVED]` on recovery.
- **Deploy-style Slack removed from the smoke**: `notify-slack.sh` here posted smoke failures to the deploy channel — wrong channel, no dedup, no recovery. Deploy workflows keep their deploy notifications.
- **Prod canary leg** (schedule + dispatch only, not the release gate): unauthenticated content probes of what prod can answer today — `platform.solana.com` up with the dashboard title marker, `api.solana.com/health` returning `status:ok, environment:production`. The authed fixture flow against prod needs seeded fixtures (no e2e user exists in prod Clerk) — tracked in PRO-1763; this leg already catches the "platform down during a demo" class.
- Dev leg unchanged (full authed smoke, still the prod-promotion gate via `workflow_call`).

## Verified

- actionlint clean; `pnpm test:scripts` 103 pass.
- `report-smoke-to-loki.sh` executed live against Grafana Cloud Loki with dev_ci creds: push accepted.
- Prod probes executed live from this machine: dashboard title present, health ok.
- A branch `workflow_dispatch` validation run follows in the PR checks.

Part of PRO-1760 / PRO-1763.


multipletwigs:
thanks for the pr! what's the best way for the smoke test to handle feature flagged ui? currently the tests will fall back to default flag values without the vercel creds. Should we use a deterministic flag configuration for the smoke user, or provide credentials so the test exercises live dev targeting?

niks3089:
@multipletwigs went with the deterministic option, using the mechanism the app already has: every flag's `defaultValue` reads an `SDP_FLAG_*` / `*_ENABLED` env override, so the workflow now pins all nine dashboard flags explicitly in its job env (and preserves them past Doppler). The gate's flag posture is a reviewable one-line diff per flag instead of a silent code fallback — flip any of them if the pinned posture should track a dashboard change. No creds, no remote dependency on the release gate. Live Vercel flag targeting is still exercised by the prod canary (sdp-infra#89), which drives the deployed dashboard with the real flag config. Validation dispatch running now.

resourcefulmind (approved):

PR #1508, opened 2026-08-27
fix(web): expand a side-menu section when its top-level item is followed

tobySolutions:
Following Payments in the side nav took you to Payments and left its submenu shut, so the pages you had just asked for stayed hidden behind a separate chevron click.

The top-level item opens its section now. It only ever opens, never toggles: a second click on the section you are already inside would collapse the submenu of the page you are looking at. The chevron is still the control that closes one, and that choice still persists.

The state transition moved into `dashboard-nav` beside the rest of the subnav vocabulary, so "opens, never closes" is pinned by tests rather than living inline in a component too large to render in one.

Two things came along with it. The subnav choice is now persisted outside the state updater, since React may replay one and a `localStorage` write placed inside it runs more than once; the toggle had that shape before this change.

And one unrelated test fix: `privy-credential-form.unit.test.tsx` asserted on the alert and then queried the submit button synchronously, so it depended on the transition behind the button having already settled. It held locally and failed under CI load once this branch shifted the suite's timing. The sibling cases only survive because they type a correction first, which yields the ticks that one does not.

Screenshot 2026-08-27 at 17 00 23
Screenshot 2026-08-27 at 17 04 44


multipletwigs:
nice one tobi! how do i visit the main page now say in markets?

multipletwigs (approved):

GuiBibeau (approved):

tobySolutions:
> nice one tobi! how do i visit the main page now say in markets?

Hey! Thank you, you just click on the market nav button like before chief.

PR #1512, opened 2026-08-27
feat(earn): translate vault simulation errors into readable verdicts

keeganthomp:
The deposit/withdraw modal surfaced Solana's raw `TransactionError` verbatim: a 0-SOL wallet depositing on dev showed `Vault deposit simulation failed: "AccountNotFound"`, which reads as a mystery instead of "your wallet has no SOL".

New `describeVaultSimulationError` maps the verdict to fee-mode-aware prose, e.g. wallet-pays `AccountNotFound` becomes "the wallet holds no SOL, so it cannot pay the network fee. Send SOL to the wallet and retry (AccountNotFound)". Also covers `InsufficientFundsForFee`, `InstructionError` (custom codes and common string variants), `InsufficientFundsForRent`, blockhash/already-processed, with unknown variants humanized or passed through raw.

**Reviewer notes:**

- The raw variant stays in parentheses so operator log searches and `failure_reason` greps still hit.

- Wording splits on fee mode: under sponsorship a broke fee payer is SDP's problem, and telling the customer to fund their wallet would send them fixing the wrong thing.

- No behavior change beyond the error string; simulation flow, logging, and the `badRequest` shape are untouched.

Verified: 24 tests pass across the new unit suite and the extended `vault-execution.service.test.ts`; typecheck clean for the touched files (pre-existing `helius-rings` errors unrelated).


cutbow7 (changes requested):
## Review: `earn-readable-vault-simulation-errors`

Nice direction — the fee-mode-aware attribution is genuinely the right call, and keeping the raw variant for log searches is thoughtful. A few real problems though, two of which undercut the PR's stated goal.

### Should fix

**1. The default catch-all produces broken English for most unmapped variants** — `vault-simulation-error.ts:93`

```ts
return `the transaction ${humanizeVariantName(err)} (${err})`;
```

Most `TransactionError` string variants are noun phrases, not verb phrases: `AccountInUse` → *"the transaction account in use (AccountInUse)"*, `SanitizeFailure` → *"the transaction sanitize failure"*, `TooManyAccountLocks` → *"the transaction too many account locks"*. The one variant the test picks (`WouldExceedMaxAccountCostLimit`) happens to start with a verb and reads fine, so the suite certifies the branch as working. Suggest `the transaction failed with "${humanizeVariantName(err)}" (${err})` or similar phrasing that works for nouns.

**2. Sponsored-mode failures self-declare an SDP fault but still surface as HTTP 400** — `vault-intent-execution.service.ts:80` + `vault-simulation-error.ts:80,84`

The new sponsored copy says *"This is an SDP configuration problem, not a problem with the wallet"*, but `executeSignedVaultIntent` still wraps it in `badRequest(...)`. Two consequences:

- Client retry middleware treats 4xx as permanent caller fault and abandons the deposit; SDP's 5xx SLO stays green while 100% of sponsored deposits fail. `simulateVaultPlan` itself documents the convention for the adjacent lookup-table case ("return a retryable 5xx instead of a caller-fault 400").
- SDP's internal sponsor funding state becomes an externally observable, pollable signal in a 400 body.

Sponsorship is devnet-only today, which softens this — but the message and status code now actively contradict each other, and this is the code path that goes to mainnet later.

**3. The sponsored `AccountNotFound` copy misdiagnoses a drained sponsor** — `vault-simulation-error.ts:80`

This file's own governing doc (`routes/earn/CLAUDE.md`) pins the meaning of this variant: *"a zero-SOL wallet simulated as its own fee payer dies with `AccountNotFound` and no logs"*. A sponsor that runs dry produces exactly this — but the sponsored branch says *"account does not exist on this cluster... This is an SDP configuration problem"*, routing the on-call operator toward Kora/allowlist config instead of topping up the sponsor. The wallet-pays branch gets the doc-consistent wording ("holds no SOL"); the sponsored branch should too.

**4. Unguarded `account_index` can render "account undefined"** — `vault-simulation-error.ts:108`

```ts
const accountIndex = (rent as Record<string, unknown>).account_index;
return `account ${String(accountIndex)} would be left below the rent-exempt minimum; ...`;
```

If the field is absent or shaped differently (RPC-provider/kit drift; the guard also passes for arrays), the user sees *"account undefined would be left below the rent-exempt minimum"* — strictly worse than the raw JSON this replaced. Guard on `typeof accountIndex === "number" || typeof accountIndex === "bigint"` and fall through to `raw` otherwise.

**5. The same modal still shows raw JSON for the failure the customer actually paid for** — `services/jobs/reconcile-earn-vault-movements.ts:132`

`failMovement(ledger, movement, JSON.stringify(status.err))` writes `failure_reason`, which the deposit/withdraw modal renders verbatim. So preflight rejections now read as prose while a landed-and-failed deposit — the one that cost the customer a fee — still shows `{"InstructionError":[0,{"Custom":6001}]}`. Worse, this `JSON.stringify` has no bigint replacer, so a bigint-carrying `err` throws inside the reconciler tick and leaves the movement stuck instead of `failed`. Same pattern at `track-pending-withdrawals.ts:274`. Worth either routing through the new helper (see next point about its signature) or at least `safeStringify`.

### Architecture / reuse

**6. This forks `describeTransactionErr` with the opposite policy and no cross-reference** — `services/private-channels/tx-error.ts:82`

The existing helper for the identical `TransactionError` shape documents a deliberate decision: *"Kept verbatim rather than mapped to prose: the operator reading `failure_reason` needs the real variant"* — and caps output at 2000 chars. The new module adopts the reverse policy with no length cap (the `stringifyRaw` fallback lands unbounded in the 400 body and error log). If earn deliberately diverges, both doc comments should say so; and the required `fee: Pick<VaultFeeMode, "kind">` param (only 2 of ~16 variants use it) is what keeps this decoder trapped in the earn folder — an optional attribution hint would let the reconciler and other call sites reuse it.

**7. `stringifyRaw` is roughly the seventh copy of the bigint replacer in the repo** — `vault-simulation-error.ts:19`

`safeStringify`/`bigIntReplacer` are already exported from `@sdp/solana` (`packages/sdp-solana/src/token-2022.utils.ts`), which sdp-api depends on and already imports in tests. The wrapper here can be one line over the shared helper.

**8. Hardcoded English prose bypasses the dashboard's five-locale i18n** — `vault-simulation-error.ts:76-93`

The dashboard ships en/es/fr/pt/vi and the old value was at least a locale-neutral token. The new sentence is free text composed server-side, so a Spanish-locale user gets one English sentence in an otherwise translated modal, and there's no stable code for the web layer to translate against (compare `apps/sdp-web/src/lib/policy-denial-reason.ts`, which returns a code and maps it client-side). Fine to accept as a known tradeoff, but worth an explicit decision.

### Minor

- **`CLAUDE.md` overclaims** (`routes/earn/CLAUDE.md:448-451`): "raw variant kept in parentheses" — the unrecognized-shape fallback returns bare JSON with no prose or parentheses, pinned by this branch's own test. Also `(vault-simulation-error.ts)` reads as a `routes/earn/` file per this doc's own conventions; qualify it as `services/earn/vault-simulation-error.ts`.
- **Inconsistent raw rendering**: `const raw = stringifyRaw(err)` is computed eagerly but unused on the whole string path, which interpolates the bare variant — so string errors render unquoted (`(AccountNotFound)`) while object errors render quoted JSON. Saved log greps against the old `"AccountNotFound"` (quoted) format silently stop matching. Move `raw` into the object branch.
- **Dead cases**: `BlockhashNotFound` and `AlreadyProcessed` are unreachable from this call site — `simulateVaultPlan` simulates with `replaceRecentBlockhash: true`.
- **`INSTRUCTION_ERROR_DETAILS[detail]`** indexes a plain object, so prototype keys (`"constructor"`, `"toString"`) resolve to inherited functions and skip the `??` fallback. Use `Object.hasOwn` or a `Map`. Only reachable via a hostile/buggy RPC, but `err` is `unknown` and BYOK RPC URLs exist.
- **`stringifyRaw` can return `undefined`** (JSON.stringify of `undefined`/symbol/function returns `undefined`, which the try/catch doesn't cover), violating the declared `string` return. `JSON.stringify(...) ?? String(err)` covers it in one line.
- **`instruction ${index}`** renders the 0-based index as human prose; readers counting from one will inspect the wrong instruction of a multi-instruction plan. Consider "instruction #1 (0-based)" or the ordinal.
- **`feePayerNoun(fee)`** is only ever called inside branches that already tested `fee.kind`, so each call site has one statically dead arm; hoisting `const noun = ...` and `const remedy = ...` once would collapse the four near-duplicate templates to two.

keeganthomp:
All addressed in 00043a2f4 except the i18n point (8), which needs a product call.

| # | Fix |
| -- | -- |
| 1 | Catch-all is now `the transaction failed with "account in use" (AccountInUse)`, and the test covers a noun variant |
| 2 | The helper returns a `fault` alongside the message; a sponsor-fault simulation now throws `internalError` (5xx) with a generic body, the detailed reason stays in the log line. No sponsor funding state in a 400 |
| 3 | Sponsored copy now says the sponsor holds no SOL, matching the wallet-pays wording and the doc's pinned meaning |
| 4 | `account_index` guarded to number/bigint, anything else falls through to raw |
| 5 | Earn reconciler routes `failure_reason` through the helper (fee param is optional now, neutral wording there). The private-channels tracker keeps its verbatim policy via `describeTransactionErr`, whose internal stringify is now bigint-safe instead of degrading to `[object Object]` |
| 6 | Both doc comments cross-reference the opposite policies; raw payload capped at 2000 like tx-error.ts |
| 7 | Both modules now use `safeStringify` from `@sdp/solana` |

Minors: CLAUDE.md path and claim corrected; string variants now render the quoted raw (`("AccountNotFound")`) so old greps keep matching; instruction details use a Map (no prototype keys); `stringifyRaw` can't return undefined; index phrased as "instruction at index 1"; noun/remedy/fault hoisted once.

Kept `BlockhashNotFound`/`AlreadyProcessed`: the helper is now also the reconciler's decoder for landed errors, so I left it as a generic TransactionError decoder rather than pruning to one call site's reachable set.

On (8), i18n: accepted as a known tradeoff for now. The raw variant in parentheses gives the web layer a stable token to key translations on later, and moving to a code + client-side mapping (policy-denial-reason style) is a straightforward follow-up if we want the modal fully localized. Happy to file it.

cutbow7 (approved):

PR #1513, opened 2026-08-27
feat(payments): MoonPay lifecycle, transferId polling, counterparty provider links

multipletwigs:
**MoonPay lifecycle**

- Ramp quote responses now return `transferId` alongside the quote; the wizard polls `GET /v1/payments/transfers/{id}` instead of re-POSTing `/quote` on a 4s interval (each retry minted a fresh provider quote and burned the `ramp-quote` metered quota) and instead of re-finding its own row through the `providerReference` list filter.
- New `counterparty_provider_accounts` table links a counterparty to its provider-side customer (`provider_customer_reference` = MoonPay `customerId`), upserted **in the same transaction** as the settlement status transition. **First-write-wins**: the first customer seen is canonical; a later event reporting a different customer appends to `metadata.mismatchedReferences` and logs `sdp_api_counterparty_provider_reference_mismatch` instead of overwriting. 

To solve the above we need moonpay headless - this is just a temporary bridge. Groundwork for PRO-1752 and MoonPay headless sessions (headless later enforces the pairing on MoonPay's side).

- A refused settlement event (lost race, out-of-order redelivery) has **no effects**: no status write, no customer-link write, no workflow trigger. `onramp_settled`/`offramp_settled` fire only when the settled transition actually landed.
- `expired` is revivable by provider events proving checkout activity (signed widget URLs never lapse), while `completed`/`failed`/`canceled` stay terminal.
- Providers now receive `counterparty.id` as `externalCustomerId` (never `external_id`); MoonPay webhook bodies and settlement economics are zod-parsed (`moonpayBuyTransactionSchema`, economics schema) instead of cast/hand-checked; mapping lives in `@sdp/payments/ramps/providers/moonpay/settlement.ts`.
- Web: quote creation is one-shot with an error panel + retry (no more silent `catch {}`); all payments SWR keys move to a typed key factory (`payments-query-key.ts`) with object params.

Deliberate scope cut: webhooks are the sole settlement authority in this PR. Reconciliation (pulling MoonPay's transaction listing to heal lost webhooks and expire abandoned checkouts) comes as its own PR — the agreed design is a single account-wide listing sweep rather than per-transfer lookups.

**UI Changes**
CleanShot 2026-08-28 at 12 34 00@2x

CleanShot 2026-08-28 at 12 35 26@2x

**dark mode widget support**
CleanShot 2026-08-28 at 12 45 12@2x

**before** — MoonPay deposit lifecycle:

1. Wizard POSTs `/ramps/onramp/quote` every 4s until one succeeds, errors swallowed; transfer row found back via list filter on `providerReference`.
2. Webhook settles in the background; MoonPay's `customerId` is discarded; a settled event that lost a status race still fired settlement workflows.
3. Terminal statuses were absolute — a late provider event after `expired` was dropped forever.

**after**:

1. One quote POST persists the transfer and returns its id; the wizard polls the transfer resource until terminal.
2. Webhook settlement atomically records status + economics + the counterparty↔MoonPay-customer link; workflow triggers fire only on applied transitions (eventKey dedup unchanged).
3. `expired` revives on provider-proved activity; other terminal states are unchanged.

## Settlement write path

```text
webhook → verify signature → zod-parse transaction → moonpayTransactionSettlementEvent
  → applyRampSettlementEvent (services/payments/ramp-settlements.ts, env-based):
      one tx: guarded status transition (+ economics) + counterparty_provider_accounts upsert
      emit onramp_settled/offramp_settled ⇐ only if the transition landed
```

**Verification**

- `pnpm -C apps/sdp-api typecheck && pnpm -C apps/sdp-web tsc --noEmit && pnpm -C packages/sdp-payments typecheck` — clean
- `pnpm -C apps/sdp-api vitest run` — green (1 pre-existing local flake: `kv-redis.node.test.ts` vs shared local Redis; passes in isolation)
- `pnpm -C apps/sdp-web vitest run src/app/dashboard/payments` — 115 passed
- Live E2E against local stack: counterparty → quote (`transferId` + signed sandbox widget URL) → transfer `pending` → signed webhook → `completed` + delivered amount + `counterparty_provider_accounts` row
- New tests: provider-link upsert + redelivery stability, expired revival, refused-transition emits nothing, race emits at most one settled trigger

Known gaps (deliberate, follow-up PRs)
- Reconciliation: lost webhooks and abandoned checkouts are not yet healed/expired — next PR (account-wide listing sweep + 1h expiry, design agreed).
- `cryptoTransactionId` captured in `provider_data.settlement` but not yet promoted to `signature` / on-chain finality poll.
- Workflow event delivery is at-most-once by decision: the settled-trigger dispatch is detached fire-and-forget. Safe on the Node-only deployment (the process outlives the response); a Workers-style runtime would drop triggers torn down mid-dispatch. The planned rework is a durable outbox written inside the settlement transaction — deferred to the workflows rework, not patched here.
- MoonPay `allowedIpAddress` (documented as mandatory for production widgets) not yet sent.


GuiBibeau (approved):

tobySolutions (approved):
LGTM

PR #1520, opened 2026-08-28
feat(ci): ephemeral per-PR API environments on dev (PRO-1767)

niks3089:
Implements the ephemeral per-PR dev API design locked today (see PRO-1767 comment).

**How it works**
- Label a PR `ephemeral-api` → builds the branch image and stands up `sdp-dev-api-pr-<n>` (public run.app URL, scale 0–2) + `sdp-dev-worker-pr-<n>` (internal, crons enabled), cloned live from the real dev service manifests so config never drifts from terraform.
- Isolation is two env overrides applied at boot (`getProcessEnv`): `EPHEMERAL_DB_NAME` re-points `DATABASE_URL` at `sdp_api_pr_<n>` on the dev Cloud SQL instance; `EPHEMERAL_REDIS_DB` re-points `REDIS_URL` at a logical db (1–15) claimed collision-free via a `sdp-ephemeral-redis-db` label on the PR's resources — offset scan for a free slot, then a post-claim verify where the lower-numbered PR wins, so no two live environments ever share a db (hard cap: 15 concurrent). Index 0 stays dev/main.
- A per-PR clone of the migrate job runs the new bundled `ephemeral-db.js` in-VPC: `ensure` creates the database and applies the branch's migrations from zero; `drop` (args override at teardown) terminates connections, drops it, and flushes the redis db.
- The URL is commented on the PR (updated in place per push). Teardown on close/unlabel; a nightly reaper sweeps environments whose PR closed while the event was missed.

**Coexistence with dev/main**: `sdp-dev-api-public`/`api-dev.solana.com`, its database, and redis db 0 are untouched; ephemerals share only instance-level capacity. Not behind the LB → no Cloud Armor; app-level Clerk auth still applies (dev-tier tradeoff).

**Depends on** sdp-infra#94 (project-level `run.admin` for the dev deployer — dev terraform auto-applies on merge).

**Rollout note**: first `ensure` run requires the dev app DB role to have `CREATEDB`; if it doesn't, one-time `ALTER ROLE <app_user> CREATEDB` on the dev instance.

**Security posture** (Greptile round): the privileged orchestration script (`ephemeral-env.sh`) always runs from a trusted `main` checkout — the PR checkout is used only as the Docker build context; teardown checks out `main` explicitly. The nightly reaper discovers environments by label on both services *and* jobs, so a deploy that dies between the db job and the API service still gets reaped once the PR closes.

### DIFF
--- a/.github/scripts/ephemeral-env.sh
+++ b/.github/scripts/ephemeral-env.sh
@@ -24,9 +24,49 @@ api_service="sdp-dev-api-pr-${pr}"
 worker_service="sdp-dev-worker-pr-${pr}"
 db_job="sdp-dev-api-pr-${pr}-db"
 db_name="sdp_api_pr_${pr}"
-redis_db=$(((pr % 15) + 1))
 
 run() { gcloud run "$@" --project "${PROJECT}" --region "${REGION}"; }
+
+list_redis_claims() { # prints "<pr> <db>" for every labeled job and service
+  local kind
+  for kind in jobs services; do
+    run "${kind}" list --filter 'metadata.labels.sdp-ephemeral-redis-db:*' \
+      --format 'value(metadata.labels.sdp-ephemeral-pr,metadata.labels.sdp-ephemeral-redis-db)'
+  done
+}
+
+# On a same-second race two PRs could pick the same slot; the offset scan makes
+# that need pr1 ≡ pr2 (mod 15), and verify_redis_claim (lower PR wins) turns the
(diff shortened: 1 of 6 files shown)


niks3089:
Restored the pre-existing deployment-mode tests in runtime-env.test.ts — the earlier commit had overwritten them instead of appending. The file now carries both suites.

multipletwigs (approved):

PR #1525, opened 2026-08-28
fix(api): run tsx through node so dev:local starts on Windows

unsafePtr:
`dev-local.mjs` spawns the tsx `.bin` shim, which on Windows is `tsx.cmd`, and spawning a `.cmd` without a shell has thrown `EINVAL` since Node's CVE-2024-27980 fix — so `pnpm dev` dies before the API server starts.

This resolves tsx's published `./cli` export and runs it with the current `node` binary, which drops the platform branch entirely rather than special-casing Windows.


multipletwigs (approved):
nice 👍

multipletwigs:
@unsafePtr i noticed that the commit isn't signed, could you repush with a signed commit? thank you! a hard requirement from sf

multipletwigs (approved):

PR #1526, opened 2026-08-28
fix(credentials): accept the canonical project number in GCP secret refs

tobySolutions:
Secret Manager canonicalizes resource names to the numeric project id. The store built and validated refs with the project id string, so both checks on the addVersion response failed (`credential-secret-store.ts:510` and `:352`). The version is created before that validation runs, so every credential write on a GCP backend threw `CredentialSecretStoreError` and left an orphaned version holding the customer's key.

That error is not in the mapped set at `app.ts:390-436`, which is why it surfaced as a bare "An internal error occurred".

Affects every caller of the store, not just RPC BYOK: custody submission and installation, workflow secrets. Broken since #884.

**Fix:** accept either spelling of the project, compare the parsed secret id instead of prefix-matching a ref built from the project id, and resolve the number from the metadata server (cached per process, since a store is constructed per request).

**The fixtures were the reason this stayed green** — they answered with the project id form. They now answer the way GCP does. Reverting just the project check fails the round-trip and metadata tests; confirmed by running.

Root cause found by Andrey from the Sentry issue.

HOO-1390

Not included, both worth their own change: purging the orphaned versions already on dev (needs GCP access), and mapping `CredentialSecretStoreError` in `app.ts` so this class of failure stops being opaque.


resourcefulmind (approved):
lgtm. Went through it after Andrey's Sentry find. 
It's nice that the fixtures now answer with the number, that was the whole reason this slipped. 
I tried to talk myself into "any digits now pass" and couldn't, the 9999999999 cases shut that down. I also checked that we already go to the metadata server for the token, so nothing new to reach in prod because it worried me too. Happy for the `app.ts`mapping to be its own PR.

PR #1529, opened 2026-08-28
refactor(web): consolidate SWR query keys into per-feature factories

multipletwigs:
**What changed**
- Every SWR cache key in the dashboard now comes from a colocated per-feature factory, mirroring the existing `payments-query-key.ts` pattern — no more inline string/array keys at call sites.
- New factories: `custody-query-key.ts` (3 keys), `issuance-query-key.ts` (2 keys), `markets/earn/earn-query-key.ts` (11 keys); 8 call-site files switched over.
- Local one-off helpers (`walletPolicyRevisionsKey`, `tokensCacheKey`, `TOKENS_SWR_KEY`) deleted in favor of the factories.
- No user-visible change; the work exists so a key and its `mutate()` invalidations can't drift apart across files.

**before** — keys built inline at each call site:

1. A hook builds `["dashboard-earn-program-deposits", programId]` inline; the surface that later invalidates it re-types the literal.
2. Two custody files share a key only via an exported helper living inside a drawer component.
3. A typo or drifted literal silently forks the cache — two entries, double fetch, stale invalidation.

**after** — keys resolved through one factory per feature:

1. Call sites request `earnQueryKeys.programDeposits({ programId })`; conditional gating (`cond ? key : null`) stays at the call site.
2. Read sites and `mutate()` sites reference the same factory entry, so invalidation can't miss.
3. Key values are unchanged except two custody keys normalized from template strings to `["prefix", id] as const` tuples — SWR keys are session-only cache identities, so no migration concerns.

**Verification**

- `pnpm --filter sdp-web exec vitest run src/app/dashboard/markets/earn src/app/dashboard/issuance src/app/dashboard/custody` — 51 files, 541 tests passed
- `npx tsc --noEmit` (apps/sdp-web) — clean
- `biome check` on all touched files — clean


keeganthomp (dismissed):
Nice.

Make sure CI passes, may just be flaky

h4rkl (approved):

PR #1535, opened 2026-08-28
feat(earn): register the wisdomtree provider

cutbow7:
Registers WisdomTree Connect as the sixth Earn provider id, end to end through every compiler-forced registry: surfacing (false — go-live flips last, per the pluggability playbook), deposit style (vault_direct: the on-receipt deposit wallet must never be presented as a fundable address, because attribution runs through the sending wallet's KYC registration), payout tokens (none), the @sdp/earn client registry (StubEarnClient scaffold, rwa/USDC envelope), and the API availability definitions (key-paired: one packed OAuth-quad JSON credential per environment).

Adds @sdp/types/wisdomtree-programs: the on-chain fund registry, measured against mainnet 2026-08-28 rather than copied from WisdomTree's docs — whose registry labels the mint accounts "Program" and carries at least one mislabeled row. WTGXX only, deliberately: the money market fund is the one instrument on their 15-fund shelf whose mechanics match a stablecoin yield facility. WisdomTree deploys on Solana mainnet ONLY (their sandbox is Ethereum Sepolia), so the per-cluster helpers answer empty for devnet instead of carrying a placeholder.


packages/sdp-types/src/provider-access.ts
```diff
@@ -39,7 +39,14 @@ export type RampProviderId = (typeof RAMP_PROVIDERS)[number];
...
  * excludes it rather than demanding a `KAMINO_API_KEY` that nothing reads.
  */
-export const EARN_PROVIDERS = ["veda", "upshift", "perena", "ground", "kamino"] as const;
+export const EARN_PROVIDERS = [
+  "veda",
+  "upshift",
+  "perena",
+  "ground",
+  "kamino",
+  "wisdomtree",
```
keeganthomp, line 48: [P1] Update the exhaustive provider test fixtures

Adding this provider changes the Earn availability map, but `provider-availability.service.test.ts` still expects the five-provider shape. The required Unit Tests job now fails with `wisdomtree` present only in the received value. Update the exhaustive expectations alongside this registry addition so this PR returns the suite to green.

keeganthomp (changes requested):
The provider registration leaves the required unit suite red. The inherited Turnkey failure is unrelated, but the Earn availability assertion deterministically fails because its exhaustive provider map was not updated.

PR #1536, opened 2026-08-28
feat(earn): wisdomtree catalogue client + deposit-eligibility capability

cutbow7:
Implements the SDK-free half of the WisdomTree integration in @sdp/earn:

- providers/wisdomtree/connect.ts — the Connect REST client (OAuth2 password grant with a per-credential token cache, products, on-receipt wallets, organization wallets, an orders tooling read). Wire shapes come from WisdomTree's published OpenAPI spec; SDP holds no credentials yet, so every field the spec leaves unprinted is marked UNVERIFIED and read through one constant or accessor so the live-measurement fix is one edit.
- listStrategies — production only (WisdomTree is Solana-mainnet-only; other environments answer [] and get browse rows via the PRO-1742 mirror). Identity comes from the measured on-chain registry in @sdp/types; availability comes from the products API, fail-closed on can_trade !== true; a malformed response throws so the sync skips the pass rather than delisting the shelf. No currentApy on purpose — the 7-day yield lives in a Fund Data credential SDP does not hold, and a missing rate beats a fabricated one.
- EarnDepositEligibilityProvider — a NEW optional capability: a provider- side eligibility check money-IN paths consult (never exits, per ADR 0002). Exists for regulated instruments whose transfer hook would fail settlement to an unverified wallet: the check turns 'your USDC left and nothing came back' into a refusal with WisdomTree's reason attached, BEFORE money moves. WisdomTree implements it over the Connect wallet registry, fail-closed on unregistered/unapproved/unknown statuses.

### DIFF
--- a/packages/sdp-earn/src/providers/wisdomtree/connect.ts
+++ b/packages/sdp-earn/src/providers/wisdomtree/connect.ts
@@ -1,0 +1,428 @@
+import type { EarnProviderId } from "@sdp/types/provider-access";
+import { providerNotConfigured, providerUnavailable, SdpEarnError } from "../../errors";
+import { providerFetchJson } from "../../fetch";
+import type { EarnRuntimeContext } from "../../types";
+
+/**
+ * WisdomTree Connect REST client — the HTTP half of the integration, shared by
+ * the catalogue client in this package and the instruction builders in
+ * `@sdp/wisdomtree` (which depends on this package, never the reverse).
+ *
+ * Plain `providerFetchJson` over OAuth2, no chain SDK — this module rides the
+ * hourly catalogue cron, so the @sdp/earn dependency rule (nothing heavier
+ * than @sdp/types) binds it.
+ *
+ * ── Wire shapes are documented, not yet measured ────────────────────────────
+ * Endpoint paths, parameter names and response fields below come from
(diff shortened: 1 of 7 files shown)


packages/sdp-earn/src/providers/wisdomtree/connect.ts
```diff
@@ -0,0 +1,371 @@
...
+
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(raw);
+  } catch {
+    throw providerNotConfigured(
+      `${keyName} is not valid JSON. Expected {"clientId","clientSecret","username","password"}.`
+    );
+  }
+  const record = parsed as Partial<Record<(typeof CREDENTIAL_FIELDS)[number], unknown>>;
```
keeganthomp, line 93: [P2] Reject non-object credential JSON cleanly

`JSON.parse("null")` succeeds, then `record[field]` throws a raw TypeError instead of the promised `PROVIDER_NOT_CONFIGURED` error. Guard that the parsed value is a non-null, non-array object before indexing so malformed packed credentials stay a classified configuration failure.

keeganthomp (dismissed):
No blocking issue is introduced by this delta. The stack still inherits #1535's provider-availability test failure; the parser edge case and existing review threads are non-blocking.

PR #1538, opened 2026-08-28
feat(wisdomtree): execution package — subscription plans and position…

cutbow7:
… reads

New workspace package @sdp/wisdomtree, mirroring @sdp/kamino's shape: the kit-native half of the integration, consuming @sdp/earn's provider contract and Connect REST client (now a package subpath export) with the arrow pointing strictly inward.

What a deposit IS here: there is no vault instruction. The plan is the on-chain leg of a primary-market subscription — TransferChecked of USDC to WisdomTree's on-receipt Purchase wallet, resolved from their API at build time (never configuration: a stale settlement address is money sent to the wrong place). Fund tokens settle back after NAV strike, outside the transaction, which is why the plan also creates the owner's Token-2022 fund ATA when it is measured absent (the createsShareAccount contract, same read-then-broadcast residual as Kamino's share ATA) — so WisdomTree's settlement leg has an account to land in.

Safety spine, in the package rather than in callers:
- verifyFundMint compares the LIVE mint (owner program, decimals, transfer-hook program) against the measured registry before any build — builder truth for a vaultless provider, refusing drifted instruments. Tested against the verbatim mainnet WTGXX account image.
- acceptAtMintScale refuses sub-atomic precision instead of flooring, and the canonical amount the instructions encode is what gets ledgered.
- assertPlanTargetsCluster re-checks every emitted plan against a closed per-cluster program allowlist, and sponsoredPrograms declares that SAME set to the paymaster harness (the Kamino no-drift rule).
- minSharesOut is refused outright: a NAV-settled subscription has no instruction that could enforce a share floor, and accepting the field while enforcing nothing is slippage protection in appearance only.
- Position reads are exact base units from the owner's ATA (never uiAmount); discovery drops empty holdings, explicit requests may truthfully answer zero, and devnet answers an empty page because no instrument exists there.

supportsVaultWithdraw stays FALSE on purpose — redemptions (the fund-token transfer through the compliance hook) are their own change, per the playbook's money-out-is-separate rule.

### DIFF
--- a/packages/sdp-wisdomtree/src/fixtures.test-helper.ts
+++ b/packages/sdp-wisdomtree/src/fixtures.test-helper.ts
@@ -1,0 +1,42 @@
+import { SPL_TOKEN_PROGRAMS } from "@sdp/types";
+import type { WisdomTreeChainReader } from "./chain";
+
+/**
+ * The LIVE WTGXX mint account, captured verbatim from mainnet-beta on
+ * 2026-08-28 (`getAccountInfo Em46fxx…`, base64). Real bytes on purpose: the
+ * mint parser and the build-time verification are tested against the exact
+ * account the production path will read, not a hand-built approximation.
+ */
+export const WTGXX_MINT_ACCOUNT_BASE64 =
+  // biome-ignore lint/security/noSecrets: a public on-chain account snapshot, not a credential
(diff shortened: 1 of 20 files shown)


packages/sdp-wisdomtree/src/fixtures.test-helper.ts
```diff
@@ -0,0 +1,42 @@
...
+ * mint parser and the build-time verification are tested against the exact
+ * account the production path will read, not a hand-built approximation.
+ */
+// biome-ignore lint/security/noSecrets: a public on-chain account snapshot, not a credential
+export const WTGXX_MINT_ACCOUNT_BASE64 =
+  "AQAAAB2xnEqSO8sZPGeQEhmQ3X+IzhAbQLpJPMnVWbA7sPSP8mss1wgAAAAJAQEAAABzwlqNUhBgDnyX+YoPa5Wi0md+4WM6nid/A/8VcEEz0wAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAARIAQAAPuJ2FTwKjm2ceUyKz9jasrbpf1dVq06bCcHiC7tAl/cxw/QLM7rMhcoPZgn6ttq9HO8Ub2ACWfUePx93M9ItcGQA4AA+4nYVPAqObZx5TIrP2Nqytul/V1WrTpsJweILu0CX9AAAAAAAA8D8AAAAAAAAAAAAAAAAAAPA/GgAhADPnFWqaIb56oewidAqLHfWwauBugPTClN7oWn2AHJOWAAwAIAAPuJ2FTwKjm2ceUyKz9jasrbpf1dVq06bCcHiC7tAl/Q4AQAAPuJ2FTwKjm2ceUyKz9jasrbpf1dVq06bCcHiC7tAl/dEFsuk4srWpLVcZMSkBnjwGNmWaz3RcDYmrDPu6DxFQEwDhAA+4nYVPAqObZx5TIrP2Nqytul/V1WrTpsJweILu0CX9zHD9AszusyFyg9mCfq22r0c7xRvYAJZ9R4/H3cz0i1wvAAAAV2lzZG9tVHJlZSBHb3Zlcm5tZW50IE1vbmV5IE1hcmtldCBEaWdpdGFsIEZ1bmQFAAAAV1RHWFhdAAAAaHR0cHM6Ly9nYXRld2F5LnBpbmF0YS5jbG91ZC9pcGZzL2JhZmtyZWlobm5oeWx6aGtjaXV3aWs1aG4yNW1nbGFpZ3hsc2plcXZ3NWxtcG1wdTNpb3lvZTMyeTc0AAAAAA==";
```
keeganthomp, line 12: [P1] Restore the required lint check

Biome flags this high-entropy literal because the ignore above the declaration does not cover this expression; `mint.ts` also flags the base58 alphabet, and several new ignores are reported unused. The required Lint job exits 123. Put suppressions on the reported expressions and remove the unused ignores so this PR's package passes lint.

keeganthomp (changes requested):
This head introduces required-check failures. Lint fails in the new WisdomTree package, and CodeQL has an existing inline alert on the amount parser that also needs to clear.

PR #1539, opened 2026-08-28
feat(api): wire wisdomtree execution + the deposit-eligibility gate

cutbow7:
Composes the WisdomTree executing client into both provider maps — the execution registry branch (the ONE sanctioned provider-id mapping) and the API's singleton overlay registry — with the same proven-RPC resolver and per-operation deadline Kamino uses, and the portfolio-mutual- exclusion assertion at construction. Adds @sdp/wisdomtree to the API's dependency graph everywhere the graph is enforced: package.json, the Dockerfile's two COPY stages, the Dockerfile.dockerignore re-include allowlist (the build-context trap), module boundaries, and the dependency-boundary script (only the API may ship the package).

Introduces the MONEY-IN eligibility gate: assertVaultDepositEligible, capability-dispatched via supportsDepositEligibility and shared by the custody-signed deposit service and the external-wallet (B2B2C) build so the rule cannot drift between the two paths. It runs after the replay reads (replays stay pure durable reads) and before sponsorship resolution or any provider build — an ineligible wallet's deposit would settle nowhere, so it must not cost a paymaster reservation or an RPC build either. The refusal carries the provider's own reason, which for Embedded Yield is the sentence the partner UI shows a user who still needs to complete issuer verification. No exit path consults it (ADR 0002: money out never inherits a money-in gate) and a provider without the capability is untouched.

### DIFF
--- a/apps/sdp-api/Dockerfile
+++ b/apps/sdp-api/Dockerfile
@@ -32,6 +32,7 @@ COPY packages/sdp-private-channels/package.json ./packages/sdp-private-channels/
 COPY packages/sdp-spc-escrow/package.json ./packages/sdp-spc-escrow/
 COPY packages/sdp-spc-withdraw/package.json ./packages/sdp-spc-withdraw/
 COPY packages/sdp-kamino/package.json ./packages/sdp-kamino/
+COPY packages/sdp-wisdomtree/package.json ./packages/sdp-wisdomtree/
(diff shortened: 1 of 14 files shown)


keeganthomp (dismissed):
No blocking issue is introduced by the API wiring or eligibility ordering. The required Lint failure is inherited from #1538 and must clear in the stack base.

PR #1540, opened 2026-08-28
feat(wisdomtree): redemptions — the money-out half, through the compl…

cutbow7:
…iance hook

Implements buildVaultWithdrawal, flipping supportsVaultWithdraw to true (the PRO-1702 mechanism: implementing the capability is what opens the exit route). A redemption's on-chain leg is a Token-2022 TransferChecked of fund tokens to WisdomTree's on-receipt Sale wallet; USDC settles back to the owner's registered wallet after NAV strike, outside the transaction. Per ADR 0002 the path takes no surfacing, entitlement, or eligibility gate — a wallet holding the tokens proved its standing on-chain, and the hook re-proves it at execution.

The load-bearing piece is transfer-hook.ts: the standard SPL tlv-account-resolution algorithm over @solana/kit, because a transfer of a hooked mint is invalid without the hook's account set appended. The ExtraAccountMetaList is fetched live from the hook program's validation PDA and its entries — literal pubkeys and PDA recipes over literal / instruction-data / account-key / account-data seeds — are evaluated exactly as SPL's addExtraAccountMetasForExecute does, appended in interface order (extras, hook program, validation account). The execute discriminator is pinned to its sha256 derivation by test.

This resolver is also where WisdomTree's KYC model surfaces on the exit: a compliance account that does not exist for one of the wallets fails resolution as HOOK_UNRESOLVED with prose naming the cause, and anything the resolver cannot see fails at simulation when the hook refuses — both BEFORE money moves. A hook entry demanding an extra SIGNER is refused outright: the only signer this transfer carries is the owner.

No account is ever closed by a redemption (the fund ATA survives a full exit), so rentRefundTo is accepted and deliberately unused.

### DIFF
--- a/packages/sdp-wisdomtree/src/plan.ts
+++ b/packages/sdp-wisdomtree/src/plan.ts
@@ -2,7 +2,7 @@ import { SPL_TOKEN_PROGRAMS } from "@sdp/types";
 import type { WisdomTreeFund } from "@sdp/types/wisdomtree-programs";
 import { WISDOMTREE_TRANSFER_HOOK_PROGRAM_IDS } from "@sdp/types/wisdomtree-programs";
 import type { Address, Instruction, TransactionSigner } from "@solana/kit";
-import { address } from "@solana/kit";
+import { AccountRole, address } from "@solana/kit";
 import {
   findAssociatedTokenPda,
   getCreateAssociatedTokenIdempotentInstruction,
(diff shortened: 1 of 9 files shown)


packages/sdp-wisdomtree/src/plan.ts
```diff
@@ -172,3 +173,132 @@ export async function buildWisdomTreeDepositPlan(
...
+  });
+  const [onReceiptFundAta] = await findAssociatedTokenPda({
+    owner: input.onReceiptWallet,
+    tokenProgram: TOKEN_2022_PROGRAM,
+    mint: fundMint,
+  });
+
+  const instructions: Instruction[] = [];
+  const onReceiptFundAccount = await reader.getAccount(onReceiptFundAta);
+  if (onReceiptFundAccount === null) {
```
keeganthomp, line 240: [P1] Resolve hook seeds against the ATA this plan creates

The live WTGXX meta list derives compliance PDAs from destination token-account data. When WisdomTree returns a fresh Sale wallet, this branch queues its fund ATA creation, but `resolveTransferHookAccounts` immediately reads that still-missing ATA and throws `HOOK_UNRESOLVED`. Redemptions stay blocked until someone creates the ATA out of band. Resolve against projected Token-2022 account data for the planned ATA, or provision it before building the transfer.

keeganthomp (changes requested):
The money-out path cannot build against a fresh WisdomTree Sale wallet because hook resolution reads destination account data before the plan creates that account.

PR #1541, opened 2026-08-28
test(wisdomtree): mainnet-fork smoke proof + agent notes

cutbow7:
Adds the surfpool smoke test (env-gated, never in CI): against a surfnet forking mainnet — real WTGXX mint, real compliance hook, real ExtraAccountMetaList — it proves the two facts the offline suites cannot:

1. The subscription leg (USDC to the on-receipt wallet + the owner's fund-token ATA creation) simulates cleanly for any funded wallet.
2. An UNVERIFIED wallet's redemption fails at the KYC stage, and the test accepts exactly the two shapes that failure can honestly take: the hook resolver refusing (HOOK_UNRESOLVED) or the on-chain hook program rejecting the fully-resolved transfer.

Measured on the 2026-08-28 run: all three tests pass; the resolver handled WTGXX's live recipe set (a literal-seeded compliance-config PDA, two literal accounts, and two account-data-seeded PDAs on an external program keyed by the source/dest owners), and the hook program rejected the unverified wallet's transfer in simulation — the compliance model working end to end, with no mainnet money at risk.

Adds packages/sdp-wisdomtree/CLAUDE.md: what differs from the Kamino precedent (no vault; a deposit is a primary-market order's on-chain leg; minSharesOut refused; confirmed ≠ settled), the three-layer compliance spine, the UNVERIFIED wire fields to re-measure when Connect credentials arrive, and the go-live checklist — surfacing stays FALSE per the playbook until credentials + PRO-1703 allow a real end-to-end deposit.

### DIFF
--- a/packages/sdp-earn/CLAUDE.md
+++ b/packages/sdp-earn/CLAUDE.md
@@ -416,12 +416,13 @@ rates come from the same paged endpoint the catalogue uses.
   `resolveEarnProviderClient` — DB provider ids are open strings and MUST be
   resolved through this, never direct-indexed.
 - Optional capabilities so far: portfolio wallets, withdrawal approvals, live
-  metrics, vault-direct (deposit + read) and vault-withdraw. All are
-  method-presence guards in capabilities.ts and a provider may implement any
-  subset — Kamino has live metrics and vault-direct, Veda has vault-direct.
-  NOBODY implements vault-withdraw yet, which is a statement about SDP's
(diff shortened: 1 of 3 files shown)


keeganthomp (dismissed):
No new blocking production issue in this delta. The two confirmed P2 smoke-test weaknesses are already called out inline; the stack base #1540 still needs its withdrawal fix.

PR #1562, opened 2026-08-31
feat(earn): curated V1 Kamino shelf, TVL ranking, liquidity/provider columns

keeganthomp:
## What

Three Earn V1 shelf tickets, one per commit: [PRO-1727](https://linear.app/solana-fndn/issue/PRO-1727), [PRO-1732](https://linear.app/solana-fndn/issue/PRO-1732), [PRO-1721](https://linear.app/solana-fndn/issue/PRO-1721).

| Ticket | Change |
| -- | -- |
| PRO-1727 | `CURATED_VAULTS` populated (six BD-picked mainnet vaults + devnet equivalents, reason per address), PYUSD admitted to the deposit set, Jupiter Lend excluded |
| PRO-1732 | Catalogue ordered TVL-desc in SQL, NULLS LAST, `(created_at, id)` tiebreaker |
| PRO-1721 | Liquidity + Provider columns on both strategy tables |

## Reviewer notes

- **Jupiter Lend is a name term, not an address entry.** No such row exists on any registry today (verified live 2026-08-31), so there is no address to key on; the term list can only remove rows. A test pins the worst case (a Jupiter row squatting a curated address stays hidden).

- **Devnet picks are a judgment call**: devnet has no Sentora and no USDG vaults, so the sandbox shelf mirrors the production shelf's shape (house mirrors + the one real devnet-PYUSD vault). Easy to swap.

- Curation config moved to `handlers/curation.ts` so `earn.test.ts` can bypass the shipped picks (same pattern as the surfacing mock); the real config gets its own describe.

- Money paths untouched: nothing outside the strategy reads consumes `isHiddenStrategy`, so a curated-away vault still funds and exits.

## Verification

| Suite | Result |
| -- | -- |
| sdp-earn (node:test) | 152 pass |
| sdp-api earn routes + repo (vitest) | 178 pass |
| sdp-web workspaces + i18n (vitest) | 203 pass |
| typecheck (types/earn/api/web) | clean |

sdp-web `lint` crashes on every file (pre-existing eslint 10 / eslint-plugin-react incompat, verified on untouched files).


apps/sdp-web/src/app/dashboard/markets/earn/earn-market-presentation.tsx
```diff
@@ -49,6 +54,48 @@ export function sumDecimalStrings(values: readonly string[]): string | undefined
...
+ */
+const EARN_PROVIDER_LABELS = {
+  ground: "Ground",
+  kamino: "Kamino",
+  perena: "Perena",
+  upshift: "Upshift",
+  veda: "Veda",
+} as const satisfies Record<EarnProviderId, string>;
+
+export function earnProviderLabel(provider: string): string {
```
keeganthomp: Moved earnProviderLabel and earnStrategyLiquidityLabel (plus the label map) into earn-format.ts, the existing pure-formatter module, and pointed importers there — no non-component exports added to this file. 7d210de3b

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/sdp-web/src/app/dashboard/markets/earn/earn-market-presentation.tsx
```diff
@@ -49,6 +54,48 @@ export function sumDecimalStrings(values: readonly string[]): string | undefined
...
+}
+
+/**
+ * Liquidity term for a catalogue row. `liquidityTerm` is typed closed but the
+ * read model is open wire data, so a value outside the vocabulary returns
+ * undefined and the caller renders its placeholder: a term is never invented.
+ * A delayed row without a usable day count still says "Delayed"; the day count
+ * renders only when it is a positive integer.
+ */
+export function earnStrategyLiquidityLabel(
```
keeganthomp: Same fix as the L71 thread: both helpers now live in earn-format.ts. 7d210de3b

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

keeganthomp:
The red "Kora / Live Smoke" check is environmental, not this PR: the shared devnet Kora's daily budget is exhausted (`-32031 Usage limit exceeded: sdp:unscoped 1001/1000 per 86400s`), so the one ATA-create signing test fails for every branch today. Re-run the failed shard once the window rolls. Everything else in the run is green or unaffected.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

cutbow7 (changes requested):
## Review: curate the V1 Kamino shelf, admit PYUSD, exclude Jupiter Lend

**Verdict: LGTM with one robustness suggestion and one doc-generator fix.** No blockers found. Reviewed as `origin/main...575c73e88` (merge-base is origin/main, so this is the clean PR diff). Static review only — I did not execute the test suites.

### What I verified

- **Curated addresses are real and correctly picked.** All 6 mainnet addresses in `CURATED_VAULTS` exist in the regenerated `kamino.inventory.json` with the expected token per comment (both PYUSD picks are PYUSD; the Commodity Yield pick is the $32M **USDC** vault, and its documented near-name trap — the $0.90-AUM USDG twin at `DM5ECR…` — is really in the registry). All 12 mainnet+devnet addresses base58-decode to 32 bytes. Devnet addresses I could not cross-check offline (on-chain source), trusting the 2026-08-31 verification note.
- **PYUSD admission is coherent end to end.** `PYUSD` already has well-known mints on both clusters; the devnet on-chain path filters via `KAMINO_DEPOSIT_SYMBOLS = Set(EARN_DEPOSIT_TOKEN_SYMBOLS)`, so the `PyUSDC` devnet vault admits with no further change. Perena/Upshift stubs keep their own explicit 3-token envelopes — correct to leave untouched.
- **TVL ordering is total and injection-safe.** `ordering` is only ever a literal at both call sites; `DESC NULLS LAST` is needed and present (Postgres DESC defaults to NULLS FIRST); the `(created_at, id)` tail keeps OFFSET pages from repeating/skipping, and the repository test covers exactly the tie shapes (equal-TVL run, no-TVL run).
- **Curation split + test bypass design is sound.** `isHiddenStrategy` and the list route apply identical rules (allowlist scoped per provider in SQL, so one shelf's allowlist can't curate another); the `vi.mock` getter + `beforeEach` reset means only the shipped-curation describe sees the real lists; those tests read addresses from the config itself, so a BD re-pick moves them instead of breaking them. The term-vs-address stacking test (Jupiter name squatting a curated address) is a nice touch.
- **Money paths untouched.** `requireEarnStrategy`/`isHiddenStrategy` have no consumers outside the strategy read handlers; the vault deposit/withdraw route tests seed uncurated references and correctly need no bypass.
- **UI columns:** both tables' width percentages still sum to 100; `getByText("Kamino")` in the row tests can't false-match the identity subtitle (it renders as one `"USDC · Kamino"` text node); new i18n keys fall back to English for the locales the translation bot hasn't covered.

### Suggestion (robustness): guard the `tvlUsd` cast

`earn.repository.postgres.ts`:

```sql
(risk_metadata->>'tvlUsd')::numeric DESC NULLS LAST
```

`risk_metadata` is JSONB with only an object-typeof CHECK, and `EarnStrategyRiskMetadata` is an open bag (`[key: string]: unknown`). The "JSON number or absent, never any other type" invariant is enforced by convention in two provider clients, not by any schema. If any future writer ever lands a non-numeric `tvlUsd` (string `"12M"`, bool, object), the cast throws **at read time** and every `GET /strategies` for that environment 500s — one bad row bricks the whole shelf, and the failure is far from the write that caused it. Cheap immunization:

```sql
CASE WHEN jsonb_typeof(risk_metadata->'tvlUsd') = 'number'
     THEN (risk_metadata->>'tvlUsd')::numeric END DESC NULLS LAST
```

which also demotes a malformed value to "unsized" instead of an outage. Today's writers are well-behaved, so this is non-blocking — but the codebase's own comments elsewhere treat "read path must not trust the bag" as doctrine.

### Fix: inventory generator contradicts the PR

`apps/sdp-api/scripts/inventory-kamino-catalogue.ts` hardcodes the prose

> Earn V1 is a stablecoin facility (`EARN_DEPOSIT_TOKEN_SYMBOLS` = USDC, USDG, USDT)

so the regenerated `docs/earn/kamino-catalogue-inventory.md` now contradicts itself: its top catalogued row is a PYUSD vault while the census section says PYUSD is outside the deposit set ("what widening that union would put in reach"). Render the list from `EARN_DEPOSIT_TOKEN_SYMBOLS` so it can't drift again.

### Nits

- `en/dashboard-earn.json` now carries **two** liquidity vocabularies: the new `DashboardMarkets.liquidity` ("Delayed · 7 days") and the pre-existing `DashboardEarn.liquidity` ("T+{days}"), for which I found no consumer in `apps/sdp-web/src`. If it's dead, delete it; if it's alive somewhere I missed, the same term is now phrased two ways on different surfaces.
- `selectPage`'s `NEWEST_FIRST` default is now unused — both call sites pass an explicit ordering, but the doc comment still sells the default. Fine to keep for future lists; just noting the comment oversells it.
- Comment drift: the Sentora PYUSD pick says "$112M" while the same-day inventory records $109M. Harmless ("figures are the shelf as picked"), but the two claim the same verification date.
- Observation, no action: the 5-minute metrics refresh moves `tvlUsd`, so rows can shift between OFFSET page fetches. Inherent to OFFSET paging, shelf is ~25 rows, tiebreakers keep the order total — acceptable.

keeganthomp:
Thanks — took all of it. `d65f33d06`.

**The `tvlUsd` cast guard: took it, and you undersold it.** Applied your `CASE WHEN jsonb_typeof(...) = 'number'` verbatim. Agreed it is doctrine rather than paranoia, and the blast radius is worse than "one bad row": the malformed row need not even be visible to the caller, because `ORDER BY` runs over every row the `WHERE` clause admits — so a curated-away or paused row could 500 a shelf it does not itself appear on. Added a repository test seeding a string `"12M"` and a bool `tvlUsd`, asserting the read still succeeds with both demoted to unsized.

**Inventory generator: fixed.** It renders the list from `EARN_DEPOSIT_TOKEN_SYMBOLS.join(", ")` now, and I regenerated the doc, so the census says "USDC, USDG, USDT, PYUSD" and no longer contradicts its own top row.

**Dead liquidity vocabulary: deleted.** Confirmed no consumer in `apps/sdp-web/src` and no dynamically-built message keys, then removed `DashboardEarn.liquidity` from all five locales. Parity kept, so `check:i18n` is unchanged — its findings are all pre-existing in `src/flags/index.ts`.

**`NEWEST_FIRST` default: removed rather than re-worded.** `ordering` is a required parameter now. The two callers want genuinely different orders, which is exactly where an invisible default is the wrong answer; the constant stays as the shared tiebreaker tail.

**Figure drift: fixed at the root.** All six comments carry the same-day census figures (Sentora $109M, HY USDG $45M, Steakhouse USDC $19M, HY USDC $3.6M, Commodity $32M, and the twin as "under $1 of AUM, dropped by the TVL floor"). The header now states that only the ADDRESSES are authoritative and that AUM moves continuously with the metrics refresh, so the figures cannot quietly become a claim again.

**OFFSET vs metrics-refresh observation:** agreed, no action.

One thing you flagged as un-checkable offline — the devnet addresses. I re-read the devnet kvault program (`devkRng…`) just now: 21 vaults live, and all six picks are present with the expected names and mints (`PyUSDC` is genuinely devnet-PYUSD-denominated, which is what makes the PYUSD admission exercisable in sandbox).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

cutbow7 (dismissed):

multipletwigs (approved):

PR #1571, opened 2026-09-01
ci: authenticate CI and deploy workflows to Doppler via OIDC

G1de0n:
Converts every workflow off the static `DOPPLER_TOKEN_CI` secret to Doppler OIDC (HOO-442), following the pattern already running in sdp-infra's grafana-push and the private repo's CI.

## What changed
- Every Doppler-using job gains `id-token: write` and a `Doppler OIDC login` step after the CLI install; all `DOPPLER_TOKEN` env plumbing is removed.
- `CI_HAS_DOPPLER_TOKEN` is now derived from the event (`pull_request` from a fork = false) instead of secret presence — same fork-safe behavior, no secret probe.
- `DOPPLER_TOKEN_CI` is removed from `workflow_call` contracts and every call site (deploy.yml → dev deploy → smoke; release flow → prod deploy).
- vendor-asset-drift logs in only `if: failure()` — the login exists solely for the Slack alert.
- Inline `doppler secrets get`/`doppler run` calls get explicit `--scope=.`.
- Contract test added: no `DOPPLER_TOKEN_CI` remains in the prod deploy workflow and the OIDC login step is present. `pnpm test:scripts` 119/119.

## Required setup before merge
- Doppler service-account identity with OIDC rule: issuer `https://token.actions.githubusercontent.com`, audience `https://github.com/solana-foundation`, repository claim `solana-foundation/solana-developer-platform`; read access to project `solana-developer-platform`, config `dev_ci`.
- Repository variable `DOPPLER_CI_IDENTITY_ID` set to that identity's UUID.

After a clean run on main, the `DOPPLER_TOKEN_CI` repository secret can be revoked.

PR #1575, opened 2026-09-01
ci: unit tests on the self-hosted dev-VM runner, packages split out

niks3089:
## What

Splits the monolithic Unit Tests job and routes both legs to the self-hosted runner in the dev GCP project:

- **`test` ("Unit Tests")** — runs only `@sdp/api` (`TEST_WORKSPACE_SPLIT=api`).
- **`test_packages` ("Unit Tests (packages)")** — everything except `@sdp/api`/`@sdp/api-integration` (`TEST_WORKSPACE_SPLIT=rest`, still changed-scoped on PRs), keeping the scripts tests, SecretRef serialization, and UI flow matrix checks.
- Both legs: `runs-on` routes pushes and same-repo PRs to `[self-hosted, sdp-ci]` — the JIT ephemeral runner pool from sdp-infra#136 (label standardized per review there). **Fork PRs never reach self-hosted** — they route to `ubuntu-latest`, keeping the public-repo safety line and the existing fork-safe steps.

`run-workspace-tests.mjs` gains the `TEST_WORKSPACE_SPLIT` env (`api` | `rest`); unset preserves today's behavior exactly. Contract tests: 119 pass.

## Merge order

1. sdp-infra#136 merges and applies, its GitHub App gets installed, and the `sdp-ci` pool comes online (repo-level JIT runners — no org runner-group toggle involved).
2. Then this merges. Until the pool is live, these jobs would queue forever — do not merge first.
3. At merge: add "Unit Tests (packages)" to the required-check list alongside "Unit Tests".
4. After cutover: retire the interim dev-project VM `gha-runner-sdp-dev-1` (it never served a job — the org runner-group toggle was never enabled, which the repo-level JIT pool makes moot).

## Note

An earlier revision of this PR targeted a dedicated 2× n2d-standard-16 spot pool in `solana-developer-platform-ci` (label `sdp-ci`). That was superseded by the dev-project VM; the two spot VMs never registered (their registration secret has no versions) and are now stopped. The `terraform/envs/ci` stack in sdp-infra can be destroyed or revived later if the 8-core VM proves too small.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,8 +175,11 @@ jobs:
     name: Unit Tests
     runs-on: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork && 'ubuntu-latest' || fromJson('["self-hosted", "sdp-dev-vm"]') }}
     timeout-minutes: 40
+    permissions:
+      contents: read
+      id-token: write
     env:
-      CI_HAS_DOPPLER_TOKEN: ${{ secrets.DOPPLER_TOKEN_CI != '' }}
+      CI_HAS_DOPPLER_TOKEN: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
       TEST_WORKSPACE_SPLIT: api
       TEST_DATABASE_URL: postgresql://sdp:sdp@127.0.0.1:5432/sdp_test
       REDIS_URL: redis://127.0.0.1:6379
@@ -225,13 +228,11 @@ jobs:
       - name: Install
         run: pnpm install --frozen-lockfile
 
-      - name: Validate Doppler bootstrap secret
+      - name: Validate Doppler OIDC configuration
         if: env.CI_IS_FORK_PULL_REQUEST != 'true'
-        env:
-          DOPPLER_TOKEN: ${{ secrets.DOPPLER_TOKEN_CI }}
         run: |
-          if [ -z "${DOPPLER_TOKEN}" ]; then
-            echo "DOPPLER_TOKEN_CI must be configured for secret-aware CI jobs." >&2
+          if [ -z "${{ vars.DOPPLER_CI_IDENTITY_ID }}" ]; then
+            echo "DOPPLER_CI_IDENTITY_ID must be configured as a repository variable for secret-aware CI jobs." >&2
             exit 1
           fi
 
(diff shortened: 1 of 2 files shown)


G1de0n (changes requested):
Nice speedup, and the api/rest split in run-workspace-tests.mjs is clean. Two things need fixing before this lands though, both easy:

Rebase onto the OIDC world. Branch predates #1571: the new job code reintroduces DOPPLER_TOKEN: ${{ secrets.DOPPLER_TOKEN_CI }}, the bootstrap-secret validation step, and the secret-probe CI_HAS_DOPPLER_TOKEN, and that repo secret got deleted yesterday (HOO-442 closed). As written, every same-repo run fails at validate. Current pattern on main is the Doppler OIDC login step plus event-based CI_HAS_DOPPLER_TOKEN plus permissions: id-token: write (this diff drops that too), just copy it from the merged test job.
Label mismatch with sdp-infra#115. Workflow targets [self-hosted, sdp-dev-vm] but the terraform registers runners with the sdp-ci label, jobs would queue forever. Pick one (PR body's also still describing the earlier dev-project iteration, worth a refresh while you're in there).

Fork→ubuntu-latest routing looks right. 👍

niks3089:
@G1de0n both addressed, thanks — the OIDC regression was a bad conflict resolution in yesterday's rebase (took the pre-#1571 side of the test job). Both legs now carry main's merged pattern verbatim: event-based `CI_HAS_DOPPLER_TOKEN`, `Validate Doppler OIDC configuration` against `vars.DOPPLER_CI_IDENTITY_ID`, the OIDC login step, and `permissions: id-token: write` on each job. No `DOPPLER_TOKEN_CI` references remain anywhere in the workflow.

On the label: `sdp-dev-vm` is the intended target, not a mismatch with sdp-infra#115 — that PR's `sdp-ci` pool is the superseded iteration. The decision landed on a single VM in the dev project (`gha-runner-sdp-dev-1`, org runner `gha-sdp-dev-1`, label `sdp-dev-vm`, online now); the #115 spot VMs never registered (their registration secret was never populated) and were stopped yesterday. Commenting on #115 to close the loop. The PR body here already describes the dev-VM iteration.

One heads-up for merge sequencing: jobs won't reach the runner until an org admin allows public repositories on its runner group (and restricts the group to this repo) — that's in the body's activation checklist.

G1de0n (changes requested):
The previous review's points are addressed: Doppler OIDC validation fails loud when the identity var is missing, fork PRs stay on GitHub-hosted runners, and the api/rest workspace split in run-workspace-tests.mjs filters correctly. Two things left:

1. The jobs target ["self-hosted", "sdp-dev-vm"], but the runner pool in sdp-infra#136 registers runners as ["self-hosted", "sdp-ci"]. One of the two has to change or every job queues forever, see my comment there; I'd move this PR to sdp-ci.
2. Nit: actions/checkout@3d3c42e5… is annotated # v7.0.1 here but the same sha is annotated # v6 on main, the trailing comment should match what the sha actually is.

niks3089:
@G1de0n both done:

1. **Standardized on `sdp-ci`** — both legs now target `["self-hosted", "sdp-ci"]`, matching the sdp-infra#136 pool. Merge order noted in the body: #136 applies and the pool comes online first, then this lands (repo-level JIT runners also mean no org runner-group changes are needed at all). The interim dev VM gets retired at cutover.
2. **The annotation nit cuts the other way**: `git ls-remote --tags actions/checkout` resolves `3d3c42e5…` to `refs/tags/v7` / `refs/tags/v7.0.1` — so `# v7.0.1` here is correct and it's main's `# v6` (in `sdp-api-ephemeral.yml`) that lies about the sha. Fixed main's annotations in #1674 rather than making this PR match the wrong comment.

G1de0n (approved):
Verified against the diff: both legs now carry main's merged OIDC pattern verbatim (event-based `CI_HAS_DOPPLER_TOKEN`, the `DOPPLER_CI_IDENTITY_ID` validation step, OIDC login, `id-token: write` per job), forks stay on `ubuntu-latest` with the fork-safe leg, both target `["self-hosted", "sdp-ci"]`, and `TEST_WORKSPACE_SPLIT` survives Doppler via `DOPPLER_PRESERVE_ENV`.

PR #1577, opened 2026-09-01
feat(payments): Lightspark per-corridor payouts and Grid settlement webhooks

multipletwigs:
- Off-ramp collection is destination-first: the wizard collects country → rail → that rail's exact fields from `resolveOfframpDestination`; the cross-rail union path is deleted.
- Payout external accounts are per `(counterparty, provider, fiat, destinationCountry)` corridor rows in `counterparty_provider_accounts` (migration 0075); `provider_data.payoutAccounts` is gone.
- Grid `OUTGOING_PAYMENT.*` webhooks settle ramp transfers: match by the `xfr_` id sent to Grid as the transaction `description`, promote `provider_reference` to the Grid transaction id, persist the on-chain hash and settlement economics idempotently.
- 7 newly covered payout countries (BD, CO, EG, GH, HT, PK, ZM) from the regenerated ramp-support snapshot.
- Ramp cancellation is by `transferId`; progressive collect responses no longer render as error toasts; deposit/pay wizards show transfer status in the header status bar and the complete screen links the settled on-chain transaction.

**before** — collect_account and settlement:

1. Collect step returned the union of every rail's fields with requiredness relaxed; exact validation only failed later at account build.
2. Payout accounts lived in `provider_data.payoutAccounts` keyed by a content hash; non-ACTIVE accounts were re-polled inline, quotes read the latest hash entry.
3. Lightspark transfers had no settlement loop: the in-app send 400d (`Transfer does not match the off-ramp deposit instruction`) because no `cryptoDeposit` was ever persisted, and completion never reached the row.

**after** — destination-first collection, corridor accounts, webhook settlement:

1. Requirements return a `PayoutRequirementTree` (`countryRails`/`railFields`/`accounts`); the wizard cascades country → rail → exact fields, reusing an ACTIVE corridor account when one exists for the chosen country.
2. Account creation reserves the corridor row first and uses the row id as Grid `platformAccountId`, so retries are idempotent; advance refreshes a non-ACTIVE Grid status before rejecting.
3. Quote persistence stores the quote's crypto-deposit instruction, the server rejects expired quote bindings on send, and Grid webhooks drive `awaiting_payment → settling → completed/failed/expired` with the on-chain hash on the row.

## How webhook matching works

1. The transfer row is created with `provider_reference` = the Grid quote id, which is never rewritten — the single reconciliation key both sides committed to before money moved.
2. Webhook event → exact lookup by `(provider, quoteId)`; unmatched events are logged and acked. No description or transaction-id matching exists — one identifier means partial correlation is unrepresentable.
3. Guarded status transitions make redelivery and out-of-order events no-ops (a stale `EXPIRED` cannot regress a transfer already `settling`). The reserved `xfr_` id is still sent to Grid as `description`, for human traceability in Grid's dashboard only.

**Verification**

- `tsc --noEmit` — clean across sdp-types, sdp-payments, sdp-api, sdp-web.
- `vitest` (DB-backed included) — webhook 4/4 with the real Grid payload verbatim, corridor repository 4/4, lightspark unit 35/35, OpenAPI spec 17/17, ramps+counterparties routes 53/53, web hooks 11/11.
- Live click-through on devnet: USD→MY SWIFT corridor account created via Grid sandbox, quote issued, in-app send submitted and settled on-chain.
- Two Codex review passes; both P1 findings (stuck non-ACTIVE accounts, expired-deposit send) plus the stale-EXPIRED regression and cancel-after-promotion 404 fixed in-range.

Known gaps

- GTQ/JMD have payout account specs but no destination country in Lightspark's docs table, so the resolver cannot offer them anywhere yet.
- Our country table still covers BW/CA/CG/HK, which Lightspark's current docs table no longer lists — needs confirmation with the provider.
- `provider_reference` promotion would drop a settlement if Grid ever emitted one transaction id for two transfers (unique-index conflict after webhook ack); provider-bug territory, logged loudly.
- Grid quotes are created before the transfer row persists; a persist failure orphans the provider quote (pre-existing ordering).


GuiBibeau (dismissed):

tobySolutions (dismissed):

multipletwigs:
@greptileai On the settlement-correlation findings — stating the trust model explicitly, since "don't trust the event" has to bottom out somewhere:

**Signature ≠ correctness.** Events are ECDSA-P256-verified against Grid's webhook public key before parsing, but a signed event can still be garbage (provider bug, sandbox noise, operator activity in Grid's dashboard). So we don't treat "signed" as "true" — we treat it as "attributable to Grid", and then bound what any single event is allowed to do:

1. **Shape**: strict-enough zod parsing; anything that doesn't carry the consumed fields is rejected or acked-and-ignored.
2. **Correlation (5f14e2fe5)**: centralized in `correlateRampSettlementTransfer`, fail-closed. Every supplied identifier resolves independently; an identifier that resolves to nothing refuses the event (never ignored); identifiers naming different transfers refuse the event; the survivor's `provider_reference` must equal the event's quote or transaction reference. No caller can bypass this by pre-supplying `transferId` — the service re-derives the match.
3. **Effect bounds**: even a fully correlated garbage event can only move its own row along the guarded transition table (compare-and-swap on allowed source statuses), can never regress a terminal transfer, can't touch tenancy it doesn't own, and redelivery is a no-op.

**The irreducible trust.** Pre-mainnet the webhook is deliberately the sole settlement authority — there's no on-chain reconciliation yet. At a transfer's first event, the Grid-signed `quoteId` is the *only* binding between our row and Grid's transaction, because the transaction id cannot exist on our side before its own first event promotes it. Refusing that binding means never settling anything: reconciliation requires trusting the provider's signed quote→transaction assertion exactly once per transfer, and after promotion, exact transaction-id equality takes over.

So the residual "partial correlation" is precisely that bootstrap step, and its worst case is bounded by (3) to a wrong-but-allowed status transition on the one transfer that provably owns the quote. If there's a concrete event sequence + row state that still mis-settles under this model, please spell it out; otherwise please re-review the correlation path at 5f14e2fe5.

GuiBibeau (dismissed):

tobySolutions (dismissed):

multipletwigs:
@greptileai The correlation design is superseded as of b902478da: reconciliation is now by the Grid quote id alone. `provider_reference` holds the quote id for the row's whole life and is never rewritten (no promotion, so the unique-index rewrite and post-promotion quote-uncheckability findings no longer have code to apply to); the webhook does one exact `(provider, quoteId)` lookup, unmatched events are acked and logged, and the transaction id and description are no longer consulted at all — with a single identifier, partial correlation is unrepresentable rather than checked for. Guarded transitions still bound what any event can do. Please re-review the settlement path at b902478da.

multipletwigs:
@greptileai One more scoping note for the re-review: legacy data is out of scope. SDP is pre-mainnet and devnet-only — there are no production transfer rows, so no compatibility path is needed for rows created under the earlier matching designs (description matching, provider_reference promotion, provider_data.quoteReference). The quote-only reconciliation at b902478da is the only semantics that has to hold, and any pre-existing devnet rows that don't conform can simply be left unsettled or reseeded. Please don't raise findings about migrating or falling back for legacy transfer rows.

keeganthomp (approved):

PR #1580, opened 2026-09-01
feat(web): streamline counterparty creation and section ramp requirement forms

multipletwigs:
**Counterparty create flow**
- Collapse the add-counterparty dialog to a single create step — the review step is gone; Create fires from Basic info, then flows into the optional crypto-accounts phase
- Entity type is a compact Individual/Business toggle inline with the display-name input
- Balanced dialog padding, no footer border, crypto-accounts phase grows naturally instead of a fixed 70vh scroll region

**Compliance screening**
- "Hold to add anyway" is now a regular button that opens a confirmation modal (same pattern as the delete-counterparty dialog)

**Counterparty selector**
- Popover instead of a modal, with each row subtly showing the entity type

**Ramp requirements UI**
- Lightspark rail bank fields are keyed `bankAccount.<field>` on the wire and mapped back to bare Grid payload keys at payout time
- Collect forms are sectioned into cards by the key's top-level dotted segment (`bankAccount`, `customer`, plus the synthesized destination selects); section copy is a provider-discriminated map in the web layer so it stays translatable
- Address cards gained a description; purpose of payment moved to the bottom of the individual PII form

**Provider branding**
- The wizard summary button is now "Checkout powered by [logo] Provider" and stays hidden until a provider is selected

**Notes**
- Any in-flight client submitting bare rail keys after deploy gets a loud 400 from the unchanged zod validation — no silent mis-mapping
**Screenshots**

Counterparty selector with entity types:
counterparty-selector

Single-step create dialog with the inline entity toggle:
create-dialog

Natural-height crypto-account phase:
crypto-account-phase

"Checkout powered by Lightspark" on the provider step:
checkout-powered-by

Grouped requirements cards (Recipient details + Residential address):
requirements-cards


multipletwigs:
Good catch on BVNK — its collect fields already carry dotted keys (`taxIdentification.*`, `cdd.*`, `address.*`) that weren't in the section-copy map, so the requirements screen threw.

Fixed in 62878dd: grouping is now opt-in per provider — providers without a copy-map entry render their dotted fields ungrouped (exactly the pre-PR layout), while an opted-in provider still fails loudly on a missing slug. BVNK's grouped cards + section copy land in the next PR.

tobySolutions (approved):

PR #1582, opened 2026-09-01
feat(custody): expose connection inventory runtime state

pashpashkin:
## What changed

- Replaced the internal Connections response field `providerCredential: { id, label, status }` with a connection-level contract containing `label`, `isDefault`, `isRuntimeExecutionAllowed`, and `defaultCustodyWalletId`.
- Kept the existing Connection lifecycle status unchanged.
- Extracted and reused the existing connection runtime-availability predicate without changing signing semantics.
- Normalized unknown stored failure codes to `null`.
- Added Zod validation in Web while accepting additive response fields.
- Updated the Connections page to consume the new label field and started the independent count and page queries concurrently.

## Why

HOO-1024 needs Connection selection and runtime state for the Dashboard without exposing provider credential internals.

Sharing the existing runtime-availability rules keeps the inventory consistent with runtime target resolution.

## Impact

- This is a breaking internal contract change between `sdp-api` and `sdp-web`; mixed old/new versions can break the Connections page.
- The current table continues to display the same label and lifecycle status.
- No default/runtime UI or management actions are added yet.
- No public API, OpenAPI, database, persistence, or dependency changes.
- No intended signing or value-moving behavior change.
- This PR does not enable the API or Web BYOK feature flags.

### Pre-release

- Deploy API and Web together, or keep the Web BYOK flag disabled until both revisions are live.

### Post-release

- Smoke-test the Connections page with selected, unselected, available, and unavailable Connections.
- If rollback is required, roll back API and Web together.
- No database rollback, migration, backfill, dependency installation, or configuration migration is required.

## Result Validation

- `pnpm typecheck` — passed.
- `pnpm check:module-boundaries` — passed.
- Focused API route tests — 7/7 passed.
- Full Web test suite — 185 files / 1,533 tests passed.
- Focused changed Web tests — 17/17 passed.

### Before

1. The internal API returned Connection lifecycle data with nested provider credential ID, label, and status.
2. Web rendered the nested credential label.
3. Selection and runtime availability were not exposed.

### After

1. The API returns the credential-derived label without exposing credential ID or status.
2. `isDefault` is `true` when BYOK runtime is enabled and the Connection is selected for the organization/project.
3. `isRuntimeExecutionAllowed` is calculated independently from local Connection, Credential, fingerprint, health-check, and default-wallet state.
4. Web validates required response fields and ignores additive fields.
5. Malformed required data produces the existing Connections load-error state.

When the API BYOK flag is disabled, rows still return with `isDefault: false` and `isRuntimeExecutionAllowed: false`. When the Web BYOK flag is disabled, the Connections page continues to redirect to Wallets.

## Known gaps

- The Dashboard does not yet display the new default/runtime fields or provide create, switch, or deactivate actions.
- There is no compatibility bridge between the old and new internal response formats.
- This remains an internal admin-only API; no public Connections API or non-admin restricted-state UX is added.

## Notes

- The endpoint remains dashboard-session-only with `custody:admin`; API keys are rejected.
- Both inventory queries remain scoped to the authenticated organization and project.
- `isDefault` and `isRuntimeExecutionAllowed` are independent: a selected Connection can be unavailable.
- Runtime availability represents local admission state, not live provider health or a guarantee that signing will succeed.
- Concurrent count and page queries preserve the existing non-atomic pagination model.

### DIFF
--- a/apps/sdp-api/src/routes/internal-custody/connections.test.ts
+++ b/apps/sdp-api/src/routes/internal-custody/connections.test.ts
@@ -22,6 +22,7 @@ const USER = {
   clerkId: "clerk_connections_read",
 };
 const SECRET_PAYLOAD = "encrypted-connections-read-secret";
+const ORIGINAL_PRIVY_BYOK_ENABLED = env.PRIVY_BYOK_ENABLED;
 
 function encodeJwtPart(value: Record<string, unknown>): string {
   return Buffer.from(JSON.stringify(value)).toString("base64url");
@@ -104,6 +105,7 @@ async function seedScope(): Promise<void> {
 async function seedCredentialAndConnection(input: {
   credentialId: string;
   connectionId: string;
+  label?: string;
   status: string;
   createdAt: string;
   failureCode?: string;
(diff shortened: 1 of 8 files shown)


pashpashkin:
@greptileai review pls

pashpashkin:
@greptileai review pls

tobySolutions (approved):

PR #1584, opened 2026-09-01
fix(docs): serve partner logos and raw img assets under /docs

keeganthomp:
## What

Prefix raw `` srcs in sdp-docs with `/docs` (24 partner logos on the home page, 2 guide screenshots) and add a rewrite mapping `/docs/images/:path*` onto `public/images`. CLAUDE.md gotcha added; `llms-full.txt` regenerated.

## Why

Users browse docs through sdp-web's `/docs/:path*` proxy, which never forwards root-level `/images` requests, so every raw img 404'd (works on docs.platform.solana.com directly, broken on platform.solana.com/docs). Markdown-syntax images were already fine: fumadocs compiles them to static imports.

## Reviewer notes

- **Fix lives in sdp-docs, not the proxy.** Adding an `/images/:path*` rewrite to sdp-web would claim the web app's root `/images` namespace and break silently if sdp-web ever adds its own. Everything under `/docs` is already guaranteed to proxy.
- **Rewrite ordering is safe.** afterFiles rewrites run before the `[[...slug]]` catch-all, same mechanism as the existing `.md` rewrite.

## Verification

| Check | Result |
| -- | -- |
| `pnpm build` | passes |
| `/docs/images/partners/privy.svg` on built server | 200, `image/svg+xml` |
| Home page srcs prefixed / unprefixed | 24 / 0 |
| Existing `.md` rewrite | still 200 |


keeganthomp:
Screenshot 2026-09-01 at 6 18 02 PM

multipletwigs (approved):
lfg

PR #1585, opened 2026-09-02
feat(ramps): country requirement-field kind — codes-only wire format

multipletwigs:
Implements PRO-1805 (part of the PRO-1804 reuse-JIT-accounts epic).

**Country field kind**

- Adds `{ kind: "country", key, label, required }` to the `RequirementField` union — the server stops inlining the 249-entry country option list (~27KB of duplicate JSON per collect response per provider).
- `countryField()` constructor + `fieldToZod` branch validating against `COUNTRY_CODES` (module-level singleton schema); values stay ISO 3166-1 alpha-2 end-to-end.
- Replaces all six inline country selects: Lightspark `customer.nationality` / `customer.region` / `customer.address.countryCode`, BVNK `taxIdentification.taxResidenceCountryCode` / `nationality` / `birthCountryCode`.
- Dashboard renders the country combobox from its own `COUNTRIES` copy with `regionFlagEmoji`-prefixed labels; client validation via `isCountryCode`.

**US_STATES removal**

- `address.stateCode` (BVNK, US-only) becomes a pattern-validated text field — address requirements are autofilled via place search, so the hardcoded state list is dead weight. Server-side `normalizeBvnkStateCode` still gates what reaches BVNK.
- Drops `US_STATES` / `UsState` / `UsStateCode` from `@sdp/types` and `usStates` from the field-options endpoint (zero dashboard consumers), OpenAPI schema, and regenerated playground catalog.

**before** — collect response inlines options:

1. `POST /v1/counterparties/:id/requirements` returns each country field as `kind: "select"` carrying all 249 options
2. three such fields per Lightspark collect, three per BVNK — same list serialized six times
3. client renders whatever options arrive

**after** — codes-only wire:

1. same endpoint returns `kind: "country"` with no options
2. client renders its own flag-prefixed dropdown from `COUNTRIES` and validates `isCountryCode`
3. server still rejects non-ISO values at the advance boundary via `z.enum(COUNTRY_CODES)`

## API before/after (live-verified against the local API)

`POST /v1/counterparties/:id/requirements` (Lightspark, collect_counterparty), per country field:

```jsonc
// before
{ "kind": "select", "key": "customer.nationality", "label": "Nationality", "required": true,
  "options": [ { "value": "AD", "label": "Andorra" }, /* …248 more */ ] }
// after
{ "kind": "country", "key": "customer.nationality", "label": "Nationality", "required": true }
```

`GET /v1/counterparties/field-options`:

```jsonc
// before: { "fields": { "entityTypes": [...], "countries": [...], "usStates": [ /* 55 entries */ ] } }
// after:  { "fields": { "entityTypes": [...], "countries": [...] } }
```

BVNK US onramp `address.stateCode`:

```jsonc
// before: { "kind": "select", "label": "State", "options": [ /* 55 US states */ ], ... }
// after:  { "kind": "text", "label": "State", "pattern": "^([A-Za-z]{2}-)?[A-Za-z0-9]{2}$", "placeholder": "CA", ... }
```

**Verification**

- `pnpm turbo typecheck` — 20/20 passed
- `apps/sdp-api` `vitest run src/routes/counterparties.test.ts` — 22/22 (full-shape field assertions updated)
- `apps/sdp-api` BVNK counterparty tests — 9/9; `packages/sdp-payments` — 44/44
- `biome check` on changed files — clean

## Screenshots

Lightspark requirements step — country combobox with flag-prefixed labels

**Verified in the running app (post-merge-to-main rebase)**

- local app: deposit wizard → Lightspark → requirements step — nationality/region/address-country render as flag-prefixed comboboxes, search filters, selection sticks (🇲🇾 Malaysia); zero console errors
- live API (local key): `POST /:id/requirements` collect response is now 1,879 bytes with `kind: "country"` fields carrying no options; `GET /counterparties/metadata` no longer returns `usStates`
- CI: the Unit Tests failure was `@sdp/rpc` `transient.test.ts` ("stops retrying once the elapsed budget is spent") — a timing flake in an untouched package, passes 3×3 locally; rerun triggered


resourcefulmind (approved):
Approving. Client and server validate from the same countries module so the lists can't drift, and I checked all 249 codes have flag sequences, so the combobox's module-level throw can't fire today.

PR #1590, opened 2026-09-02
feat(web): offramp wizard — reuse an existing corridor account or add another

multipletwigs:
Implements PRO-1810 (PRO-1804 epic). Stacked on #1589.

**Account chooser on the payout-details step**

- Picking a `destinationCountry` with existing corridor accounts now renders them as selectable rows (flag, bank name, `••••` last-4, rail badge) plus an explicit **Add new account** row — replacing the silent skip.
- Selection is a discriminated tri-state (`none` / `existing(id)` / `new`) in `useCounterpartyRequirements`; the single-account case resolves to that account **during render** (no effects), so today's zero-click flow is unchanged.
- Selection resets with the same semantics as collected data: subject change (counterparty/provider/currency) and `destinationCountry` change.
- "Add new account" falls through to the rail select + bank-field collection exactly like the zero-account path; multiple accounts per rail stay legal.
- Step gating: `isComplete` now also requires a resolved account choice — with several accounts and nothing picked, Next is disabled instead of deferring the failure to the quote's ambiguity 400.
- The chosen id rides dedicated wizard state (not `collectedData`) through the memo step into the quote payload as `providerAccountId` (#1589); rows without enrichment render plain (id + rail), never placeholder-defaulted.

**before**: country picked + any account → collection silently skipped, no choice, quote resolves implicitly.
**after**: country picked + accounts → chooser (single account pre-selected); add-new collects rail + fields; the quote pays out to whichever was chosen.

## Screenshots (USD→MY corridor, two active accounts, live app)

Chooser — nothing selected yet, Next disabled:

chooser

Existing account selected (enriched: Maybank •••• 5678, SWIFT), Next enabled:

selected

"Add new account" falls through to rail + field collection (screenshot predates 96ce15515, which moves these fields inline into the chooser card):

add new

**Verification**

- typecheck + biome clean; hook unit tests 8/8 (pre-selection, add-new fall-through, reset on country change); CI gates the full suites
- local app walk (screenshots above): chooser, selection gating, add-new rail fields, unenriched-row rendering — zero console errors
- second corridor account for the walk was seeded directly in the dev DB and removed afterward

**Greptile P1 fixes (folded in, last commit)**

- Advance takes explicit account intent instead of guessing: `providerAccountId` in the submit body reuses exactly that account (validated via shared `requireLightsparkPayoutAccountById`, same predicate as the quote); `paymentRails` present always creates/resumes a **new** account — a completed corridor account is never silently adopted, so "Add new account" details can no longer be discarded in favor of the old account.
- Choosing among several existing accounts now advances: the wizard sends the selected `providerAccountId`, so the gate no longer loops the collect tree.
- Ready's resolved account id flows synchronously through `onReady(providerAccountId)` into quote creation — no stale React state read, no ambiguity 400 after an add.
- Verified: typecheck + biome clean; counterparties/ramps/lightspark API suites 64/64, hook unit tests 8/8.


multipletwigs:
Fixed the P1 (new-account identity lost) in 96ce15515, plus a UX change:

**P1 fix — the advance now returns which account it resolved**

- `CounterpartyRequirements` ready variant gains optional `providerAccountId`; `advanceLightsparkRequirements` sets it on all three offramp ready paths (stored-active fast path, Grid-refreshed, and newly created via `ensureLightsparkPayoutAccount`) — so both the add-new path AND a rail-matched reuse recover the account identity.
- The wizard hook adopts it on a ready result (`setPayoutAccountSelection({ kind: "existing", id })`), so the deferred quote after the memo step always carries `providerAccountId` and the multi-account corridor can no longer 400 on its own creation.
- Contract test pins the shape: offramp advance for a provisioned corridor returns `{ status: "ready", providerAccountId }` with zero provider calls on the fast path. OpenAPI snapshot updated.

**UX — add-new collects inline**

- Choosing "Add new account" now renders the rail + bank fields inside the chooser card directly under the button, instead of splitting them into the Destination card above.

resourcefulmind (dismissed):

G1de0n (dismissed):

multipletwigs:
Update: plans changed — the three Greptile P1 fixes are now **folded into this PR** (commit 1f3b66799) rather than deferred. Advance takes explicit account intent (`providerAccountId` to reuse, `paymentRails` to create — never silently adopting an existing corridor account), and ready's account id threads synchronously into quote creation. PRO-1826 closed accordingly.

multipletwigs:
@greptileai please re-review — the P1 fixes landed in 1f3b66799.

pashpashkin (dismissed):

pashpashkin (dismissed):

multipletwigs:
@greptileai re-review — addressed both blockers: the stale-selection race is closed in the wizard (form/Next/Back locked during the advance; requirements subject now covers the full corridor so a pending onboarding never survives a corridor change), and lightspark bank-field validation now runs before any pending account row is persisted, with advance-route regression tests for both the no-pending-row and cross-counterparty providerAccountId cases.

multipletwigs:
@greptileai re-review — pending-row resumption removed entirely (fresh platform identity per submission; bank-detail sameness can't be proven under the store-nothing PII policy, so identity reuse is never safe), and the status poll now discards results that land after the corridor subject changed.

multipletwigs:
@greptileai re-review — the applied-response guard now covers the collected destinationCountry alongside the request subject, and the submit-ready path checks the same captured corridor identity before committing results (a stale advance resolves null and the wizard does not act on it).

multipletwigs:
@greptileai re-review — a destinationCountry change now drops the pending onboarding lifecycle and its poll payload, so no later poll tick can start under the new country with the abandoned corridor's advance payload. Only the Lightspark payout form collects destinationCountry, so provider KYC verification polls are unaffected.

multipletwigs:
@greptileai re-review — the cryptoToken→assetRail contract migration is backed out of this PR and ships separately from main as #1607; this PR now carries only the corridor-account selection feature and its hardening.

multipletwigs:
@greptileai re-review — the country-change handler now retires the corridor identity ref synchronously, closing the pre-render window where an already-resolving advance or poll could pass the staleness guard.

multipletwigs:
@greptileai re-review — fresh-identity retries now archive the corridor's earlier incomplete reservations before inserting the new one, so orphaned local state cannot accumulate. The unlinked remote account from a provider-success/local-failure attempt is unreachable by design: corridor resolution and the account chooser only read completed local rows, so it can neither receive a payout nor make unqualified resolution ambiguous.

multipletwigs:
@greptileai re-review — the staleness guards are gone entirely, replaced by the structural fix: advance responses are stored tagged with the corridor they answered for and every read filters on the current corridor, the status poll is a pure SWR fetch keyed by that corridor (no fetcher side effects; a corridor change kills the key), the ready-resolved account and progressive collection are derived at render, and quote firing is a single readiness effect behind the existing once-per-wizard guard. There is no application-time window because responses are never applied.

multipletwigs:
@greptileai re-review — the poll cache is now keyed per advance (an advance sequence in the key), so an earlier advance's verdict can never answer for a later same-corridor advance; and payout reservations are serialized: archive+insert run in one transaction under a new partial unique index (at most one live reservation per corridor and rail — concurrent submissions get a 409, never a second active identity).

multipletwigs:
@greptileai re-review — the poll identity is now a per-advance random UUID rather than a counter, so a subject round trip cannot recreate an earlier cache key; collision with a prior advance's cached verdict is impossible by construction. Note the scenario is unreachable with current providers (only BVNK/Mural ever poll, and their ready responses carry no providerAccountId; Lightspark, the only provider with corridor-bound accounts, answers synchronously and never polls) — this hardens the provider-generic hook.

multipletwigs:
@greptileai re-review — the request path no longer archives anything: a live reservation is protected by the partial unique index and a concurrent submission gets a 409 while the first is in flight (its parent row can no longer be invalidated mid-completion). A failed request archives only its OWN reservation before rethrowing, so retries are clean; crash-orphaned reservations are a cron follow-up. Repository test updated to pin the one-live-reservation-per-corridor invariant.

GuiBibeau (approved):

PR #1593, opened 2026-09-02
feat(earn): add WisdomTree integration

cutbow7:
Ticket: PRO-1703 (WisdomTree Connect provider integration, earn)

## What

Full WisdomTree provider integration across the Earn stack:

- **`@sdp/wisdomtree`** — the provider package: fund-mint verification against the measured registry (owner program, decimals, transfer hook, live pause state), the hand-rolled SPL TLV transfer-hook account resolver mirroring `addExtraAccountMetasForExecute`, deposit/redemption instruction builders (USDC in, Token-2022 shares out), chain reader, and an opt-in mainnet-fork Surfpool smoke test.
- **`@sdp/earn`** — the Connect REST client (OAuth2 token cache keyed by a SHA-256 digest of the credential tuple, products/wallets/organization reads, fail-closed deposit eligibility), catalogue sync, and the provider-order withdrawal capability (`EarnVaultProviderOrderWithdrawProvider`).
- **`@sdp/api`** — execution-registry wiring, provider-aware movement settlement filters, and reconciliation that caps provider-order rows at `confirmed` and never stamps `settled_at` or closes positions from the chain leg alone.
- **`sdp-web`** — provider-order deposit/withdrawal modals (share-denominated inputs, no fake floors, "awaiting provider settlement" presentation), settlement-aware watch predicates, and Treasury surfacing (still disabled via `EARN_PROVIDER_SURFACING.wisdomtree = false` until go-live).
- Docs, OpenAPI text, module map, i18n (en-only per translation-bot convention), and boundary scripts updated.

## Reviewer notes

Both review rounds addressed on this branch:

- **React Doctor (merge-blocking)**: `WithdrawalReviewStep` exceeded the complexity budget; the settlement-conditional `<dl>` rows are extracted into `ReviewAmountRows` → `ProviderOrderReviewRows` / `AtomicReviewRows`. React Doctor is now clean (91/100).
- **Permanent pollers (blocking)**: `isEarnVaultDepositInFlight` lost its provider-order override — the watch stops at the legacy wire-terminal set, so a confirmed provider-order deposit is settled. `isEarnVaultWithdrawalInFlight` now treats provider-order `confirmed` as watch-terminal (`PROVIDER_ORDER_WATCH_TERMINAL_STATUSES`, the same fork risk the legacy deposit poll accepts), so reconciler-parked rows stop polling and `onSettled` fires; the "awaiting provider settlement" copy stays in presentation. Pinned unit tests flipped accordingly; the CLAUDE.md poll docs updated.
- **Error vocabulary (blocking)**: the package taxonomy now speaks the shared API vocabulary — `CHAIN_UNREADABLE` → `VAULT_UNREADABLE` (retryable 503), paused-mint refusals raise `DEPOSIT_REFUSED`/`WITHDRAW_REFUSED` by build direction, hook (KYC) refusals raise `WITHDRAW_REFUSED`, and `MINT_MISMATCH` remains for genuine registry drift. The dead `FUND_UNKNOWN` code was dropped.
- **Advisory round 1**: eligibility wallets/products reads now run in `Promise.all` (one fewer Connect round-trip on the money-in path); `readVaultPositions` throws `CLUSTER_UNSUPPORTED` instead of claiming `tokenMint: ""`.
- **Advisory round 2**: product/wallet readers treat JSON `null` as absence, not a type error; the deposit provider-order copy is provider-neutral (the fail-closed settlement answer for an unknown provider no longer names WisdomTree); the Surfpool failure log redacts the remote RPC URL; `transfer-hook` gains offline tests for the external-program discriminator branch (>= 128) and an offline pin of the measured 11-account WTGXX layout; `client.test.ts` asserts the Purchase/Sale on-receipt threading and destination ATAs, plus a `buildVaultWithdrawal` success path; `execution-registry` pins the client-capability ↔ settlement-table agreement for every provider; the withdrawals list test seeds a finalized provider-order row on both sides of `?settled=`; the reconciler test re-sweeps a parked confirmed row asserting idempotence (stable `confirmed_at`, no `settled_at`, still claimable).
- **OAuth (non-code)**: Connect documents only the ROPC password grant; whether WisdomTree shipped a replacement is unmeasured without a tenant. Recorded in the package CLAUDE.md UNVERIFIED list: accepted-over-TLS for now, re-check at credential time.
- Treasury position badges being settlement-aware is tracked as follow-up work before the surfacing flip (not in this diff).

## Verification

- [x] React Doctor: `--scope changed --project sdp-web,sdp-docs` — 91/100, no issues
- [x] sdp-web: 299 files, 2,883 tests passed
- [x] @sdp/earn: 179 tests passed
- [x] @sdp/wisdomtree: 37 tests passed, 1 expected skip (opt-in Surfpool e2e)
- [x] @sdp/api targeted: execution-registry, vault-withdrawals, movement reconciliation — 97 tests passed (full API suite green on the base commit)
- [x] Workspace typecheck: 25/25 tasks
- [x] Biome on all changed files: clean
- [ ] `pnpm wisdomtree:surfpool:e2e` — Docker-only, requires real mainnet RPC, funded mainnet signer, and production Connect credentials; the smoke test correctly skips in unit runs


packages/sdp-wisdomtree/src/client.ts
```diff
@@ -0,0 +1,332 @@
...
+   * a Token-2022 TransferChecked of fund tokens to WisdomTree's on-receipt
+   * Sale wallet, hook accounts resolved live. Implementing this is what flips
+   * `supportsVaultWithdraw` to true (the same PRO-1702 mechanism Kamino used).
+   *
+   * Per ADR 0002 this path takes NO surfacing/entitlement/eligibility gate —
+   * a wallet holding the tokens proved its standing on-chain, and the hook
+   * re-proves it at execution. `rentRefundTo` is accepted and unused: no
+   * account is closed by a redemption (see the builder).
+   */
+  async buildVaultWithdrawal(
```
keeganthomp, line 218: [P2] Reject unsupported withdrawal floors at the provider boundary

The custody withdrawal service forwards `minAmountOut` here, but this builder ignores it and returns no `accepted.minAmountOut`. The shared validator then throws `INTERNAL_ERROR`, so a caller supplying this valid optional field gets a 500 instead of the explicit 400 used for unsupported `minSharesOut`. Reject `input.minAmountOut` before the Connect/RPC work, or implement and report it in the plan.

keeganthomp (changes requested):
The required semantic PR-title check is failing because `WisdomTree (mono)` has no release-type prefix. Rename it to a conventional title such as `feat(earn): add WisdomTree integration`. I also left one non-blocking withdrawal-contract issue inline.

Unanchored findings:

- [P1] Use a semantic PR title

  The required `semantic-pr-title` job fails because `WisdomTree (mono)` has no allowed release-type prefix. Rename the PR with an allowed conventional type so the required check can pass.

keeganthomp (dismissed):
woo!

We do need to make sure sdp-infra things are PR'd (if needed) + api keys are set accordingly for staging env (smokey)

cutbow7:
Converted to draft because WisdomTree requires **passing username and password in plaintext in their OAuth flow.** They are fixing this.

apps/sdp-web/src/app/dashboard/markets/earn/earn-program-data.ts
```diff
@@ -933,11 +935,17 @@ function useEarnVaultMovementOutcome<Movement extends WatchableVaultMovement>(in
...
  */
 export function isEarnVaultDepositInFlight(deposit: EarnVaultDepositRecord): boolean {
+  if (
+    deposit.status === "confirmed" &&
+    earnProviderDepositSettlement(deposit.provider) === "provider_order"
+  ) {
+    return true;
+  }
   return !SETTLED_VAULT_MOVEMENT_STATUSES.has(deposit.status);
 }
```
keeganthomp, line 961: This predicate is also the poll stop condition (`isEarnVaultDepositSettled = !inFlight` feeds `earnVaultMovementRefreshInterval`) and the Treasury recovery filter. A provider-order deposit never gets a wire state after `confirmed` (the API deliberately parks it there, see `vaultSettlementFilter`), so every WisdomTree deposit ever made is re-discovered on each Treasury visit via `settled=false`, mounts a tracker, and polls at 5s forever with `onSettled` never firing. Your own CLAUDE.md edit at L473-481 describes exactly this failure mode.

Keep the watch predicate on the wire terminal set and express "awaiting provider settlement" in presentation only (the modals already branch on settlement kind). The unit test at `earn-program-data.unit.test.ts` L106-121 pins the current behaviour and needs flipping too.
```suggestion
export function isEarnVaultDepositInFlight(deposit: EarnVaultDepositRecord): boolean {
  return !SETTLED_VAULT_MOVEMENT_STATUSES.has(deposit.status);
}
```

apps/sdp-web/src/app/dashboard/markets/earn/earn-program-data.ts
```diff
@@ -1155,11 +1163,17 @@ const SETTLED_VAULT_WITHDRAWAL_STATUSES: ReadonlySet<EarnVaultDirectMovementStat
...
 /** Shared by the recovery filter and the poll's stop condition — one rule. */
 export function isEarnVaultWithdrawalInFlight(withdrawal: EarnVaultWithdrawal): boolean {
+  if (
+    withdrawal.status === "finalized" &&
+    earnProviderWithdrawalSettlement(withdrawal.provider) === "provider_order"
+  ) {
+    return true;
+  }
   return !SETTLED_VAULT_WITHDRAWAL_STATUSES.has(withdrawal.status);
 }
```
keeganthomp, line 1201: Same class of problem, different cause: the reconciler parks a provider-order withdrawal at `confirmed`, which is already outside `SETTLED_VAULT_WITHDRAWAL_STATUSES`, so its tracker polls forever with or without this branch (the branch only adds legacy `finalized` rows to the forever set). Either treat provider-order `confirmed` as watch-terminal here (same fork risk the legacy deposit poll already accepts) and keep the pending copy in presentation, or add a wire fact for the chain leg before surfacing. As written, an org with N WisdomTree withdrawals spends N detail reads per 5s whenever Treasury is open.

packages/sdp-wisdomtree/src/errors.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export type SdpWisdomTreeErrorCode =
+  /** The providerReference names no fund in the measured registry. */
+  | "FUND_UNKNOWN"
+  /** The fund's instrument does not exist on the requested cluster. */
+  | "CLUSTER_UNSUPPORTED"
+  /** Caller-fault amount problems: unparsable, zero, or sub-atomic precision. */
+  | "INVALID_AMOUNT"
+  /** A chain read failed or returned an account this package refuses to trust. */
+  | "CHAIN_UNREADABLE"
```
keeganthomp: The API maps provider build failures by the shared code vocabulary in `apps/sdp-api/src/services/earn/vault-refusals.ts` (`INVALID_AMOUNT` / `DEPOSIT_REFUSED` / `WITHDRAW_REFUSED` / `COMPLIANCE_APPROVAL_REQUIRED` to 400, `VAULT_UNREADABLE` to 503, anything else rethrown as a 500), and Kamino, Veda and Jupiter Lend all speak it. Nothing in this union is mapped, so an RPC outage (`CHAIN_UNREADABLE`), a paused fund (`MINT_MISMATCH`) and the KYC hook refusing a redemption (`HOOK_UNRESOLVED`) all reach the caller as 500s, contrary to the docstring above.

Cheapest fix: rename `CHAIN_UNREADABLE` to `VAULT_UNREADABLE`, and raise the paused-mint and hook refusals as `DEPOSIT_REFUSED` / `WITHDRAW_REFUSED` (or `COMPLIANCE_APPROVAL_REQUIRED`), keeping the mismatch codes for genuine registry drift. Not blocking while unsurfaced, but please land it here rather than at flip time.

packages/sdp-earn/src/providers/wisdomtree/connect.ts
```diff
@@ -0,0 +1,461 @@
...
+  const record = value as Record<string, unknown>;
+  const fields = [
+    ["id", "number"],
+    ["name", "string"],
+    ["exchange_code", "string"],
+    ["issuer", "string"],
+    ["can_trade", "boolean"],
+  ] as const;
+  for (const [field, expectedType] of fields) {
+    if (record[field] !== undefined && typeof record[field] !== expectedType) {
```
keeganthomp: nit: `typeof null === "object"`, so a null optional field (`issuer: null`, `exchange_code: null`, routine in DRF-style APIs) throws `PROVIDER_UNAVAILABLE` and skips the whole catalogue pass because of one product. Treat null like undefined here and in `readWisdomTreeWalletRecord` (L341); the wire shape is unverified anyway, so fail closed on wrong types, not on absence.

apps/sdp-web/messages/en/dashboard-earn.json
```diff
@@ -802,6 +803,14 @@
       "vaultConfirmedBody": "The deposit is confirmed on Solana and your position is active.",
       "vaultConfirmedStatus": "Deposited",
       "vaultConfirmedNote": "The live position will reflect the provider's next successful chain read.",
+      "providerOrderConfirmedTitle": "Subscription sent",
+      "providerOrderConfirmedBody": "Solana confirmation has been observed for the USDC transfer. WisdomTree still needs to strike NAV and deliver the fund shares; this is not a completed position yet.",
+      "providerOrderConfirmedStatus": "Processing at WisdomTree",
```
keeganthomp: nit: this copy is selected by `earnProviderDepositSettlement(provider) === "provider_order"`, which is also the fail-closed answer for any provider id the web build does not know, so an API deploy that adds a provider ahead of the web ships would render "Processing at WisdomTree" for it. Interpolate `{provider}` via `earnProviderLabel` (already extended in `earn-format.ts`), or use neutral wording like the withdraw copy does.

scripts/kora-surfpool/e2e-wisdomtree.sh
```diff
@@ -0,0 +1,111 @@
...
+cleanup() {
+  local status=$?
+  trap - EXIT
+  if [ -n "${SURFPOOL_PID}" ] && kill -0 "${SURFPOOL_PID}" >/dev/null 2>&1; then
+    kill "${SURFPOOL_PID}" >/dev/null 2>&1 || true
+    wait "${SURFPOOL_PID}" >/dev/null 2>&1 || true
+  fi
+  if [ "${status}" -ne 0 ] && [ -f "${SURFPOOL_LOG}" ]; then
+    echo "Embedded Surfpool log:" >&2
+    tail -200 "${SURFPOOL_LOG}" >&2 || true
```
keeganthomp: nit: on failure this tails Surfpool's own log, which was handed the raw `SURFPOOL_REMOTE_RPC_URL`; for most RPC providers the API key is in that URL, and the 1.5.0 binary has "connecting to" / datasource-error format strings. That undoes the credential-by-name discipline the rest of the script keeps.
```suggestion
    tail -200 "${SURFPOOL_LOG}" | sed "s|${WISDOMTREE_SURFPOOL_MAINNET_RPC_URL}|<redacted rpc url>|g" >&2 || true
```

packages/sdp-wisdomtree/src/transfer-hook.ts
```diff
@@ -0,0 +1,279 @@
...
+    let resolvedAddress: Address;
+    if (entry.discriminator === 0) {
+      resolvedAddress = getAddressDecoder().decode(entry.addressConfig);
+    } else {
+      const seeds = await unpackSeeds(reader, entry.addressConfig, executeKeys, executeData);
+      let programAddress: Address;
+      if (entry.discriminator === 1) {
+        programAddress = input.hookProgram;
+      } else {
+        const programIndex = entry.discriminator - (1 << 7);
```
keeganthomp, line 249: nit: the offline suite only covers discriminators 0 and 1; this external-program branch (`>= 128`, program from an execute-key index) has no test, yet CLAUDE.md says the real WTGXX list is exactly two account-data-seeded PDAs on an external program. A regression in this index arithmetic would pass CI and only surface in the opt-in smoke run. Capturing the live ExtraAccountMetaList verbatim (like the mint image in `fixtures.test-helper.ts`) and asserting the resolved 11-account layout offline would also pin the hardcoded indexes the smoke test relies on.

packages/sdp-wisdomtree/src/client.test.ts
```diff
@@ -0,0 +1,244 @@
...
+
+    const plan = await client.buildVaultDeposit(productionCtx, {
+      providerReference: WTGXX.mint,
+      owner: OWNER,
+      amount: "100.25",
+    });
+
+    expect(plan.cluster).toBe("mainnet-beta");
+    expect(plan.assetIdentity).toEqual({ depositTokenMint: USDC, shareMint: WTGXX.mint });
+    expect(plan.accepted).toEqual({ amount: "100.25" });
```
keeganthomp, line 146: nit: the test is titled "against the API-resolved on-receipt wallet" but never asserts the transfer destination is `ATA(ON_RECEIPT, USDC)` or that the Connect call carried `trade_type=Purchase`, `fund=WTGXX`, `currency=USDC`, `blockchain=Solana`, so a client that ignored the Connect response would pass. There is also no `buildVaultWithdrawal` success test (L165-190 only covers the `minAmountOut` refusal), so the `Sale` threading and the shares `INVALID_AMOUNT` to 400 mapping are offline-untested.

apps/sdp-api/src/services/earn/execution-registry.test.ts
```diff
@@ -227,6 +227,26 @@ describe("resolveVaultDirectClient", () => {
...
+      supportsVaultWithdraw,
+    } = await import("@sdp/earn/capabilities");
+
+    const client = resolveVaultDirectClient(executionEnv, "wisdomtree", createVaultDeadline());
+
+    expect(client).not.toBeNull();
+    if (!client) throw new Error("expected WisdomTree vault-direct client");
+    expect(supportsVaultDirect(client)).toBe(true);
+    expect(supportsVaultWithdraw(client)).toBe(true);
+    expect(supportsVaultProviderOrderWithdraw(client)).toBe(true);
```
keeganthomp, line 245: nit: the reconciler decides withdrawals by this client capability while `?settled=` and the dashboard decide by `EARN_PROVIDER_WITHDRAWAL_SETTLEMENT`. Nothing asserts the two agree, so a future provider that sets one and not the other would have the API parking rows the list calls settled. One loop over `EARN_PROVIDERS` here comparing `supportsVaultProviderOrderWithdraw(client)` to the table would pin it.

apps/sdp-api/src/db/repositories/earn-movements.repository.ts
```diff
@@ -917,18 +921,44 @@ function allowedSourceStatuses(model: EarnExecutionModel, toStatus: string): rea
...
+  withdrawal: Object.entries(EARN_PROVIDER_WITHDRAWAL_SETTLEMENT)
+    .filter(([, settlement]) => settlement === "atomic")
+    .map(([provider]) => provider),
+} as const satisfies Record<EarnMovementDirection, readonly string[]>;
+
+const ATOMIC_SETTLED_STATUSES_BY_DIRECTION = {
+  // The legacy deposit DTO has no finality state and has always stopped at
+  // confirmed for atomic providers. Keep that compatibility boundary here.
+  deposit: ["confirmed", "finalized"],
+  withdrawal: ["finalized"],
```
keeganthomp, line 945: nit: the deposit list test now covers both sides of the provider-aware filter, but `earn.vault-withdrawals.test.ts` still seeds Kamino only, so this withdrawal set and the provider gate on the withdrawal clause have no test. A `wisdomtree` withdrawal advanced to `finalized` asserted present under `settled=false` and absent under `settled=true` would close that.

apps/sdp-api/src/services/earn/vault-movement-reconciliation.service.ts
```diff
@@ -336,6 +339,19 @@ async function reconcileMovement(
     return "failed";
   }
   if (status?.confirmationStatus === "finalized") {
+    if (usesProviderOrderSettlement(env, movement)) {
+      // The payment/share leg is irreversible, but Connect has not yet
+      // reported that the subscription/redemption order settled. The current
+      // ledger has no `awaiting_provider` state, so preserve the strongest
+      // honest non-terminal fact it can express. In particular, never stamp
+      // settled_at/token_amount_settled or close the position from this leg.
+      if (movement.status === "confirmed") return "unchanged";
```
keeganthomp, line 384: nit: every provider-order test runs one pass from `requested`; none re-runs the sweep on the row it just parked here, which is the state every WisdomTree row sits in on every tick until a Connect reconciler exists. Without this line the branch would re-issue a `confirmed -> confirmed` write the transition table refuses and still report `confirmed`. A second `reconcileEarnVaultMovements(env)` call asserting `unchanged`, `confirmed_at` stable and `settled_at` null would pin it.

keeganthomp (changes requested):
Solid integration and the package notes are unusually honest about what is measured versus assumed. The smoke harness claims (no broadcast, credentials by env name, cheatcode lockout ordering) all check out against the code, and both items from my earlier review (semantic title, `minAmountOut` refusal) are resolved.

Blocking: `apps/sdp-web/src/app/dashboard/markets/earn/earn-program-data.ts:937` makes provider-order `confirmed` in flight, which is also the poll stop condition, so every WisdomTree movement polls forever and remounts on every Treasury visit; `:1165` has the same outcome via the reconciler parking withdrawals at `confirmed`. Please also land the error-vocabulary fix at `packages/sdp-wisdomtree/src/errors.ts:14` here (RPC outages, paused fund and hook refusals currently surface as 500s).

Outside the diff: `TreasuryPositionStatusBadge` maps `confirmed` to "Active" regardless of settlement kind ([workspace](https://github.com/solana-foundation/solana-developer-platform/blob/c33ba9b0a0756de499f21e60e069f592614a6801/apps/sdp-web/src/app/dashboard/markets/treasury-solutions/treasury-solutions-workspace.tsx#L499-L515), [ui-state](https://github.com/solana-foundation/solana-developer-platform/blob/c33ba9b0a0756de499f21e60e069f592614a6801/apps/sdp-web/src/app/dashboard/markets/earn/earn-vault-ui-state.ts#L16-L29)), so a first WisdomTree deposit shows an Active position before any shares exist. Should be settlement-aware before surfacing flips.

Two non-code asks: the 09-03 draft note said WisdomTree was fixing the plaintext username/password OAuth flow, and `connect.ts` still does `grant_type=password`; say whether their fix landed or we are accepting ROPC over TLS (and record it in the package CLAUDE.md UNVERIFIED list). And the PR body is empty on a 96-file change; the repo shape is `## What` / `## Reviewer notes` / `## Verification`, plus a link to the Earn ticket.

keeganthomp (approved):

PR #1596, opened 2026-09-02
fix(ci): accept platform-manifest digests in the candidate identity check

niks3089:
The v0.70.0 release deploy (run 33588593013) failed its post-deploy candidate identity check and auto-rolled back — correctly executed rollback, wrong reason. Release images are multi-arch: verify-and-promote records the **OCI index** digest (`d419…`), but Cloud Run pins and reports the **platform manifest** it serves — the index's linux/amd64 child (`896e…`, confirmed against the registry). Comparing those two levels can never succeed, so every multi-arch release would roll back. v0.69.0 predated the signed-artifacts path; v0.70.0 was its first live run.

The check now fetches the verified index's child list from the registry (same authenticated AR endpoint the deploy uses) and accepts the served digest only if it is exactly the verified index digest or one of its platform-manifest children — still a closed allowlist derived from the cryptographically verified digest, so the security property is unchanged.

Contract test added pinning the allowlist construction; 124/124 pass.

After merge: re-run the v0.70.0 deploy (workflow_dispatch on deploy-sdp-api-gcp-prod with the release sha) to ship the release that was rolled back.

### DIFF
--- a/.github/workflows/deploy-sdp-api-gcp-prod.yml
+++ b/.github/workflows/deploy-sdp-api-gcp-prod.yml
@@ -381,8 +381,18 @@ jobs:
               --format='value(status.imageDigest)'
           )"
           expected_digest="${IMAGE##*@}"
-          if [[ "${revision_digest##*@}" != "${expected_digest}" ]]; then
-            echo "Candidate ${candidate_revision} resolved to ${revision_digest}, expected ${expected_digest}." >&2
+          accepted_digests="${expected_digest}"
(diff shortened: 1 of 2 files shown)


niks3089:
Superseded by #1594 — resourcefulmind landed the same fix first with a stronger shape (crane-based, media-type-scoped, hard-fails when the index can't be enumerated). Same root cause and allowlist semantics; nothing here adds to it.

PR #1646, opened 2026-09-04
refactor(payments): derive ramp terminal and cancelable statuses from shared records

multipletwigs:
Independent hand-kept lists of ramp status semantics — which statuses are terminal, and which are cancelable — now derive from exhaustive records in `@sdp/types`.

*(Previously stacked on #1574; rebased onto `main` and no longer depends on it. #1574's `isTerminalTransferStatus` helper move left this PR with that rebase — it rejoins #1574.)*

**Shared source**

- `RAMP_TRANSFER_STATUS_TERMINAL` and `RAMP_TRANSFER_STATUS_CANCELABLE` in `packages/sdp-types/src/payments.ts`: exhaustive `Record<PaymentTransferStatus, boolean>`, so adding a status forces both decisions at compile time.
- Lookups: `isTerminalRampTransferStatus`, `isCancelableRampTransferStatus`; `CANCELABLE_RAMP_TRANSFER_STATUSES` (derived array) for status-guarded updates. Scoped to ramps — onchain sends terminate at `finalized`.
- `PaymentTransferStatus` is now derived from a `PAYMENT_TRANSFER_STATUSES` const (same union, enables deriving the arrays without casts).

**Call sites now derive instead of re-list**

- `apps/sdp-api` ramp events handler: local `TERMINAL_RAMP_STATUSES` array deleted, identical set.
- `apps/sdp-api` ramp settlements: `isSettlementFinalStatus` = shared set minus `expired` — the documented carve-out stays (expiry is derived from provider absence, a provider event proving activity must revive it).
- `apps/sdp-api` cancel-transfer handler: inline `["pending", "awaiting_payment"]` replaced by the shared cancelable set for both the guard and `fromStatuses`.
- `apps/sdp-api` `routes/payments/schemas.ts`: `transferStatusSchema` now `z.enum(PAYMENT_TRANSFER_STATUSES)` instead of re-listing all ten statuses.
- `sdp-web` `ramp-transfer-state.ts`: both `terminal` and `cancelable` derived from the shared records; the local record is deleted.

**before** — five lists, drift by omission:

1. Each surface keeps its own status array/set.
2. A new `PaymentTransferStatus` compiles everywhere; any list that forgot it silently treats the status as non-terminal or non-cancelable (webhook events keep mutating a settled transfer, or the dashboard offers cancel on a transfer the API refuses).

**after** — two records, forced decisions:

1. All surfaces call the shared lookups.
2. A new status fails `sdp-types` compilation until its terminality and cancelability are declared, then every surface picks both up.
3. The intentional deviation (`expired` revivable in settlements) is an explicit one-line carve-out at its call site, not a divergent copy.

**Verification**

- `pnpm turbo typecheck --filter @sdp/types --filter @sdp/api --filter sdp-web` — 21 tasks pass
- `vitest run` on `ramp-transfer-state` (sdp-web) — 11 tests pass
- `vitest run` on `ramp-settlements`, `payments.ramps`, openapi routes (sdp-api) — 11 + 40 + snapshot tests pass
- Behavior-preserving: each derived set is value-identical to the list it replaces

### DIFF
--- a/apps/sdp-api/src/routes/payments/handlers/ramps.ts
+++ b/apps/sdp-api/src/routes/payments/handlers/ramps.ts
@@ -37,7 +37,9 @@ import type { RampRuntimeContext } from "@sdp/payments/ramps/types";
 import { redactCredentialString } from "@sdp/redaction";
 import { parseDecimalAmount } from "@sdp/solana/amount";
 import {
+  CANCELABLE_RAMP_TRANSFER_STATUSES,
   getCryptoRailAssetLabel,
+  isCancelableRampTransferStatus,
   isCountryCode,
   type PaymentRampEstimate,
   type PaymentRampInstruction,
@@ -1417,8 +1419,7 @@ export async function cancelRampTransfer(c: ValidatedBodyContext<typeof cancelRa
   if (!isRampTransferType(transfer.type)) {
     throw badRequest("Only ramp transfers can be canceled through this endpoint.");
   }
-  const cancelableStatuses: readonly PaymentTransferStatus[] = ["pending", "awaiting_payment"];
-  if (!cancelableStatuses.includes(transfer.status)) {
+  if (!isCancelableRampTransferStatus(transfer.status)) {
     throw badRequest(`Transfer can no longer be canceled (status: ${transfer.status}).`);
   }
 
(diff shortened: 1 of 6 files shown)


resourcefulmind (approved):
Checked every status against the old lists at all five places and nothing changed, expired carve-out included. The red policy check is from an earlier run of the same commit; the later one passed. 
Can you also check later that the BVNK and Mural webhook handlers still keep their own status lists.

PR #1648, opened 2026-09-04
ci(smoke): stage-only smoke workflow on api-preview.solana.com

niks3089:
- Renames `sdp-dev-smoke.yml` → `sdp-stage-smoke.yml` and removes the `dev` target entirely: the smoke suite's only callers are the stage deploy and the prod promotion gate, both already stage-scoped. The filename and dev default were vestiges of when this workflow gated api-dev (retired in #1622).
- Points the probe/build at the canonical `https://api-preview.solana.com` (was `api-stage`, which remains a serving alias).
- Deletes the dead `Private-channels vendor sandbox smoke` step — it only ran for `target: dev`, which nothing passes since #1622. If that vendor check should live on, it needs a new stage- or local-CI home (flagging, not deciding here).
- Updates both callers, the `deploy.yml` path filter, and the workflow contract test.

**Verification**

- `curl https://api-preview.solana.com/health/ready` — 200; `api-stage` alias also 200
- `actionlint` — clean on all four touched workflows
- `node --test scripts/deploy-sdp-api-gcp-prod-workflow.test.mjs` — 22/22 pass


G1de0n (commented):
Clean rename, callers, path filter, and contract test all updated. One blocker before merge: STAGE_SMOKE_READY isn't set anywhere in the repo, just checked. So as of this PR the smoke job goes green having run exactly zero probes, and that's the only gate for both prod promotion on the release path and the orchestrator's stage bake. Stage is already serving 200s, so I'd flip the var in the same commit as this merge. Either way, the job needs to fail loudly when the var's unset instead of quietly skipping to green, otherwise a deleted var can silently disarm the gate again.

G1de0n (approved):

PR #1654, opened 2026-09-04
feat(payments): remove MagicBlock private transfers

arseniy-nikitochkin:
> **Stacked PR.** Base this on HOO-1017 fix, not `main`, until that PR merges —
> it rewrites the same two files. Once HOO-1017 lands, rebase and retarget `main`.

SDP integrated exactly one MagicBlock product: a hosted private SPL transfer API.
Passing `privateTransfer` to `POST /v1/payments/transfers` asked
`payments.magicblock.app` to build a transaction that routed the movement through
MagicBlock's own on-chain programs, swapped the provider's sponsor slot for the
Kora fee payer, signed it with custody keys, and submitted it. That integration is
withdrawn, so the provider client, its types, its schemas, and its execution path
go with it.

Net **−2,353 / +72** lines across 20 files. No migration: MagicBlock never added
anything to the schema.

## The field rejects rather than disappears

This is the one part worth reviewing closely. `createTransferSchema` is **not**
`.strict()`, so zod strips unknown keys. Deleting `privateTransfer` outright would
have let an old caller's private-transfer request fall through and execute as an
ordinary **public** transfer — publishing on-chain exactly what the caller asked to
keep private, with a `200` and no indication anything had changed.

So the key stays in the schema purely to fail:

```ts
privateTransfer: z
  .undefined({ message: "privateTransfer is no longer supported: …" })
  .optional(),
```

A test pins it: `400`, the message names the field, `createOrgSignerMock` and
`sendTransactionMock` never fire, and `payment_transfers` stays empty. The OpenAPI
request body `.omit()`s the key entirely, so generated clients see no such property.

## The fingerprint keeps a vestigial key

`buildPaymentTransferFingerprint` still serializes `privateTransfer: null`. Nothing
passes it any more, but removing it would rewrite **every** stored fingerprint — an
`Idempotency-Key` recorded before this deploy would stop matching its own row, and
the retry it exists to absorb would transfer a second time. A test asserts the
byte-exact serialization so a future cleanup pass can't quietly break replay, and
the interface field carries a comment saying why it survives.

## Removed

| Area | What went |
|---|---|
| Handler | the `if (privateTransfer)` branch + 11 MagicBlock helpers (~665 lines) |
| Service | `apps/sdp-api/src/services/private-transfers/` (3 files) |
| Types | `sdp-types/private-transfers.ts`, its barrel export, `PreparedPrivateTransfer` and the request/response fields |
| Schemas | runtime option schemas; OpenAPI request schemas, example, and `preparedPrivateTransferSchema` |
| Env | `MAGICBLOCK_PRIVATE_PAYMENTS_{API_BASE_URL,AUTH_TOKEN,EPHEMERAL_RPC_URL}` from `env.d.ts`, `turbo.json`, `secret-keys.mjs`, `.env.example` — the third was never read by any code |
| Infra | the two MagicBlock program IDs from `infra/kora/kora.toml` `allowed_programs` |
| Tests | 14 MagicBlock tests + their fixtures; `magicblock.test.ts` wholesale |
| Generated | Postman collection + API playground catalog regenerated |

Two removals the plan didn't anticipate, both orphaned by the change:

- **`enforcePaymentTransferOperationPolicy`** — its only caller was the multi-signer
  private-transfer path. The ordinary path is gated by the policy middleware via
  `extractTransferPolicyCandidate`, so this was dead once the branch went.
- **the `privateTransfer` policy-candidate context field** — no policy rule reads it
  (checked `packages/sdp-policy/src` and `services/policy`: zero references).

Also gone: `assertMagicBlockTransactionMatchesOperation` and the rest of the
hostile-provider hardening added in the HOO-1017 commit below. It has nothing left
to guard once no provider bytes are signed — the hardening dies with the feature it
protected, which is the intended outcome, not a regression.

## Kept deliberately

- **`transfer_confidential`** as a transfer type in `payments.repository.ts`, the
  list-filter enums, the OpenAPI type enum, and the dashboard's read-only
  "Confidential" label. Historical rows keep that type and must stay queryable and
  renderable.
- **Kora**, entirely. It is the platform-wide fee-payment provider (~100 files: earn,
  issuance, recurring payments, custody signer-check). MagicBlock was one consumer;
  only the two allow-list lines were provider-specific.
- **Private Channels (SPC)**, untouched. Independent feature, zero imports either
  way — the names are similar and they shared an audit, nothing more.
- `CHANGELOG.md` and `docs/prd-artifacts/*` mentions, as historical records.

## Breaking change

`privateTransfer` on `POST /v1/payments/transfers` is no longer accepted and returns
`400`. Private transfers are not available through any other endpoint. The commit
carries a `BREAKING CHANGE:` footer for release-please.

The route is API-key authenticated and the field was documented in the public
OpenAPI spec, so this is a real external break — callers get a clear error naming
the field rather than silent misbehavior, but **this assumes product sign-off on
ending the MagicBlock relationship.**


multipletwigs (dismissed):

multipletwigs (dismissed):
lfg

resourcefulmind (approved):

PR #1657, opened 2026-09-04
feat(earn): Jupiter Lend USDT

cutbow7:


packages/sdp-jupiter-lend/src/client.ts
```diff
@@ -0,0 +1,431 @@
...
+    }
+    const owner = publicKey("owner", input.owner);
+    const asset = new PublicKey(JUPITER_LEND_USDT.assetMint);
+    return this.withConnection(
+      ctx,
+      "Reading the Jupiter Lend position",
+      async (connection, assertActive) => {
+        try {
+          const [position, context] = await Promise.all([
+            getUserLendingPositionByAsset({ user: owner, asset, connection, market: "main" }),
```
keeganthomp: [P1] Preserve unavailable balance semantics

In `@jup-ag/lend@0.2.0`, this helper catches every `getTokenAccountBalance` failure and returns `0`. A transient RPC error while hydrating an existing position therefore becomes a real zero below, so position reads show zero shares/value and earnings can report a full loss instead of omitting unavailable live fields. Read the jlUSDT ATA through an error-preserving path and treat only a confirmed missing account as zero.

packages/sdp-jupiter-lend/src/client.ts
```diff
@@ -0,0 +1,431 @@
...
+          if (position.lendingTokenShares.isZero() && input.providerReferences.length === 0)
+            return [];
+          const shares = fromAtoms(position.lendingTokenShares, JUPITER_LEND_USDT.decimals);
+          return [
+            {
+              providerReference: JUPITER_LEND_USDT.assetMint,
+              owner: owner.toBase58(),
+              cluster: "mainnet-beta",
+              shares,
+              withdrawableShares: shares,
```
keeganthomp: [P1] Cap shares by current withdrawal liquidity

Jupiter exposes a live withdrawal limit (`liquiditySupplyData.withdrawable`), but this reports the entire jlUSDT balance as immediately redeemable. When utilization or withdrawal limits bind, the API offers a max withdrawal that the program will reject, even though `withdrawableShares` promises the amount redeemable now. Derive and cap this value from current protocol liquidity, and reflect the same constraint in withdrawal quotes.

packages/sdp-jupiter-lend/src/client.ts
```diff
@@ -0,0 +1,431 @@
...
+              tokenValue: fromAtoms(position.underlyingAssets, JUPITER_LEND_USDT.decimals),
+              tokenMint: JUPITER_LEND_USDT.assetMint,
+              shareMint: context.fTokenMint.toBase58(),
+            },
+          ];
+        } catch (error) {
+          if (error instanceof SdpJupiterLendError) throw error;
+          throw new SdpJupiterLendError(
+            "MARKET_UNREADABLE",
```
keeganthomp: [P2] Use the shared unreadable-vault error code

`rethrowVaultProviderFailure` maps only `VAULT_UNREADABLE` to the retryable 503 contract. Wrapping Jupiter RPC failures as `MARKET_UNREADABLE` leaves deposit and withdrawal previews/builds to fall through as generic 500s, so clients lose the provider-unavailable retry signal. Use the shared code or extend the bridge for this provider code.

keeganthomp (changes requested):
The Jupiter transaction construction and slippage floors look sound, but the live position adapter can turn RPC failures into zero balances and overstates immediately withdrawable shares. Those need fixes before enabling production deposits.

keeganthomp (approved):

PR #1664, opened 2026-09-05
fix(issuance): HOO-1013 harden pause state, deploy claims, metadata origins, and provider actions

arseniy-nikitochkin:
Fixes the eight confirmed DeepSec issuance findings (HOO-1013): pause-state writers, deploy-confirmation races, request-derived metadata origins, authority backfill erasure, active-content links, token path injection, and Vulcan Forge residue.

**Pause state**
- `status` is removed from the token PATCH schema and `UpdateTokenInput`: the settled `tokens:admin` pause/unpause paths are now the only writers of `issued_tokens.status`.
- Workflow `pause`/`unpause` actions record their ledger transaction (`recordWorkflowTransaction` now returns the row id) and mirror through `applySettledTokenStatus` — slot-ordered, once-only, tenant-scoped — instead of blind `updateToken({status})` writes; the `alreadyPaused`/`alreadyActive` converge branches write nothing.

**Deploy confirmation**
- `deploy/confirm` takes the same atomic `pending→deploying` claim as the custodial path; the raced caller gets 409, and a failed verification releases the claim so retries re-claim (including the retryable `not_indexed` case).
- `setTokenDeployed` commits only a claimed row with no recorded mint — an established mint address and its authorities can never be overwritten.

**Metadata & inputs**
- The SDP-hosted metadata origin comes only from `PUBLIC_API_ORIGIN` (validated http(s), normalized via `URL.origin`); the request-Host/`x-forwarded-proto` derivation is deleted and deploys fail closed on missing or malformed config.
- Authority backfill treats a missing mint account or unparsed RPC payload as `SOLANA_RPC_ERROR` and never overwrites a stored `metadata_authority` with NULL; a null read falls back to the stored value.
- `uri`/`imageUrl` restricted to http(s) and max 512 chars (create + update); asset-profile link keys (`asset.website`, `*url|uri|link|logo|image|icon`) must be bounded http(s) URLs; the symbol regex bars `..` shapes; `canonicalMetadataUrl` percent-encodes the token id.
- Orphaned Vulcan Forge (AlphaLedger) i18n keys removed from the es/fr/pt catalogs (the engine itself was deleted in #1203; localized-only edits pass the catalog gate).

**before** — pause state had three writers:

1. Admin pause/unpause: on-chain tx, then slot-ordered `applySettledTokenStatus`.
2. `PATCH /v1/issuance/tokens/:id` with only `tokens:write`: flipped `status` between `active`/`paused` directly — no admin permission, no on-chain transaction, desynchronizing the DB flag from the Token-2022 pausable extension in both directions.
3. Workflow pause/unpause: unordered `updateToken({status})` after the chain call — a slow rule tick could silently reverse a newer manual unpause, and the already-paused branch wrote status with no settled evidence at all.

**after** — one write mechanism:

1. Admin path unchanged.
2. PATCH: `status` no longer exists in the schema or service input; the stored value is untouched.
3. Workflow actions: chain effect → ledger row (confirmed, with slot) → `applySettledTokenStatus(txId, tokenId, status)` under the same newer-transaction ordering and once-only marker; converge branches leave the DB alone.

**before** — non-custodial deploy confirm was read-then-write:

1. Read token, check `pending` + no mint in process memory.
2. Several RPC round-trips (verify landed, link signature→mint, recompute authorities).
3. `setTokenDeployed` ran an unconditional UPDATE — a second confirmation racing through steps 1–2 overwrote the winner's mint address, mint/metadata/freeze authorities, and ABL list address.

**after** — claim, verify, guarded commit:

1. Read token → friendly 400s for already-deployed/non-pending.
2. `beginTokenDeploy` atomically claims `pending→deploying`; the loser 409s.
3. Verification failures release the claim (guarded on `deploying`/no-mint) so a legitimate retry re-claims.
4. Commit requires `status = 'deploying' AND mint_address IS NULL`; zero rows → CONFLICT, never an overwrite.

**Operational prerequisite**

- `PUBLIC_API_ORIGIN` must be set to the public API origin in every deployed environment (Doppler) **before** this rolls out. When unset, only deploys of tokens without an issuer-supplied `uri` fail — closed, with a 500 naming the variable; everything else is unaffected. Local dev is covered by `.env.local.example`.


arseniy-nikitochkin:
@greptileai review pls

G1de0n (approved):

PR #1669, opened 2026-09-07
feat(dvp): deliver a leg to an address other than the one that funded it

tobySolutions:
A trade can now deliver a leg to an address other than the one that funded it.

The program always took `user_a_settlement_destination` and `user_b_settlement_destination`, and everything downstream already read them — `settle-preflight` derives the ATAs, `settle-instructions` wires them into `SettleDvp`. Only `create.ts` dropped them, hardcoding `null`. Settling into an account other than the one you funded from is ordinary for an execution desk.

Both are now optional on create, defaulting to the party.

**Two things worth review**

- A named destination is checked before signing: an address owned by a token program is a mint, token account or multisig, none of which can hold the ATA settlement pays into. Getting this wrong strands a trade *both* parties already funded, since settle moves both legs at once. Ownership decides it — a size check was wrong, because a Token-2022 mint with extensions is 368 bytes, not 82.
- A destination that differs from its party is also what a forged trade looks like, so the detail page names both addresses instead of rendering the redirect quietly.

Both idempotency fingerprints gained the fields; without that, two trades differing only in payee share a key and the second is refused as a mismatched replay.

**Verification**

- `vitest run src/services/dvp/` — 136 passed; sdp-web DvP — 167 passed; full web suite 1936 passed.
- `tsc --noEmit` clean in both apps; `check:module-boundaries` and `biome check` clean.
- Mutation-checked both guards: breaking each fails exactly its own tests.
- Local API boots with the stack merged; `/v1/dvp/trades` 401, unknown route 404.


tobySolutions:
@greptileai review

multipletwigs (approved):

tobySolutions:

PR #1676, opened 2026-09-07
feat(smoke): stage smoke signs in via Clerk ticket flow on the canary identity

niks3089:
First half of Andrey's #1648 ask (the `STAGE_SMOKE_READY` flip) needs the authed suite to actually be able to sign in on stage — stage runs the **production** Clerk instance, which rejects `@clerk/testing` testing tokens. This PR makes the suite stage-capable:

- **Ticket-flow sign-in**: when the publishable key is `pk_live_*`, global setup mints a 5-minute sign-in token via Clerk's backend API for the resolved canary admin and completes `/sign-in?__clerk_ticket=…` in the browser — the exact flow the prod canary has run every 5 minutes since Aug 28 (`sdp-infra observability/canary/tests/prod-authed.spec.ts`). Dev-instance runs keep today's `clerkSetup` path untouched.
- **Secrets, least-privilege**: the workflow's Doppler login moves to a dedicated identity (`github-sdp-stage-smoke-ci`, OIDC sub pinned to `refs/heads/main`, project viewer scoped to the `stg` environment only) and reads a new `stg_ci` branch config (inherits stg's prod Clerk keys + Loki creds; adds `E2E_CLERK_ORG_ID`/`E2E_CLERK_TEST_EMAIL` for the SDP Canary org). Prod Clerk secrets are therefore **not** reachable from PR-triggered workflows — the existing `dev_ci` identity stays dev-scoped. Repo var `DOPPLER_STAGE_SMOKE_IDENTITY_ID` is already set.

**before** — smoke on stage:
1. `STAGE_SMOKE_READY` unset → build/suite skip → gate passes on the readiness probe alone
2. even if flipped, `clerk.signIn` would fail against the prod instance — no working auth path existed

**after**:
1. global setup resolves the canary admin in the SDP Canary org (`org_3IVBeDF5…`), signs in via ticket, pins the org and project cookie, saves storage state
2. the read-only suite runs against `api-preview.solana.com` as a real authenticated session

**Remaining before flipping `STAGE_SMOKE_READY=true`** (not in this PR):
- `E2E_SDP_PROJECT_ID` in `stg_ci`: the canary org's project id as it exists in the stage DB (one SQL lookup; I need a fresh `gcloud auth` to run it)
- one dispatched proving run of the workflow, then flip the var
- the fail-loud-when-unset gate change is Opeyemi's sub-issue under PRO-1767 and rides separately

**Verification**
- `actionlint` — clean
- ticket flow itself — running in prod canary 5-minutely since Aug 28 (12/12 green)
- Doppler: `stg_ci` inherits `CLERK_SECRET_KEY`/`NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY`/`GC_LOKI_*` from stg — confirmed


niks3089:
@greptile review

apps/sdp-web/playwright/tests/auth.global.setup.ts
```diff
@@ -4,21 +4,47 @@ import { clerk, clerkSetup } from "@clerk/testing/playwright";
...
 setup("authenticate admin test user and save auth state", async ({ page }) => {
   setup.setTimeout(360_000);
   const env = getE2EEnv();
   const identity = await resolveClerkTestIdentity();
 
-  await clerkSetup({
-    publishableKey: env.clerkPublishableKey,
-    secretKey: env.clerkSecretKey,
-  });
+  if (env.clerkPublishableKey.startsWith("pk_live_")) {
```
G1de0n: The ticket flow is picked from env.clerkPublishableKey.startsWith("pk_live_"). If that key is ever misconfigured, the smoke quietly falls back to the @clerk/testing path and goes green without exercising the flow this PR is for. Can we pick the mode from an explicit flag, and hard fail when stage mode doesn't have a pk_live_ key?

apps/sdp-web/playwright/tests/auth.global.setup.ts
```diff
@@ -4,21 +4,47 @@ import { clerk, clerkSetup } from "@clerk/testing/playwright";
...
+      });
+      if (!response.ok) {
+        throw Object.assign(
+          new Error(`sign_in_tokens failed: ${response.status} ${await response.text()}`),
+          { status: response.status }
+        );
+      }
+      return (await response.json()) as { token: string };
+    });
+    await page.goto(`/sign-in?__clerk_ticket=${token}`, { waitUntil: "domcontentloaded" });
```
G1de0n: The sign-in token goes into the URL (/sign-in?__clerk_ticket=...). Traces are retain-on-failure and artifacts upload on always(), so a failed run ships a trace with a live token for the canary identity. The 300s expiry keeps the real risk low, but it's cheap to avoid: mask it, or sign in with Clerk.client.signIn.create({ strategy: "ticket", ticket }) instead of navigating with it. Worth wrapping it in encodeURIComponent while we're in there.

G1de0n (changes requested):

niks3089:
@G1de0n both addressed, thanks — good catches.

1. **Implicit mode → explicit flag.** Sign-in path is now gated on `E2E_CLERK_TICKET_AUTH` (set to `1` in `sdp-stage-smoke.yml`, preserved through `doppler run`), not the key prefix. `getE2EEnv()` **hard-fails** when that flag is set without a `pk_live_` publishable key, so a misconfigured stage key aborts loudly instead of silently taking the @clerk/testing path and going green. Dev runs (flag unset) are unchanged.

2. **Token no longer in the URL.** The ticket is completed in-page via `Clerk.client.signIn.create({ strategy: "ticket", ticket })` + `setActive({ session })`, with the token passed as a `page.evaluate` arg. It never appears in `/sign-in?…`, so a retain-on-failure trace can't ship a live sign-in token — which also resolves the two CodeQL 'file data in outbound request' findings on those lines (the token-bearing navigation is gone).

Verified: `actionlint` clean, `tsc --noEmit` clean for sdp-web.

niks3089:
@G1de0n both review points are in as of the last two commits: (1) mode is an explicit `E2E_CLERK_TICKET_AUTH` flag that hard-fails without a `pk_live_` key — no silent fallback; (2) the ticket never reaches the URL *or* the trace — greptile pointed out `page.evaluate` args are serialized into traces too, so the whole ticket exchange now runs in a manually created (untraced) context and only the resulting storage state is saved. Ready for re-review.

G1de0n (approved):

PR #1677, opened 2026-09-07
style(dvp): retone the trade copy

tobySolutions:
Retones the DvP trade copy.

The tell was structural, not lexical. Nearly every hint was a claim, then a sentence justifying the claim, then the consequence, in the same rhythm — so the page read as an essay about delivery-versus-payment rather than labels on a form.

28 strings rewritten, both em dashes gone.

**before**

> Atomic trades: both legs move together or neither does. Publish an escrow address, the counterparty sends an ordinary transfer, and settlement delivers each side at once.

**after**

> Trade two assets so both legs move together, or neither does. You publish an escrow address and the counterparty sends an ordinary transfer.

**Kept deliberately** — the hints carrying real settlement facts: the surplus warning, the frozen escrow, the settlement authority's SOL balance, and the forged-trade warning on redirected destinations. Trading an AI voice for a wrong one is not an improvement.

**Verification**

- 721 keys before and after; none added or removed. Diff is 28 lines.
- English catalog only, so the Translation Catalog Policy gate stays green.
- Full sdp-web suite: 212 files, 1941 tests passing. i18n catalog guard passing.
- No brace shape other than a plain interpolation name, which `translate()` would throw on at render.


multipletwigs (approved):

tobySolutions:

PR #1679, opened 2026-09-07
feat(dvp): trades where SDP sets the terms and holds neither leg

tobySolutions:
Adds trades where SDP sets the terms and holds neither leg.

An execution desk commonly sets up the on-chain swap details and then has two other parties do the swaps. The program never required otherwise: `CreateDvp`'s only signer is the payer, and both parties are plain non-signer accounts. The assumption was ours, and it lived in the schema.

**Principal trades are untouched and stay the default.** An omitted `tradeKind` still means principal, so existing callers are unchanged.

**before** — the schema forbade it: `sdp_side NOT NULL`, and create put the signer on whichever side it named.

**after** — two kinds, discriminated end to end:

1. Migration 0088: `trade_kind`, nullable `sdp_side`, and a CHECK tying them (principal requires a side, agent forbids one).
2. Create body is a discriminated union; agent takes `partyA`/`partyB` and no side.
3. `sdp_wallet_id` stays NOT NULL for both — that wallet still signs and pays fees and both escrows' rent. It just delivers nothing.
4. The create flow is staged (`Role → Parties → Legs → Terms → Review`) using the same `WizardFrame` as counterparty, ramps and private channels. Choosing the agent role swaps the counterparty field for two party fields and stops the leg cards saying YOU DELIVER / THEY DELIVER.
5. "Deliver somewhere else" is rebuilt as a per-party choice with the default stated, rather than a disclosure holding two unlabelled boxes.

## The sharp edge

`sdpLegOf` reads any side that is not `"a"` as `"b"`. A null side would have resolved to **leg B** and spent a custody wallet on a leg SDP neither holds nor owes. Guarded twice — the endpoint refuses agent trades outright, and `sdpLegOf` throws rather than defaulting.

## Why the form asks both parties to verify

A PDA's derivation covers only its seeds, not the fields inside it, so a trade's address does not bind its amounts. On an agent trade neither party was present when the terms were set, so both must check the stored terms before funding. The form says so instead of leaving it implicit.

**Verification**

- API: 32 files / 265 tests across dvp services, routes, repository and migrations.
- Web: 34 files / 454 tests. 7 tests on the parties hook cover both shapes, the same-address refusal on each, and that switching kind re-decides readiness.
- Mutation-tested the funding guards: removing them fails exactly those three cases.
- Both paths walked end to end in a browser: stages gate correctly, review reached, zero page errors.
- A principal trade created and funded on devnet through this flow: escrow holds 1000 ATD against a 1000 target, `trade_kind` written as `principal`, destinations defaulted to the parties.


tobySolutions:
Screenshot 2026-09-07 at 18 24 54

Screenshot 2026-09-07 at 18 31 16

multipletwigs (approved):

tobySolutions:

PR #1680, opened 2026-09-07
feat(helius-rings): Upgrade Helius Rings to zolana 0.1.6-alpha, and make an identity mismatch recoverable

TarekkMA:
## What

Upgrades `@heliuslabs/zolana` from `0.1.2-alpha` to `0.1.6-alpha`, ports the Rings module to its new key model, and fixes three things the upgrade exposed. Devnet only, no migrations.

## The upgrade

`0.1.6-alpha` removes `WalletAuthority` and splits it into `WalletKeys` for spending and `ShieldedKeys` for reading, and moves `ProofService` onto `ZolanaClient`. `src/authority.ts` is deleted and replaced by `src/keys.ts`: `spendKeys` for the spend path, `readKeys` for sync.

One detail fails closed if missed: `buildRingLookupTableTransaction` now extends a new ring table with `SHIELDED_POOL_CPI_AUTHORITY` and both SPL token program ids, and `outer-tx-policy.ts` derives the table it will sign — without the matching change, ring bring-up refuses its own transaction.

## Prover mismatch

Ring-bound withdraw and transfer were failing on devnet as a bare `gateway_unavailable`, which reads as "retry later". The real cause was a prover build that predates `custom-ring` support rejecting the request with HTTP 400. The health probe now reports amber unless the prover explicitly advertises `custom-ring`.

## Retryability

A 4xx from the prover is deterministic, so `CLIENT_PROVER_HTTP` 4xx now maps to `config_error` instead of being retried forever; 5xx, 408 and 429 stay `gateway_unavailable`. The bridge still carries no upstream messages, but attaches a structured `{ upstream, status }` cause so a failure is nameable in a log.

## Identity mismatch, and the way out

A wallet whose material no longer derives its provisioned identity was failing sync with `conflict` on every dashboard visit, indefinitely. It is now detected by a `RINGS_IDENTITY_MISMATCH` discriminator rather than a message string, paused once, and later syncs short-circuit before reaching the gateway.

`POST /v1/helius-rings/wallets/:walletId/rekey` is the recovery: it republishes the owner's registry record under the keys the deployment derives now, making the wallet usable again and abandoning everything held under the old keys. Three safeguards:

- **The gate is the chain, not a stored flag.** The record must read `foreign` at request time. A working wallet reads `ours` and is refused; an unregistered owner is told to provision. This covers both shapes of the break — paused mid-life, and never provisioned because the registry already held a record for that owner.
- **The typed confirmation is enforced server-side.** The dialog asks for the wallet's name and the API compares it, so it cannot be skipped by not rendering the dialog.
- **The read position is dropped with the old identity.** Keeping `sync_cursor` and `last_indexed_slot` would start the new identity part-way through history and silently skip everything earlier.

On chain this is `buildRegistrationTransaction` with the newly derived address, the same call provisioning uses. Provisioning itself still refuses a mismatch exactly as before.

## Review focus

Re-key is destructive and irreversible: notes encrypted to the old keys stay on chain with nothing able to open them. Worth the most attention are the `status !== "foreign"` gate in `rekeyWalletIdentity` and the `status IN ('paused', 'pending')` compare-and-swap in `rekeyWallet` that keeps a healthy wallet out of reach of a replay.


apps/sdp-api/src/services/helius-rings/service.ts
```diff
@@ -290,6 +299,87 @@ export class HeliusRingsService {
...
+    }
+    if (input.confirmation.trim() !== wallet.name.trim()) {
+      throw new HeliusRingsError(
+        "invalid_input",
+        `re-keying discards whatever this wallet's published keys hold; confirm by sending its name, "${wallet.name}"`
+      );
+    }
+
+    const published = await this.gateway.readIdentity({ walletId: wallet.id, owner });
+    if (published.status !== "foreign") {
```
librelois: If the on-chain rotation succeeds but confirmation, the subsequent registry read, or `wallets.rekeyWallet` fails, the row remains paused with its old shielded address. On retry, `readIdentity` reports `ours`, so this guard rejects before the database can adopt the new identity. `provisionPrivateWallet` also returns early for paused wallets, leaving no API recovery path. Please allow a paused recovery to reconcile its stored identity when the registry already matches the derived keys, without sending another rotation. Add coverage for a landed rotation followed by a persistence failure and retry.

apps/sdp-api/src/services/helius-rings/service.ts
```diff
@@ -290,6 +299,87 @@ export class HeliusRingsService {
...
+      throw new HeliusRingsError(
+        "conflict",
+        published.status === "ours"
+          ? "this rings wallet already derives its published identity; there is nothing to re-key"
+          : "no identity is published for this owner; provision the wallet instead of re-keying it"
+      );
+    }
+
+    const rotated = await this.gateway.rekeyIdentity({ walletId: wallet.id, owner });
```
librelois: Check and claim an eligible wallet state before rotating

A wallet with mismatched keys can still have status `ready` before its first sync quarantines it. Calling this endpoint in that state passes the `foreign` check and rotates the registry, but `rekeyWallet` then returns null because its UPDATE only accepts `paused` or `pending`. The endpoint returns success with the old stored identity, and the next sync pauses the wallet. The database status guard cannot prevent a rotation that has already happened. Please claim an eligible recovery state before submitting the transaction and handle failed adoption explicitly. Add coverage for a `ready` row whose registry identity is `foreign`.

apps/sdp-api/src/db/repositories/helius-rings-wallet.repository.postgres.ts
```diff
@@ -165,6 +166,47 @@ export function createPostgresHeliusRingsWalletRepository(db: AppDb): HeliusRing
...
+        )
+        .first<Record<string, unknown>>();
+      return row ? mapRow(row) : null;
+    },
+
+    async claimWalletForRekey(input: HeliusRingsProjectScope & { id: string }) {
+      const row = await db
+        .prepare(
+          `UPDATE helius_rings_wallets
+              SET status = 'paused', updated_at = sdp_iso_now()
```
librelois, line 219: Keep recovery available for paused wallets without a shielded address

For a wallet that never provisioned, this claim changes its status from `pending` to `paused` while `shieldedAddress` remains null. If rotation fails, or lands but database adoption fails, that state persists. After a dashboard reload, both recovery entry points disappear: `PrivateWalletRow` renders `WalletIdentityCheck` only for `pending` wallets, and `ShieldedBalanceCard` returns the unprovisioned message before reaching its `paused` recovery branch. Retrying provisioning also returns early for a paused wallet, so the operator can no longer recover through the dashboard even though the API supports retrying.

Please keep the recovery action available for `paused` wallets with a null shielded address, and add UI coverage for a failed re-key followed by a reload.

apps/sdp-api/src/services/helius-rings/service.ts
```diff
@@ -1500,6 +1687,38 @@ export class HeliusRingsService {
...
+      if (!isRingsIdentityMismatch(error)) throw error;
+
+      // The mismatch is the finding worth surfacing, so the pause is
+      // best-effort: it is logged either way and the original error still
+      // reaches the caller.
+      const paused = await this.wallets.updateStatus({
+        ...this.tenant,
+        id: input.walletId,
+        status: "paused",
+      });
```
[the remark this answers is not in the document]
TarekkMA: Fixed in b85e9bb19. Confirmed: the pause was keyed only on the wallet id, so a mismatch that arrived after a re-key clobbered whatever the row had become. The quarantine is now a compare-and-swap pinned to `expectedShieldedAddress` — the identity the failed read was actually for — so a stale finding no longer applies. Covered by `does not re-pause a wallet whose re-key landed while the read was in flight`, which re-keys the row inside the `syncPhoton` call before throwing the mismatch.

apps/sdp-api/src/routes/helius-rings/index.ts
```diff
@@ -48,6 +49,11 @@ heliusRings.get("/wallets", requirePermissions("payments:read"), listRingsWallet
 heliusRings.post("/wallets", requirePermissions("payments:write"), createRingsWallet);
 heliusRings.get("/wallets/:walletId", requirePermissions("payments:read"), getRingsWallet);
 heliusRings.post("/wallets/:walletId/sync", requirePermissions("payments:write"), syncRingsWallet);
+heliusRings.post(
+  "/wallets/:walletId/rekey",
```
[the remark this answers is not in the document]
TarekkMA: Not doing this here, because the Helius Rings module has no OpenAPI surface at all: there is no `helius-rings` entry under `src/openapi/paths`, and none of its existing endpoints (wallet creation, sync, identity, operations) are registered either. Documenting only the re-key endpoint would make it the single documented route in an otherwise undocumented module. There is also no coverage test requiring every route to be registered, so nothing is currently asserting this gap. Worth doing for the module as a whole, as separate work — this is a devnet-only module whose routes validate with Zod rather than the OpenAPI registry.

apps/sdp-api/src/db/repositories/helius-rings-wallet.repository.postgres.ts
```diff
@@ -165,6 +166,47 @@ export function createPostgresHeliusRingsWalletRepository(db: AppDb): HeliusRing
...
+            WHERE id = ? AND organization_id = ? AND project_id = ?
+          RETURNING *`
```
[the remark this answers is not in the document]
TarekkMA: Fixed in 972d75395, though the severity is lower than P1. The observation is correct: the claim had no guard, so two concurrent requests could both claim and both reach the rotation. What that costs is narrower than it looks — the target identity is derived deterministically from (organizationId, projectId, walletId, owner), so both requests rotate to the same shielded address, and `buildRegistrationTransaction` returns undefined once the record already publishes those keys. The end state converges; the loser is a redundant identical write, not divergent state or extra abandoned balance.

Still worth closing, so the claim is now an optimistic compare-and-swap on the row as it was read before the registry call. Note that an expected-status check cannot work here: `paused` is both where a claim lands and where the common recovery starts, so a second claim from `paused` is indistinguishable from the first claim of an already-quarantined wallet. Guarding on the row being untouched picks a winner before either request reaches the chain, and any other concurrent write refuses the claim — the safe direction, since nothing irreversible has happened at that point. Covered by `lets only one of two concurrent re-keys reach the chain`.

apps/sdp-api/src/db/repositories/helius-rings-wallet.repository.postgres.ts
```diff
@@ -165,6 +168,66 @@ export function createPostgresHeliusRingsWalletRepository(db: AppDb): HeliusRing
...
+        .bind(input.id, input.organizationId, input.projectId, input.expectedShieldedAddress)
+        .first<Record<string, unknown>>();
+      return row ? mapRow(row) : null;
+    },
+
+    async claimWalletForRekey(input: ClaimHeliusRingsWalletInput) {
+      const row = await db
+        .prepare(
+          `UPDATE helius_rings_wallets
+              SET status = 'paused', updated_at = sdp_iso_now()
```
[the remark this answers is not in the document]
TarekkMA: Fixed in 12bf5dc15, and you were right on both counts — the premise and the prescription's limits.

The premise checks out: `sdp_iso_now()` is `to_char(timezone('UTC', now()), 'YYYY-MM-DD"T"HH24:MI:SS.MS"Z"')`, so the token is millisecond-precision and `now()` is transaction-start time. I reproduced the hole rather than reasoning about it: with the new lock bypassed, two concurrent re-keys both succeed and two rotations are sent.

A dedicated version token would not have been enough on its own, though. It closes the collision between requests sharing one read snapshot, but not the other interleaving: a rival that reads the row *after* the winner's claim commits gets a fresh token, claims successfully, and rotates while the first rotation is still in flight. Exclusion has to outlive the claim statement, which no guard on the row can do. A claim column could close it as a lease, but then a crash between claiming and releasing leaves the wallet unclaimable without a TTL or manual repair — the same class of dead end this PR already had to fix twice.

So the claim is now serialized with a per-wallet advisory lock via `lockedTransactionWithPostCommit`, the primitive `audit.service.ts` uses for this shape. The claim is the transaction and the rotation is the post-commit advance, so the lock is held across both: the row is committed before the chain is touched, a rival waits and is then refused by the row guard instead of rotating, and because the lock releases with the session a crash mid-rotation leaves nothing stuck. The row guard stays as the second line — it is what refuses the loser once it acquires the lock.

Covered by `serializes two genuinely concurrent re-keys into one rotation`, which fires two real concurrent calls and asserts exactly one succeeds and `rekeyIdentity` is called once. That test fails without the lock, which is how I confirmed it is testing the guarantee and not just passing.

apps/sdp-api/src/services/helius-rings/service.ts
```diff
@@ -290,6 +301,209 @@ export class HeliusRingsService {
...
+    }
+
+    let rekeyed: PrivateWallet | undefined;
+    await withRekeyLock(
+      `rings-rekey:${wallet.id}`,
+      async () => {
+        // Guarded by the row as it was before the registry read, so a rival
+        // that slipped in while this one was reading the chain loses here
+        // rather than rotating a wallet whose state it never saw.
+        const claimed = await this.wallets.claimWalletForRekey({
```
librelois: Avoid exhausting the database pool while waiting for re-key locks

`lockedTransactionWithPostCommit` reserves a connection from the shared pool for every caller, including callers blocked in `pg_advisory_lock`. This callback ignores the supplied transaction executor and calls `this.wallets`, which acquires another connection from the same pool; adoption in `afterCommit` also needs another connection.

With ten overlapping requests, the pool's ten connections can all be occupied by lock holders or waiters. The winning request then cannot acquire the connection needed to claim or adopt the wallet and hits the five-second connection timeout. Unrelated database requests can time out too. This is a pool-starvation issue separate from the previously reported duplicate-rotation race.

Please use the supplied executor for the claim and prevent lock waiters from consuming the connections needed to finish rotation and adoption, for example by using a nonblocking advisory lock and rejecting competing requests. Add coverage at the pool's concurrency limit; the current two-request test does not exercise this failure.

TarekkMA: Re-key now uses `pg_try_advisory_lock` and claims on the lock’s executor so waiters don’t occupy pool slots

librelois (approved):

snowmead (approved):

PR #1682, opened 2026-09-07
ci: unbreak the Doppler login and the Surfpool duplicate-submit path

G1de0n:
CI has been intermittently red across all PRs today. Two independent breakages, neither in anyone's product code.

## 1. Stale Doppler credential on the self-hosted runner

`doppler oidc login` refuses when its scope is already authorized from an earlier token:

```
Warning: This scope is already authorized from a previous token. Remove the existing token to authenticate via OIDC.
Exiting
```

It exits 0, so the `Doppler OIDC login` step is marked green while never having logged in. `pnpm test` then uses the leftover token and dies on `Doppler Error: Invalid Auth token`.

Self-hosted runners keep `$HOME` between jobs, so the first job to authorize the scope poisons every later job on that machine, and it does not recover on its own. That is why it looks like flakiness — it depends on which runner picks the job up. Failures today (#1681, #1680) ran on `gha-sdp-ci-1`; a run that passed at 15:04 happened to land on an ephemeral GitHub runner.

**Fix:** `doppler logout --scope=. --yes || true` before the login, in all three copies of the block. This makes the login idempotent and heals a runner already carrying a stale token, so no manual cleanup on `gha-sdp-ci-1` is needed — the first job that runs with this file clears it. Then `doppler me` after the login, because `oidc login` reports success when it did nothing: a failed login should fail the step that owns it, not the test step three steps later where the error no longer names the cause.

## 2. The Surfpool shim reported a landed transaction as a failure

Separate cause, same symptom of red CI. The Access and Kora integration suites failed with:

```
Kora Error -32000: Kora Surfpool shim request failed: Transaction verification failed for transaction
Internal error: "Transaction error: This transaction has already been processed"
```

`KoraFeePayment.signAndSend` retries on a timeout and re-sends identical bytes. That is documented as safe because ed25519 signing is deterministic, so the retry carries the same signature and the cluster deduplicates it, answering with that signature. Surfpool does not deduplicate — it errors — and the shim surfaced that as a failure. So an issuance deploy that had already landed came back as `503 NETWORK_ERROR`.

It surfaced now because the deploy took 13s on a slow runner, long enough to trip the adapter's timeout retry. Neither the shim nor the adapter has changed recently.

**Fix:** treat the duplicate as the success it is and return the signature, derived from the signed bytes exactly as the adapter derives it. The shim stands in for Kora, so it has to reproduce the cluster's semantics. The `Access` suite goes green on this alone.

The status wait before the shim's own resubmission stays at 3s, now with the constraint written down: it is bounded by the CALLER's timeout, not by how long a validator might take. `KoraFeePayment` gives signAndSendTransaction 10s (`DEFAULT_TIMEOUT_MS`), so a shim that waits longer guarantees the timeout-and-retry it would be trying to avoid — raising it to 15s in an earlier commit here turned every submit into `Kora signAndSendTransaction timed out after 10000ms`, which is what that commit's follow-up reverts.

The adapter's retry predicate is deliberately untouched: retrying a timed-out submit against a real cluster is correct and worth keeping. The defect was in the test double.

Covered by a new case in `kora-surfpool-shim.test.ts` — send, confirm, then send the same bytes again and assert the same signature comes back rather than an error.

## Alternative considered

Pointing the Doppler CLI at job-scoped state (`--config-dir "$RUNNER_TEMP/doppler"`) would make breakage 1 unrepresentable rather than merely repaired. Not done here because every later `doppler` invocation in the job would have to agree on the directory, including the ones inside `scripts/doppler/run-with-config.sh`. Worth doing separately if we want persistent runners to carry no Doppler state at all.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -243,17 +243,8 @@ jobs:
         run: |
           oidc_jwt=$(curl -sf -H "Authorization: Bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" \
             "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=https://github.com/solana-foundation" | jq -r '.value')
-          # Self-hosted runners keep $HOME between jobs, so a credential left by
-          # an earlier job makes `oidc login` refuse with "This scope is already
-          # authorized from a previous token" and exit 0 — a green step that
-          # never logged in, whose stale token then fails the test step with
-          # "Invalid Auth token". Clearing the scope first makes the login
-          # idempotent, and heals a runner already carrying a stale token.
           doppler logout --scope=. --yes >/dev/null 2>&1 || true
           doppler oidc login --scope=. --identity="${{ vars.DOPPLER_CI_IDENTITY_ID }}" --token="${oidc_jwt}"
-          # `oidc login` reports success when it did nothing, so prove the
-          # session actually authenticates rather than discovering it three
-          # steps later behind an error that no longer names the cause.
           doppler me >/dev/null
 
       - name: Unit tests (Doppler)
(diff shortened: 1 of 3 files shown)


niks3089 (dismissed):

multipletwigs (approved):

niks3089 (approved):
Both halves verified — approving.

**Doppler half**: diagnosis confirmed. `doppler oidc login` exits 0 on the already-authorized warning without logging in, and `gha-sdp-ci-1` persists `$HOME/.doppler` between jobs, so whichever job ran after the credential went stale inherited a dead token with a green login step. The `logout || true` + `doppler me` fail-loud pair is the right shape: idempotent on hosted runners, self-healing on ours. I've also wiped `~runner/.doppler` on the VM directly, so the queue unblocks even before this merges.

**Shim half**: `already been processed` → derive the signature from the signed bytes matches mainnet semantics for duplicate submits (kit's base64 codec → decoder → `getSignatureFromTransaction` chain is right), and the same guard on the resubmission path closes the race where the first send landed but its status poll missed. Integration test covers the dup path.

One systemic note: this class of bug (state leaking between jobs on a persistent runner) is exactly what sdp-infra#136 kills — its runners are one-job JIT ephemerals with nightly reimage, so `.doppler`, pnpm state, docker leftovers all die with the job. This PR is the correct immediate fix; #136 (5/5, awaiting human review) is the end state.

PR #1683, opened 2026-09-07
feat(dvp): tell a party which trades are waiting on them

tobySolutions:
An agent trade names two parties who were not present when it was created, so
neither learns it exists unless somebody sends them the address out of band.
This closes that gap and gives them the escrow to pay.

`GET /v1/dvp/trades/inbound` takes no parameters. The filter is the caller's own
custody wallet addresses, resolved server-side and scoped by the same api-key
wallet bindings `/v1/wallets` uses. A party address is the only input it could
take, and accepting one would turn it into an oracle for enumerating the trades
of any address on Solana.

**Migration 0089** adds a `FOR SELECT` policy so a named party can read the row
that names it. Writes do not move: the tenant policy keeps the only `WITH
CHECK`. The organization scoping comes from `custody_wallets`' own policy rather
than a predicate, which is subtle enough to be worth the comment it has.

The terms are already public on chain, so this discloses nothing a party could
not decode from the PDA. Everything *around* them is ours, and is withheld by a
serializer written from scratch — reaching for `toTradeResponse` and deleting
fields would leave the disclosure decision to whatever shape that response takes
next.

On the dashboard it sits above the list rather than behind a tab: it is the only
item on the page with an expiry running against it.

## Verification

- Cross-org boundary tested directly under the NOBYPASSRLS runtime role: a party
  reads, a stranger does not, a party still cannot write, no identity reads
  nothing.
- Leak guard asserts against the serialized JSON, so a field that arrives later
  under any name fails the test.
- 176 API and 208 web tests, typecheck, module boundaries, lint.

## Risk

The read policy is the first place a DvP trade crosses an organization
boundary. It is read-only and the funding half (PRO-1854) records its claim on
its own table rather than mutating this row.


multipletwigs (approved):

tobySolutions:

PR #1687, opened 2026-09-08
fix(custody): persist credential creation and reconcile shared GCP secrets

pashpashkin:
**What changed**

- Persist a `creating` Credential and its intended GCP location before writing the secret during setup or replacement.
- Finalize the Credential with the exact acknowledged GCP version, preserving recoverable state when a write or database commit has an unknown outcome.
- Reuse the predecessor's GCP container for GCP-backed replacements, while retaining a separate Credential ID and exact version reference for each attempt.
- Add recurring cleanup for retained secrets and unreferenced GCP versions, with one persisted scan schedule per container.
- Bound GCP requests, verify failed destroy outcomes, and require confirmed destruction before custody cleanup clears its marker.

**Why**

Writing to GCP before recording the attempt in PostgreSQL could leave a live secret without a durable recovery record.

Recurring container inventory also handles versions whose write responses were lost and which become visible after the original request has failed.

**Impact**

Existing setup and replacement use the new GCP persistence protocol. Managed secret payloads remain in GCP; the existing encrypted-database backend remains supported.

Migrations `0080`, `0082`, `0083`, and `0084` add Credential creation, retention, cleanup metadata, and container scheduling state.

Privy `appId` and `appSecret` now have a 4,096-character limit, enforced before secret-store I/O.

Cleanup is integrated into the existing managed job and self-hosted cron. Managed cleanup runs alongside the existing reconciliation sequence; both settle before collected failures are reported.

**Pre/post release actions**

1. Apply the included migrations before deploying the new writers and cleanup code.
2. Before managed rollout, confirm version inventory and metadata access (`secretmanager.versions.list` and `secretmanager.versions.get`), alongside the existing secret read/write/destroy permissions.
3. After rollout, monitor cleanup failures, deferred scans, and confirmed-destruction logs. Validate setup/replacement recovery in an isolated GCP environment.

**Result Validation**

- `pnpm --filter @sdp/api typecheck` and `pnpm --filter @sdp/api build` — passed.
- 315 targeted tests — passed, covering migrations, setup/replacement, lost responses, cleanup races, scheduling, RPC, workflow retirement, and job integration.
- Migration tests ran against test PostgreSQL.

**Does it affect current flows & behavior and how**

**Before — GCP:** write the secret → insert the Credential/Connection → attempt immediate compensation if persistence fails.

**After — GCP:**

1. Commit the Credential reservation and, for a new container, its scan schedule.
2. Write one GCP version.
3. Persist its exact reference and finalize the pending Credential and Connection.
4. On failure, reconcile the result or retain durable recovery work. A canceled writer cannot finalize later.

Each scan cancels stale creation attempts, reads version metadata, and refreshes database references after each inventory page.

Active/pending credentials, versions still required by Connections, and eligible rollback predecessors are protected. A failed Connection retained after rejected installation does not prevent cleanup of the rejected secret.

Empty inventory or a first-page `404` does not stop future scans. Backend changes can create another container; each existing container keeps its own scan owner.

RPC and workflow do not adopt custody's shared-container reconciliation. Their shared GCP client does gain 10-second per-request timeouts. Successful destroy requests retain their previous default semantics; failed requests are accepted only if the exact version is confirmed `DESTROYED`. Workflow retirement no longer treats `FAILED_PRECONDITION` alone as proof of destruction.

**Known gaps**

- Cleanup does not delete GCP containers or revoke credentials at the provider.
- The cleanup deadline bounds admission and HTTP work, not an outstanding SQL wait.
- Repeated slow destruction attempts can delay inventory within a container. Its persisted schedule remains available for later scans.
- Real GCP behavior and permissions remain unverified.

**Notes**

Part 2 of 4, based on the entitlement PR. Existing setup/replacement are its immediate consumers; rotation and rollback arrive in the next PR.

Container scanning uses a fixed five-minute revisit schedule, not an assumed-absence cutoff.

### DIFF
--- a/apps/sdp-api/src/cron/provider-credential-secret-cleanup.test.ts
+++ b/apps/sdp-api/src/cron/provider-credential-secret-cleanup.test.ts
@@ -1,0 +1,41 @@
+import { describe, expect, it, vi } from "vitest";
+import type { BackgroundRunner } from "@/runtime/background";
+import type { Observability } from "@/runtime/observability";
+import { cleanupRetiredProviderCredentialSecrets } from "@/services/jobs/cleanup-provider-credential-secrets";
+import type { Env } from "@/types/env";
+import {
+  PROVIDER_CREDENTIAL_SECRET_CLEANUP_CRON,
+  PROVIDER_CREDENTIAL_SECRET_CLEANUP_MONITOR,
+  runProviderCredentialSecretCleanup,
+} from "./provider-credential-secret-cleanup";
+
+vi.mock("@/services/jobs/cleanup-provider-credential-secrets", () => ({
+  cleanupRetiredProviderCredentialSecrets: vi.fn(async () => ({
+    cleaned: 0,
+    skipped: 0,
+    failed: 0,
+  })),
+}));
+
+describe("runProviderCredentialSecretCleanup", () => {
+  it("runs the cleanup through its five-minute monitor and background tracker", async () => {
+    const env = {} as Env;
+    const bg = { run: vi.fn(), awaitAll: vi.fn(), draining: false } satisfies BackgroundRunner;
+    const observability = {
+      captureException: vi.fn(),
+      withScope: vi.fn(),
+      withMonitor: vi.fn((_slug, work) => work()),
+    } satisfies Observability;
+
+    runProviderCredentialSecretCleanup({ env, bg, observability });
+
+    expect(observability.withMonitor).toHaveBeenCalledExactlyOnceWith(
+      PROVIDER_CREDENTIAL_SECRET_CLEANUP_MONITOR,
+      expect.any(Function),
+      { schedule: { type: "crontab", value: PROVIDER_CREDENTIAL_SECRET_CLEANUP_CRON } }
+    );
+    expect(cleanupRetiredProviderCredentialSecrets).toHaveBeenCalledExactlyOnceWith(env);
+    expect(bg.run).toHaveBeenCalledOnce();
+    await vi.mocked(bg.run).mock.calls[0][0];
+  });
+});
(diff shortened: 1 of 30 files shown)


pashpashkin:
@greptileai review pls

GuiBibeau (dismissed):
Reviewed cc381ed. The secret-reference compatibility blocker is resolved: cleanup recognizes project-ID and project-number aliases, and the destruction guard protects duplicate references across those aliases. I reran the original controlled reproduction against the updated cleanup function; both active-reference formats are preserved. The new regression tests cover active-version preservation and alias duplicate protection.

The PR merges cleanly with current main. CI is green, including 344 test files with 4,846 passing tests (10 skipped). No remaining merge blockers found. Apply the migrations and IAM permissions before rollout; live GCP/provider smoke validation remains a deployment check.

multipletwigs (dismissed):

pashpashkin:
@greptileai rereview pls, not sure I understand Helius Rings migration is connected to the scope of the current pr

resourcefulmind (approved):
The reserve, write, finalise ordering is right, and I checked it against the retirement sweeper from #1451: the two never touch the same container, so no race there. Approving
- A container with no scan owner never cancels a stale `creating` row. Rows inserted by the old revision during rollout have no owner, and a replacement into that container doesn't claim one, so if that replacement crashes between reserve and recovery, every later setup on the project 409s for good. Claim the owner on replacement, or re-run the 0084 backfill after deploy.
- The old "reports failed GCP cleanup without exposing the secret ref" test has no replacement, so that redaction isn't asserted anymore. Isn't it worth putting it back?

pashpashkin:
Thank you for the review @resourcefulmind!

- BYOK off will help avoid this rollout issue
- I will remove the secret ref from all credentials' cleanup jobs and return test coverage in the next PRs

PR #1696, opened 2026-09-08
ci: unit tests on hosted runners with a GitHub-cached turbo store

niks3089:
Kills PR queueing at the root. The single `sdp-ci` VM serializes every api-leg run — today's merge burst queued main pushes for an hour+ behind each other. Hosted runners give effectively unlimited concurrency; what they lacked was the persistent turbo cache, which `actions/cache` now carries:

- Both unit jobs run on `ubuntu-latest` (fork conditional gone — one code path).
- `TURBO_CACHE_DIR` points into the workspace and is restored/saved via `actions/cache` (keyed per job + os, `restore-keys` falls back to the latest, so PRs reuse main's cache). The merged #1689 script picks it up unchanged.
- Expected PR cost, unchanged `@sdp/api`: **zero queue** + setup (~60–90s) + cache restore (~10–20s) + FULL TURBO replay → **~2 min wall, no waiting**. Changed `@sdp/api`: the suite genuinely runs (~20 min on 4-core hosted — if that hurts, the follow-up is a GitHub larger-runner label for that one job, still zero-queue).
- The `sdp-ci` VM stays registered for now (deploy smoke etc. don't use it; nothing targets `sdp-ci` after this merges) — decommission follows separately.

Supersedes #1695 (its hosted-only cache tweak is moot once the job itself is hosted).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,12 +177,13 @@ jobs:
 
   test:
     name: Unit Tests
-    runs-on: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork && 'ubuntu-latest' || fromJson('["self-hosted", "sdp-ci"]') }}
+    runs-on: ubuntu-latest
     timeout-minutes: 40
     permissions:
       contents: read
       id-token: write
     env:
+      TURBO_CACHE_DIR: ${{ github.workspace }}/.turbo-ci-cache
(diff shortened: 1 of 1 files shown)


niks3089:
@greptile's cache-churn point is addressed: the combined cache step is now `actions/cache/restore` (always) + `actions/cache/save` gated to `refs/heads/main` — PRs create zero cache entries and reuse main's latest via `restore-keys`; main saves one entry per commit, which LRU handles. Measured on this branch already: packages leg **37s**, api leg replaying from cache (was 13–19 min cold), zero queue on hosted runners.

G1de0n (approved):

PR #1698, opened 2026-09-08
test: ci cache-replay canary

niks3089:
Measurement PR, close after reading: touches `scripts/doppler/` — inside the api path filter (so **Unit Tests runs**) but outside every turbo package (so `@sdp/api`'s hash is unchanged and must **replay from main's saved cache**). Pass bar: api leg ≈60–90s on a hosted runner with zero queue. If it misses, the optimization lands in this PR.


niks3089:
Measurement complete — cache replay verified: api leg **32s** (restored main's `turbo-test-Linux-1d459be…` cache), packages 36s, scope 6s, zero queue. Cold suite remains ~13 min and runs only when `@sdp/api`'s graph actually changes or on main pushes.

PR #1699, opened 2026-09-08
ci: api unit tests on PRs run only diff-related files, coverage on main only

niks3089:
Items 1+2 of the test-speed plan (items shipped as one because changed-file selection without coverage-off would fail thresholds on subsets):

- **PR api runs use `vitest --changed=origin/<base>`**: only test files whose import graph reaches the diff execute. A typical api PR runs a handful of files (seconds–2 min) instead of all 328. `--passWithNoTests` covers PRs whose api-path trigger (lockfile, CI plumbing) touches no api source at all.
- **Coverage (v8 + thresholds) now runs on main pushes only** — where the full suite still runs unconditionally, with the turbo cache save. Nothing merges past full coverage; a PR that misses an unimported coupling is caught one commit later on main.
- **Item 3 (sharding main) deliberately deferred**: main's ~13 min blocks nobody, and sharding complicates the cache for zero PR-facing gain.
- **Item 4 (slow-file surgery) — investigated, hypothesis disproven**: the top files (`payments.recurring` 86s, `payments.transfers` 76s, `helius-rings` 76s) contain **no timer loops** — they're 51–73 real integration tests each at ~1.7s/test, dominated by the per-test locked `TRUNCATE…CASCADE` + Redis sweep + reseed and real route work. That floor is the suite's real-DB design; lowering it means transaction-rollback test isolation — ticketable, not a quick fix. With this PR, that cost only lands on PRs that actually change api code, proportional to what they touched.

**Live measurement**: this PR touches `scripts/run-workspace-tests.mjs` (an api-path trigger), so its own Unit Tests job runs the new path — expected: vitest selects ~0 api test files for this diff and the job completes in well under a minute.

### DIFF
--- a/scripts/run-workspace-tests.mjs
+++ b/scripts/run-workspace-tests.mjs
@@ -121,9 +121,6 @@ try {
               ? [`--filter=...[${changedSince}]`, "--filter=!@sdp/api-integration"]
               : ["--filter=!@sdp/api-integration"];
     if (split === "api" && changedSince) {
-      // PR runs: only the test files whose import graph reaches the diff, no
-      // coverage. Every main push still runs the full suite with coverage, so
-      // an unimported coupling a PR misses is caught one commit later.
       await run("pnpm", [
         "--filter",
         "@sdp/api",
--- a/scripts/run-workspace-tests.mjs
+++ b/scripts/run-workspace-tests.mjs
@@ -1,4 +1,4 @@
-import { spawn } from "node:child_process";
+import { execSync, spawn } from "node:child_process";
 import fs from "node:fs";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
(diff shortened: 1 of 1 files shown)


niks3089:
@greptile right, and it went one deeper: `--changed` also can't see through **built workspace packages** (`@sdp/*` imports resolve to `dist`, not the changed source), so packages diffs were unsafe for subset selection too, not just the lockfile. The subset path is now gated on the diff being entirely under `apps/sdp-api/` (files vitest loads directly), `--passWithNoTests` is gone (an api-scoped diff selecting zero tests would be suspicious, and non-api triggers no longer reach this path at all), and every other trigger — lockfile, packages, turbo/CI plumbing — falls back to the full turbo suite where the content hash decides honestly: dependency bumps bust the hash and run everything; plumbing-only diffs replay from cache in ~30s.

niks3089:
Closing by choice, not by finding: main provably runs the full suite unconditionally (push events hardcode both jobs on, and the subset path requires TEST_CHANGED_SINCE which only PRs set) — but vitest's `--changed` is blind to side-effect coupling (a migration or fixture change is api-scoped yet imported by no test, so it would select zero tests and pass), and Greptile's lockfile finding was one instance of that class, not the whole class. The banked wins don't need this PR: path-gate skips irrelevant PRs in ~10s, the turbo cache replays api-adjacent PRs in ~30s, and api-changing PRs pay the suite's honest cost. If PR-time subset selection returns, it should be allowlist-based (pure src/route+service diffs) with the side-effect surfaces (migrations, test infra, fixtures) explicitly forcing the full run — branch kept for that discussion.

PR #1700, opened 2026-09-08
fix(smoke): consume the ticket via URL in the untraced context

niks3089:
The stage smoke has failed every proving run at org activation: `Clerk.organization` stays `undefined` after `setActive`.

**Root cause**: #1676's in-page `signIn.create({strategy:"ticket"})` (adopted for the trace-leak fix) creates a session that activates but is **not org-capable** — `setActive({organization})` silently no-ops on it. The prod canary (`sdp-infra observability/canary/tests/prod-authed.spec.ts`, green every 5 min) instead lets Clerk consume the ticket from the **URL**, which builds the fully-populated session `setActive` can switch.

**Fix**: revert to the URL-ticket flow — but run it on the **untraced context** that #1676 already introduced. Playwright traces only fixture contexts, so a token in that context's URL never reaches a retained artifact. Both concerns satisfied: proven org-capable session **and** no trace leak.

**Verification**: tsc/biome clean; the proving run after merge is the live check — this is the exact sequence the prod canary has run continuously since Aug 28.

### DIFF
--- a/apps/sdp-web/playwright/tests/auth.global.setup.ts
+++ b/apps/sdp-web/playwright/tests/auth.global.setup.ts
@@ -35,47 +35,15 @@ setup("authenticate admin test user and save auth state", async ({ page, browser
       }
       return (await response.json()) as { token: string };
     });
-    await target.goto("/sign-in", { waitUntil: "domcontentloaded" });
-    await target.waitForFunction(
-      () => Boolean((window as unknown as { Clerk?: { client?: unknown } }).Clerk?.client),
-      undefined,
-      { timeout: 120_000 }
-    );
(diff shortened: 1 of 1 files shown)


niks3089:
@greptile addressed: both the ticket-bearing `goto` and the session wait now run through a `redactTicket` wrapper that replaces the token with `[redacted-clerk-ticket]` in any thrown error, so a navigation failure can't surface the live ticket in the uploaded HTML report. Combined with the untraced context (no trace.zip) and Clerk stripping the ticket from the URL on consumption, the token has no path to a retained artifact.

GuiBibeau (approved):

PR #1703, opened 2026-09-08
fix(earn): enforce exit slippage floor (PRO-1861) + cross-tenant/authz test evidence (PRO-1860)

keeganthomp:
Two Earn threat-model remediations, split into two commits. They share no files; happy to split into separate PRs if preferred.

## PRO-1861 - exit slippage floor (code + tests)

The wire contract documents a typed 400 when a provider with a non-null `withdrawalSlippage` policy is sent a floor-less exit, but nothing implemented it: routes, the withdraw service, and the provider clients all treated `minAmountOut` as optional. Latent while Kamino (null policy) was the only shelf; **live now that `jupiter_lend` ships a non-null exit floor in production.**

Both custody (`POST /vault-withdrawals`) and external-wallet (`POST /external-wallet/withdrawal-transactions`) builds now refuse a missing `minAmountOut` with a caller-fixable 400 when `earnWithdrawSlippageFloor(provider)` is non-null. Derived from the exit preview, so it never traps a position (ADR 0002).

## PRO-1860 - cross-tenant / authz evidence (tests only)

The threat model's "do first" isolation gap. Most earn route tests ran with `permissions: ["*"]`, so nothing proved a read-only key is refused on a money route.

- **Route-inventory ratchet**: a new `/v1/earn` route fails until it declares its scopes, plus per-route scope conformance, the earn:read-only-cannot-reach-money guarantee, two-sided `x-project-id` (key inert / session membership-checked), and the feature gate.
- New shared fixture `test/helpers/earn.ts` (KV + `api_keys` double-write, session rows + cookie).
- Movements: external-wallet structural exclusion + `/movements/:id` no-match 404.
- Positions: sibling-project summary exclusion + `ownerAddresses` containment.

## Reviewer notes

- **PRO-1861 changes real behavior**: a floor-less production `jupiter_lend` exit now 400s where it used to build. Kamino (null policy) is unaffected. Verify that matches the intended contract.
- The matrix's `EARN_ROUTE_SCOPES` table is the source of truth the ratchet enforces; a new route must be added there.

## Verification

164 tests pass across the 5 touched suites; typecheck and lint clean. Base is current `main` (`84caa068a`).

### DIFF
--- a/apps/sdp-api/src/routes/earn.route-authz-matrix.test.ts
+++ b/apps/sdp-api/src/routes/earn.route-authz-matrix.test.ts
@@ -1,5 +1,6 @@
 import type { Permission } from "@sdp/types";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { getDb } from "@/db/client";
 import app from "@/index";
 import earnRoutes from "@/routes/earn";
 import { type EarnAuthzTenant, seedEarnApiKey, seedEarnAuthzTenant } from "@/test/helpers/earn";
@@ -261,7 +262,7 @@ describe("scope conformance: a key missing exactly one declared scope is refused
 });
 
 describe("x-project-id is inert for API-key callers, per route (EARN-027)", () => {
-  it("a sibling-project header changes nothing", async () => {
+  it("a sibling-project header changes nothing (per-route status ratchet)", async () => {
     const raw = rawKeyBySet.get(permissionSetId(ALL_EARN_SCOPES));
     if (!raw) throw new Error("missing key");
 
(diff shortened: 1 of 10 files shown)


apps/sdp-api/src/routes/earn/handlers/vault.ts
```diff
@@ -1118,6 +1119,18 @@ export async function extractEarnVaultWithdrawalPolicyCandidate(
   }
   const position = toVaultHolding(positionRow);
 
+  // The exit twin of the deposit's share floor, keyed on the provider's
+  // declared policy rather than the environment: a non-null
+  // `withdrawalSlippage` answers a floor-less body with a 400 (the wire
+  // contract the strategy row and the OpenAPI description already state).
+  // Not an admission gate — a caller-fixable 400, derived from the exit
+  // preview, so it can never trap a position.
+  if (body.minAmountOut === undefined && earnWithdrawSlippageFloor(position.provider) !== null) {
```
keeganthomp: [P2] Keep completed replays ahead of the live floor policy

This runs during extraction, before `findEarnVaultWithdrawalIdempotentKeyReplay`. If a provider moves from a null floor policy to non-null, an exact retry of a previously successful floorless withdrawal returns 400 instead of the recorded result, breaking the route's idempotency contract. Run this policy check after the durable replay lookup, for example in a `beforeEnforce` hook.

apps/sdp-api/src/routes/earn.route-authz-matrix.test.ts
```diff
@@ -0,0 +1,312 @@
...
+  it("a sibling-project header changes nothing", async () => {
+    const raw = rawKeyBySet.get(permissionSetId(ALL_EARN_SCOPES));
+    if (!raw) throw new Error("missing key");
+
+    for (const route of ROUTES) {
+      const withoutHeader = await requestAsKey(route, raw);
+      const withHeader = await requestAsKey(route, raw, {
+        "x-project-id": tenant.siblingProject.id,
+      });
+      expect(withHeader.status, route).toBe(withoutHeader.status);
```
keeganthomp, line 274: [P2] Assert the selected project, not only the status

This still passes if API-key requests start honoring `x-project-id`: unseeded GETs can return 200 for either empty project, while every POST uses `{}` and usually returns the same 400 before its handler. Since this is the EARN-027 tenant-isolation ratchet, seed project-distinct data on a valid project-scoped route and assert that the key still sees its pinned project's result.

keeganthomp (commented):
Self-authored PR, so GitHub does not allow an approval verdict. I found two non-blocking issues in the idempotency ordering and the new tenant-isolation ratchet. API typecheck and focused Biome checks pass; local route tests could not start without a container runtime, and GitHub Unit Tests were still running at review time.

apps/sdp-api/src/routes/earn/handlers/vault.ts
keeganthomp: Fixed in ba21bfe8c: moved the floor check out of the extractor into a `beforeEnforce` hook (`assertEarnVaultWithdrawalFloor`), so it runs after the gate's completed-replay exit. An exact retry of a recorded floor-less withdrawal now returns its recorded outcome even if the provider's policy later flips to non-null; the check only admits genuinely new work.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/sdp-api/src/routes/earn.route-authz-matrix.test.ts
keeganthomp: Fixed in ba21bfe8c: added a data-distinct assertion. It seeds a movement for an owner claimed only in the pinned project, then hits `GET /external-wallet/movements` with `x-project-id` naming a sibling and asserts 200 with the row. Were the header honored, the owner has no claim in the sibling and it would 404. Kept the per-route status loop as a coarse ratchet.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

GuiBibeau (approved):

multipletwigs (approved):

PR #1707, opened 2026-09-08
feat(projects): metadata-based production project enablement

multipletwigs:
- Production projects are omitted from `GET /v1/projects` unless the organization is enabled for mainnet; enabled orgs see both default projects and can select production in the dashboard switcher.
- Enablement source: Clerk org private metadata `{"sdp": {"enableProductionProject": true}}`, synced into `organizations.settings` by the existing `organization.created/updated` webhook — same pipeline as `tier`/`providerOverrides`. Removing the key strips the setting on the next sync (fail-closed `=== true` parse).
- The switcher's hardcoded production lock (disabled row + LockIcon + "mainnet soon" tooltip) is deleted; any project that reaches the client is selectable.
- Enablement flips are logged (org id + new value) in the webhook sync.
- A later disable self-heals a stale selection: the hidden project drops out of the list and the existing cookie-repair path reselects default-sandbox.

**before** — production always visible, never selectable:

1. `GET /v1/projects` returns sandbox + production for every org
2. switcher renders the production row disabled with a lock tooltip
3. no path to mainnet for anyone

**after** — org-gated visibility:

1. Clerk webhook syncs `sdp.enableProductionProject` → `organizations.settings`
2. `listProjects` filters `environment === "production"` rows unless the org setting is `true` — both auth branches, so a production-scoped API key on a non-enabled org gets an empty list
3. enabled orgs see two selectable projects; everyone else sees sandbox only

**API before/after** — `GET /v1/projects`, org not enabled:

before (reconstructed from the old handler): `projects: [default-sandbox, default-production]`

after (live, dashboard session):
```json
{"data":{"projects":[{"name":"Default Sandbox Project","environment":"sandbox", "...":"..."}]}}
```
Org enabled (live, local): both projects returned. API-key auth (live): 200 with the key's own sandbox project; unauthenticated: 401.

**Self-hosted follows the same strategy** (review feedback):

- Deployment mode no longer determines network access: `SDP_DEPLOYMENT_MODE=self_hosted` previously implied blanket entitlements and skipped the Clerk sync entirely, so a self-hosted deployment was its own authority on what it could reach. That is removed — production/mainnet enablement and provider entitlements are decided by org metadata on every deployment mode.

- Deleted every self-hosted entitlement special case: the metadata-parse bypass and skipped webhook sync in provisioning, the blanket-entitlement overlays in `getProviderAvailability` and the RPC relay, and the self-hosted availability message. Capability config (local signer existence, cipher routing, BYOK flags) is untouched.
- Behavior change for self-hosted deployments: tier/`providerOverrides`/`enableProductionProject` now come from Clerk metadata like hosted; manual providers (compliance, earn) default OFF until overridden. The relay also stops re-validating `SDP_DEPLOYMENT_MODE` (its only entitlement consumer is gone); the API still fail-fast validates it at startup.

**Verification**

- `vitest projects.test.ts + provider-availability.service.test.ts` — 52/52 passed
- typecheck `@sdp/types`, `@sdp/api`, `sdp-web` — clean; Biome — clean
- local app: switcher exercised in a real browser — production row absent for a non-enabled org, sandbox marked current, no console errors; enabled case confirmed manually (two items, production selectable)
- not run locally: full `pnpm check` / `pnpm test:integration` (CI covers them)

**Screenshots**

**Self-hosted migration note**: an existing self-hosted deployment upgrading past this change must set org metadata overrides for providers it used under the old blanket entitlement (`sdp.providerOverrides.custody.local: true`, plus any compliance/earn providers) — until set, those flows report not entitled. This is the intended cost of metadata-only entitlement.

Known gaps
- Enforcement is the list read only: `GET /v1/projects/:projectId` still returns a production project by id and `sk_live` keys remain mintable regardless of the flag (unchanged behavior; follow-ups tracked on the mainnet epic).
- Not all providers support mainnet swapping yet; onchain actions become per-project cluster-aware in the stacked RPC PR — until it lands, onchain cluster selection is deployment-level.

Projects section of the open workspace switcher for a non-enabled org: sandbox only, no production row

CleanShot 2026-09-09 at 01 01 25@2x

### DIFF
--- a/apps/sdp-api/src/routes/custody-self-hosted.test.ts
+++ b/apps/sdp-api/src/routes/custody-self-hosted.test.ts
@@ -148,6 +148,10 @@ describe("Custody routes — self-hosted deployment mode", () => {
 
   it("GET /v1/wallets/switch-options returns only the configured local provider", async () => {
     await seedAuth("individual");
+    await getDb(env)
+      .prepare("UPDATE organizations SET settings = ? WHERE id = ?")
+      .bind(JSON.stringify({ providerOverrides: { custody: { local: true } } }), TEST_ORG.id)
+      .run();
 
     const res = await app.request(
       "/v1/wallets/switch-options",
--- a/apps/sdp-api/src/routes/projects.test.ts
+++ b/apps/sdp-api/src/routes/projects.test.ts
@@ -183,6 +183,83 @@ describe("Projects Routes", () => {
       ]);
     });
 
+    it("excludes production projects when the org has no enableProductionProject setting", async () => {
+      await getDb(env)
+        .prepare("UPDATE organizations SET settings = NULL WHERE id = ?")
+        .bind(TEST_ORG.id)
+        .run();
+      await seedProject("prj_production_disabled", "Production", "production-disabled");
+      await getDb(env)
+        .prepare(
+          `UPDATE projects SET environment = 'production' WHERE id = 'prj_production_disabled'`
+        )
+        .run();
+
+      const res = await app.request(
+        "/v1/projects",
+        {
+          headers: { Cookie: `sdp_session=${TEST_SESSION_ID}` },
+        },
+        env
+      );
+
+      expect(res.status).toBe(200);
+      const body = await res.json();
+      const ids = body.data.projects.map((project: { id: string }) => project.id);
+      expect(ids).toContain(TEST_PROJECT.id);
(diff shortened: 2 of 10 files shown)


multipletwigs:
@greptileai re-review — the settings sync now runs in a transaction with SELECT ... FOR UPDATE, and the enablement log was already moved after the write in 40119bf9b.

eldarik (approved):

PR #1709, opened 2026-09-08
fix(wallets): route the signer check through the guarded RPC egress

G1de0n:
- `POST /custody/signer-check` resolved the project's RPC target and dialled it with a raw client, so a project whose `settings.rpcEndpoint` pointed at `169.254.169.254` or an internal host had that request made on its behalf, with the error text and response timing as the oracle. The endpoint is only checked for being a well-formed URL when it is written.
- The call now builds its client from the same guarded transport the RPC relay uses, so a customer-supplied target is resolved and checked before connect, on the initial request and on every redirect hop.
- No contract change: request and response schemas, metering, and auth are untouched. Platform providers, local validators, and Surfpool are not customer-supplied, so they keep the ordinary fetch and behave exactly as before.
- Nothing else about the client changes: `createRpc` with an explicit `rpcUrl` already took the single-transport branch with no provider failover, and `createRpcFromTransport` applies the same default request timeout.
- A call-site test asserts the invariant that no runtime test can see — a *new* file that resolves an RPC target and then builds a raw client. It resolves the `createRpc` binding from the import, so an alias or a namespace import is caught too.

**before** — resolved target dialled directly:

1. Handler resolves the project's RPC target (platform provider or the project's own `custom` endpoint).
2. `createRpc(env, { rpcUrl: target.endpoint, headers })` builds a plain HTTP transport over that URL.
3. Blockhash and simulation requests go to whatever that hostname resolves to — link-local, loopback, or RFC1918 included.

**after** — resolved target dialled through the guard:

1. Handler resolves the target as before.
2. `createRpcTransportForTarget(target)` classifies it: `connectionId` present or `providerId === "custom"` → customer-supplied.
3. Customer-supplied targets go through `guardedFetch` — the destination address is checked before connect, redirects are bounded and re-checked per hop, and a blocked destination fails with `EgressBlockedError` instead of a connection.
4. Platform targets take the unguarded path unchanged.

**Verification**

- `vitest run src/services/rpc-egress.test.ts src/services/rpc-byok-parity.test.ts src/routes/custody` — 11 files, 144 tests passed.
- Reverting only the handler change makes the new call-site test fail with `[ "routes/custody/handlers/signer-check.ts" ]`; re-introducing the sink under an aliased import (`createRpc as rawCreateRpc`) fails it too.
- `tsc --noEmit`, `biome check` — clean.
- Local app not exercised: this changes no route contract or UI, and the guarded destination behaviour is covered by the egress suite against a controlled resolver rather than a live endpoint.

Known gaps
- The call-site test is a regression guard over import bindings, not a type-level invariant; a helper in another module that wraps the raw client would still pass it.

### DIFF
--- a/apps/sdp-api/src/routes/custody/handlers/signer-check.ts
+++ b/apps/sdp-api/src/routes/custody/handlers/signer-check.ts
@@ -93,14 +93,9 @@ export const signerCheck = async (c: ValidatedBodyContext<typeof signerCheckSche
       }),
     ]);
 
-    // Through the guarded transport, not a raw client: `resolveRpcTarget`
-    // returns the project's own `settings.rpcEndpoint` for the `custom`
-    // provider, and that value is only checked for being a URL when it is
-    // written. Dialling it directly made this route an API-key reachable SSRF
-    // into the metadata server and the private network, with error text and
-    // timing as the oracle — the same sink the relay already guards
-    // (HOO-1560). Platform targets keep the ordinary fetch, so local and
-    // Surfpool endpoints are unaffected.
+    // The `custom` provider endpoint is project-supplied and only URL-checked
+    // on write, so it goes through the guarded transport like the relay's
+    // (HOO-1560). Platform targets keep the ordinary fetch.
     const rpc = createRpcFromTransport(createRpcTransportForTarget(rpcTarget));
 
     const { blockhash, lastValidBlockHeight } = await getRecentBlockhash(rpc, "confirmed");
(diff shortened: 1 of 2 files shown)


resourcefulmind (approved):
LGTM. I traced what the resolver can hand back and the custom endpoint and tenant connection cases are the only ones a customer controls, and both now go through the guard. Nothing else in the handler talks to the target directly.

Small thing, not for this PR: HTTP failures from the guarded transport come out as a plain Error now instead of a SolanaError with the status code. Doesn't matter for the signer check, just worth knowing if someone reuses it somewhere that checks for a 401.

PR #1718, opened 2026-09-09
feat(api): project-based cluster RPC resolution for onchain money paths

multipletwigs:
- Every onchain path derives its Solana cluster from the request's **project environment** (sandbox → devnet, production → mainnet-beta), served by one deployment. Rebuilt from the earlier revision of this PR: instead of threading a `cluster` argument through ~50 files, the project middleware swaps `c.env` for a **cluster-scoped copy** — every cluster reader already takes `env` (`createRpc`, `getSolanaConfig`, well-known mint lookups, Kora gates, sponsorship network, custody provisioning), so all ~60 call sites flip with no edits.
- `scopeEnvToCluster(env, cluster)` (`lib/cluster-env.ts`) sets `SOLANA_NETWORK`, picks the per-cluster RPC URL (`SOLANA_MAINNET_RPC_URL` / `SOLANA_DEVNET_RPC_URL`; else the deployment's provider fan only for its own cluster; else no endpoint), and keeps `KORA_RPC_URL` only on devnet (`KORA_FEE_SPONSORSHIP_CLUSTERS`).
- Fail-closed, lazily: a production project on a deployment without `SOLANA_MAINNET_RPC_URL` still serves offchain routes; the first onchain call answers **503 `RPC_NOT_CONFIGURED`** (new `@sdp/rpc` error code).
- Genesis-hash proof moves into `@sdp/rpc`: a **mainnet-only transport wrapper** inside `createRpc` (30 s memo, coalesced, no retry/failover on mismatch) covers every client; an explicit **any-cluster** `assertClusterEndpoint(env, rpcUrl, rpc)` covers Earn, whose provider program ids resolve on either chain silently. Earn's private resolver/proof is deleted; all Earn clients go through `createProvenClusterRpc` / `provenVaultRpc`, and the provider-SDK URL handoff (Kamino/Veda/Jupiter build their own transports) through `resolveProvenClusterRpcUrl`.
- Non-request entry points scope per project row: public Solana Pay route (null-project requests now **400** instead of reporting `devnet` to a paying wallet), ramp webhooks, pending-transfer + recurring-collection jobs (`createProjectEnvResolver`), and the DvP reconciler, which now sweeps **per environment** because block heights are per cluster (claim release/list queries join through `projects`).
- Removed the hardcoded `DEFAULT_KORA_URLS` fallback in `@sdp/payments`: a Kora-less env fails closed (`PROVIDER_NOT_AVAILABLE`) instead of routing fees to an unconfigured public endpoint.
- Issuance supply refresh no longer rewraps a missing-RPC config error as a 502 RPC failure.

**before** — deployment-determined cluster:

1. any request → `createRpc(c.env)` → whatever `SOLANA_NETWORK` / `SOLANA_RPC_URL` the process was booted with
2. well-known mints resolved against that same process cluster
3. a production project silently built, read and reconciled against devnet — wrong data, not an error

**after** — project-determined, endpoint-proven:

1. `projectContextMiddleware` → `c.env = scopeEnvToCluster(c.env, CLUSTER_BY_SDP_ENVIRONMENT[environment])`
2. downstream `createRpc(env)` / `getSolanaConfig(env)` / `wellKnownMint(symbol, getSolanaConfig(env).network)` are cluster-correct by construction; on mainnet the first RPC send proves `getGenesisHash` against the pinned constant, mismatch → refusal naming observed vs expected
3. no endpoint for the cluster → 503 `RPC_NOT_CONFIGURED` at the first onchain call; Kora absent off devnet → wallet path / `PROVIDER_NOT_AVAILABLE`, never a default URL

## How the scoping works

```ts
scoped = { ...env, SOLANA_NETWORK: cluster }
override = cluster === "devnet" ? env.SOLANA_DEVNET_RPC_URL : env.SOLANA_MAINNET_RPC_URL
if (override)                              scoped.SOLANA_RPC_URL = override; drop managed-provider keys
else if (cluster !== defaultCluster(env))  drop SOLANA_RPC_URL + managed-provider keys   // fails closed on use
if (!KORA_FEE_SPONSORSHIP_CLUSTERS.includes(cluster)) drop KORA_RPC_URL
```

**API before/after** (from the route tests; production API key, deployment without `SOLANA_MAINNET_RPC_URL`)

`POST /v1/payments/transfers` `{ token: "USDC", … }`

before → `200` built against devnet (devnet USDC mint persisted)
after → `503 { "error": { "code": "RPC_NOT_CONFIGURED", "message": "No Solana RPC endpoint is configured for mainnet-beta" } }`

With `SOLANA_MAINNET_RPC_URL` set: the persisted `token` is the **mainnet** USDC mint and fee payment receives an env without `KORA_RPC_URL`.

`GET /pay/:token` for a payment request with no project

before → `200 { …, "network": "devnet" }`
after → `400 { "error": { "code": "BAD_REQUEST", "message": "Payment request is not associated with a project" } }`

**Verification**

- `pnpm -C apps/sdp-api exec tsc --noEmit -p .` — clean; `@sdp/rpc`, `@sdp/payments`, `@sdp/types` typecheck clean
- `apps/sdp-api` vitest, full suite against Postgres — 391 files, 5643 passed, 0 failed, 14 skipped (a later run showed only the shared-Redis `kv-redis` / `sponsorship-budget-redis` cross-session flakes, files untouched here)
- `@sdp/rpc` node tests 33/33 (proof matrix: explicit URL, mismatch, rejected-probe eviction, TTL, coalescing, devnet transport skipped, no failover/retry on mismatch, any-cluster `assertClusterEndpoint`); `@sdp/payments` 67/67
- Biome clean on the changed set
- local app: not exercised — behaviour is covered by the new route tests above (production-project transfers/issuance 503, middleware env swap, per-project job scoping, DvP per-environment sweep, cross-environment claim isolation)

**Security notes**

- Error messages for a misconfigured cluster name the env var and the observed genesis hash; ops-useful, not secret, only on a misconfigured deployment.
- The transport-level proof is mainnet-only by design: CI and local surfpool run a **mainnet fork under `SOLANA_NETWORK=devnet`** (verified: surfpool answers `getGenesisHash` with the mainnet hash), so a devnet transport proof would refuse them. Earn keeps its devnet proof explicitly via `assertClusterEndpoint`; other devnet paths accept a misconfigured devnet URL as before this PR.

Known gaps
- Managed sponsorship has no production scope yet, so a production transfer resolves the mainnet mint and then fails closed at fee payment (`Sponsorship is disabled for this scope`). Real wallet-pays for payments on mainnet is a follow-up on the mainnet epic.
- RPC relay (`resolveRpcTarget`) / observed transfers / Helius DAS keep their own per-connection or per-process URL selection (unchanged).
- Ops prerequisite: set `SOLANA_MAINNET_RPC_URL` before enabling any production project.

### DIFF
--- a/apps/sdp-api/src/cron/earn-catalogue-sync.ts
+++ b/apps/sdp-api/src/cron/earn-catalogue-sync.ts
@@ -40,7 +40,12 @@ import type {
   EarnVaultProvider,
   ProviderStrategySnapshot,
 } from "@sdp/earn/types";
-import { CLUSTER_BY_SDP_ENVIRONMENT, type SdpEnvironment, type SolanaCluster } from "@sdp/types";
+import {
+  CLUSTER_BY_SDP_ENVIRONMENT,
+  SDP_ENVIRONMENTS,
+  type SdpEnvironment,
+  type SolanaCluster,
+} from "@sdp/types";
 import { createEarnRepository, type EarnRepository } from "@/db/repositories";
 import type { BackgroundRunner } from "@/runtime/background";
 import type { KVStore } from "@/runtime/kv";
@@ -94,7 +99,6 @@ export const EARN_CATALOGUE_SYNC_DEADLINE_SECONDS = 600;
 // devnet mints, production rows mainnet mints), so each provider syncs once per
 // environment — plus, in every non-production environment, the mirrored
 // production mainnet shelf (PRO-1742).
-const SYNCED_ENVIRONMENTS: readonly SdpEnvironment[] = ["sandbox", "production"];
 
 // Expected steady states, not incidents: stub integrations report
 // NOT_IMPLEMENTED and environments without credentials report
@@ -140,7 +144,7 @@ async function syncProviderCatalogue(
     });
   }
 
-  for (const environment of SYNCED_ENVIRONMENTS) {
+  for (const environment of SDP_ENVIRONMENTS) {
     if (environment === "production") {
       continue;
     }
--- a/apps/sdp-api/src/cron/earn-metrics-refresh.ts
+++ b/apps/sdp-api/src/cron/earn-metrics-refresh.ts
@@ -38,7 +38,7 @@ import type {
   EarnRuntimeContext,
   ProviderStrategyMetrics,
 } from "@sdp/earn/types";
-import type { SdpEnvironment } from "@sdp/types";
+import { SDP_ENVIRONMENTS } from "@sdp/types";
 import { createEarnRepository, type EarnRepository } from "@/db/repositories";
 import type { BackgroundRunner } from "@/runtime/background";
 import { getLogger } from "@/runtime/logger";
@@ -73,7 +73,6 @@ export const EARN_METRICS_REFRESH_CRON = "*/5 * * * *";
 // five-minute-fresh production twin, fine for a browse-only, fundable:false
 // review shelf. (A provider whose catalogue snapshots omit the optional
 // currentApy would mirror with no rate at all; none does today.)
-const REFRESHED_ENVIRONMENTS: readonly SdpEnvironment[] = ["sandbox", "production"];
 
 // Expected steady states, not incidents — same taxonomy as the catalogue sync.
 const SKIPPABLE_REFRESH_ERROR_CODES: ReadonlySet<string> = new Set([
@@ -130,7 +129,7 @@ export async function refreshEarnStrategyMetrics(env: Env): Promise<void> {
   const repo = createEarnRepository(env);
   const providerEnv = env;
 
-  for (const environment of REFRESHED_ENVIRONMENTS) {
+  for (const environment of SDP_ENVIRONMENTS) {
     for (const client of Object.values(EARN_PROVIDER_CLIENTS)) {
       // Capability, not a provider list: a provider joins this pass by
       // implementing `listStrategyMetrics`.
--- a/apps/sdp-api/src/db/repositories/dvp-leg-funding-claim.repository.test.ts
+++ b/apps/sdp-api/src/db/repositories/dvp-leg-funding-claim.repository.test.ts
@@ -225,12 +225,29 @@ describe("DvpLegFundingClaimRepository", () => {
    * are swept for exactly this reason and this table was not, at first.
    */
   describe("releasing claims that can no longer land", () => {
+    it("leaves a production trade's claim untouched during a sandbox sweep", async () => {
+      await getDb(env)
+        .prepare("UPDATE projects SET environment = 'production' WHERE id = ?")
+        .bind(`prj_${AGENT_ORG}`)
+        .run();
+      await runWithTenantDatabaseIdentity({ organizationId: PARTY_A_ORG }, () =>
+        repo.claim({ ...claimInput(PARTY_A_ORG, "a", "sig_mainnet"), expiryHeight: "500" })
+      );
+
+      expect(await repo.releaseExpired(900n, "sandbox")).toBe(0);
+      await expect(
+        runWithTenantDatabaseIdentity({ organizationId: PARTY_A_ORG }, () =>
+          repo.listForTrade(TRADE_ID)
+        )
+      ).resolves.toHaveLength(1);
+    });
+
     it("frees a leg whose claim is past its last-valid height", async () => {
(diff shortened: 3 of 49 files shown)


multipletwigs:
@greptile-apps there's too much to fix in one PR for mainnet - we will slowly fix every single module. this is tracked. Please review only the rpc swap if it makes sense.

resourcefulmind (commented):
The swap itself looks right. Can you check these three things before it can merge, on top of the rebase now that #1709 changed signer-check.
- The genesis probe hits the project's own endpoint with the plain client, so after the rebase it sits outside the SSRF guard #1709 added. Build it from the guarded transport.
- The non-default cluster resolves to an empty URL unless SOLANA_MAINNET_RPC_URL / SOLANA_DEVNET_RPC_URL is set, and neither is documented. Prod serves sandbox projects, so one side 503s on every money call and /ready won't notice. Are those set in Doppler?
- The six preflight failures are the fixture: it mocks the DB as `{}` and onchain.ts now loads the project through it. Mock getProject.
Also the body says Mosaic and Token-2022 are converted; they still use the process client. Fine to track separately, just say so.

eldarik:
Really like the shape of this — scoping `c.env` once in `projectContextMiddleware` instead of threading a `cluster` argument through ~50 files is the right call, and deleting `DEFAULT_KORA_URLS` is a genuine security fix riding along. A few things before I'd approve 🔴

**Risk — the genesis probe payload has no JSON-RPC `id`.** `observeGenesisHash` hand-builds `{ jsonrpc, method, params }`; every other request gets its `id` from `@solana/rpc-spec`'s `createRpcMessage`, and the raw transport only does `JSON.stringify(payload)`. Without `id` this is a **Notification** — jsonrpc-core (Solana's RPC) returns HTTP 200 with an empty body, `parseJsonWithBigInts("")` throws, and since the proof runs ahead of *every* mainnet request, every mainnet RPC call fails. `cluster-proof.test.ts` can't catch it: the fake `fetch` reads `payload.id` (`undefined`), echoes it, and returns a result anyway. One-line fix, but please verify against a real mainnet endpoint rather than the mock.

**Risk — Earn reconciler lost its per-environment isolation.** In `reconcileEnvironment`, `createProvenClusterRpc` is now outside the `try` that used to wrap `assertClusterEndpoint`. The comment above the caller loop still promises "one environment's RPC outage is counted and reported, never allowed to skip the other environment's batch" — a proof failure now throws past the loop, skips the other environment, and loses the `sdp_api_earn_vault_reconciliation_status_read_failed` event that EARN-006 alerts on.

**Risk — one null `project_id` kills the whole transfer batch.** `groupTransfersByProjectEnv` throws `internalError` for the entire batch; `payment_transfers.project_id` is nullable (hence `COALESCE(project_id, '')` in `0025`), and this PR's own test diff flips an assertion from `project_id: null`. Skip-and-log the row instead.

**Risk — `SOLANA_DEVNET_RPC_URL` is no longer an Earn-only knob.** `scopeEnvToCluster` drops the whole managed-provider fan when an override is set, so a deployment that has it configured loses Helius/Triton failover for *all* sandbox traffic. Worth checking the deployed env values before merge.

**P2s**
- The per-cluster override is no longer trimmed (old `resolveClusterRpcUrl` did, and its test went away with it).
- `reconcileDvpEnvironment`'s `createRpc(scoped)` is outside any `try` — one open production trade without `SOLANA_MAINNET_RPC_URL` throws the whole cron each tick.
- `GET /pay/:token` now 400s for null-project requests (nullable column, `pay.test.ts` untouched) — please add the case and confirm no live rows.
- `reconcile-sponsorship-budgets` still resolves the network from the *unscoped* env, so the `scopeEnvToCluster` there is a no-op that reads like a fix.
- `createProjectEnvResolver` has no test.

**P3s** — mainnet Earn does two genesis round-trips (the transport proof and `assertClusterEndpoint` use different memo keys); `KORA_FEE_SPONSORSHIP_CLUSTERS` duplicates `EARN_VAULT_SPONSORSHIP_CLUSTERS`; `lib/cluster-env.ts` now imports from `services/`; `await (await createProvenClusterRpc(…))` reads awkwardly; the `SDP_ENVIRONMENTS` swap in the Earn crons is unrelated scope and orphans the comments above it.

multipletwigs:
Closing this for now. Will come back with cleaner pr for each module

PR #1723, opened 2026-09-09
feat(payments): Coinbase requirements decide individuals only

resourcefulmind:
HOO-1590. Third of three slices moving Coinbase on-ramp orders to embedded mode. Stacked on #1711; retargets to `main` once that merges.

- Coinbase's requirements step now refuses business counterparties with a reason, instead of answering `ready` and letting Coinbase reject the buyer later.
- Individuals on the on-ramp stay `ready` with nothing to collect: the hosted flow (embedded orders, #1711) gathers everything itself.
- Off-ramp stays `unsupported`, as before.
- Pure decision file, `coinbase/counterparty.ts`; the client's `validateCounterparty` delegates to it like the BVNK, Lightspark and Mural providers. No HTTP, no DB, no route or schema change.
- The support matrix already declares Coinbase individuals-only, so the wizard greys it out for businesses today; this is the API-side backstop for callers who skip the wizard.

**before** — business counterparty, Coinbase on-ramp:

1. `GET /v1/counterparties/{id}/requirements?provider=coinbase&direction=onramp` answers `ready`.
2. Caller requests a quote; Coinbase's order proceeds with a `partnerUserRef` for a company.
3. Verification inside the hosted flow fails or verifies the wrong thing; the failure surfaces late and without a reason SDP can show.

**after**:

1. Same GET answers `unsupported` with "Coinbase Onramp supports individual counterparties only".
2. The wizard shows the reason; an API caller gets it before quoting.
3. Individuals are unaffected: `ready`, then the embedded order.

**API before/after** (live, local API on this branch, sandbox org, key redacted)

`GET /v1/counterparties/{id}/requirements?provider=coinbase&direction=onramp&assetRail=usdc.solana&fiatCurrency=USD&destinationCustodyWalletId=cwlt_…`

Business counterparty `cpty_6854ff05…`, after:
```json
{ "data": { "provider": "coinbase", "direction": "onramp", "status": "unsupported", "reason": "Coinbase Onramp supports individual counterparties only." } }
```
Before (reconstructed from the removed inline decision): `{ "provider": "coinbase", "direction": "onramp", "status": "ready" }`.

Individual counterparty `cpty_433aaf6a…`, before and after:
```json
{ "data": { "provider": "coinbase", "direction": "onramp", "status": "ready" } }
```

Off-ramp: the requirements route schema does not admit `coinbase` for `direction=offramp` (400, discriminator lists the off-ramp providers), so the decision's off-ramp arm is reachable only through the provider client, where the unit test covers it. Unauthenticated: 401.

**Verification**

- `node --import tsx --test src/ramps/providers/coinbase/counterparty.test.ts` in `packages/sdp-payments` — 3 passed (individual ready, business unsupported, off-ramp unsupported)
- `pnpm --filter @sdp/payments test` — 74 passed
- `pnpm --filter @sdp/payments --filter @sdp/api typecheck` — clean
- `biome check` on the changed files — clean
- local API (`pnpm dev`, this branch): requirements GET for a business counterparty → `unsupported` with the reason; for an individual → `ready`; unauthenticated → 401. Responses above.

**Known gaps**
- None beyond the deferred web validation ticket already under the Coinbase epic.

### DIFF
--- a/packages/sdp-payments/src/ramps/providers/coinbase/client.ts
+++ b/packages/sdp-payments/src/ramps/providers/coinbase/client.ts
@@ -7,7 +7,6 @@ import { z } from "zod";
 import { divideDecimalAmounts, sumDecimalAmounts } from "../../../decimal";
 import { badRequest, providerNotConfigured, providerUnavailable } from "../../../errors";
 import { providerFetchJson } from "../../fetch";
-import { readyCounterparty } from "../../requirements";
 import {
   isSolanaCryptoAsset,
   RAMP_RAIL_DUMPS,
@@ -27,6 +26,7 @@ import type {
   RampRuntimeContext,
   ValidateCounterpartyOptions,
 } from "../../types";
+import { coinbaseCounterpartyRequirements } from "./counterparty";
 
 // v1 API (Bearer JWT): buy options, buy quote — used for rail discovery + estimates.
 const CDP_V1_API_BASE_URL = "https://api.developer.coinbase.com";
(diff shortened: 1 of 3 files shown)


resourcefulmind:
Closing this one: it is stacked on #1711, which is itself superseded.

Coinbase's headless onramp verification API lets us run phone and email OTP server-side and pass the resulting verification IDs to create-order: https://docs.cdp.coinbase.com/onramp/headless-onramp/verification

That is precisely the capability we believed we lacked when we chose the embedded model in ADR-0002 on 2026-09-08. A verification is valid for 60 days, so it belongs to the counterparty rather than to a single quote.

The replacement collects buyer contact in the counterparty requirements step, runs the verification there, stores the verification IDs against the counterparty's customer_link, and leaves /quote unchanged. Agreed with Zach on 2026-09-22.

Closing as superseded rather than abandoned: the work was correct for the constraints we understood at the time.

PR #1725, opened 2026-09-09
test(web): MoneyGram widget receives no phone or address prefill

resourcefulmind:
The MoneyGram widget collects the customer's phone and address inside its own hosted flow, so the dashboard must never pass them through the SDK's transaction prefill. Nothing asserted that until now; this pins it.

**What changes**
- The two prefill builders move from `moneygram-ramp-widget.tsx` into `moneygram-prefill.ts` with a named `MoneygramTransactionPrefill` type; the widget imports them. No runtime change, and the component file keeps only component exports.
- New unit test pins the exact key set per direction and asserts no personal-data key (`phone`, `phoneNumber`, `email`, `address`, `firstName`, `lastName`, `name`, `dateOfBirth`) ever appears.

**before** — prefill contract was implicit:
1. Widget builds the transaction prefill from amount, asset and currency.
2. An edit adding `phoneNumber` to either builder would ship with no failing test.

**after** — prefill contract is pinned:
1. On-ramp prefill is exactly `type`, `amount`, `asset`.
2. Off-ramp prefill adds `destinationCurrency`, and `destinationCountry` only for USD (`USA`) and MXN (`MEX`); other currencies omit the key.
3. Any personal-data key added to either builder fails the unit test.

**Verification**
- `pnpm exec vitest run src/app/dashboard/payments/ramps/components/moneygram-prefill.unit.test.ts` (apps/sdp-web) — 5 passed
- `pnpm exec tsc --noEmit` (apps/sdp-web) — clean
- `pnpm exec biome check` on the two changed files — clean
- No UI change, so no screenshots.

### DIFF
--- a/apps/sdp-web/src/app/dashboard/payments/ramps/components/moneygram-prefill.ts
+++ b/apps/sdp-web/src/app/dashboard/payments/ramps/components/moneygram-prefill.ts
@@ -1,0 +1,44 @@
+import { toNumberAmount } from "@sdp/solana/amount";
+import type { RampFiatCurrency } from "@sdp/types/generated/ramp";
+import type { CryptoAssetSymbol } from "@sdp/types/payment-rails";
+
+/**
+ * Transaction prefill handed to MoneyGram's Ramps SDK. Amounts, asset and
+ * destination only: the widget collects the customer's phone, address and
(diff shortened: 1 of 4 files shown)


multipletwigs (approved):

multipletwigs:
thanks!

PR #1732, opened 2026-09-09
feat(wallets): use exact custody wallets for signer checks

pashpashkin:
## What changed

- Resolve signer-check’s existing `walletId` to one exact custody wallet record, revalidate selected API-key permissions, and run runtime admission before signer, Kora, or RPC work.
- Sign through that same record. Preserve retained Connection ambiguity checks so inactive duplicates cannot silently redirect the check to a Config wallet.
- Document additional runtime errors and refresh generated playground examples.

## Why

This prepares signer-check for BYOK by keeping wallet authorization, credentials, and execution tied to the same custody record.

It reuses the shared execution safeguards without enabling BYOK or changing the public wallet selector.

## Impact

- Request and success-response formats remain unchanged: `walletId` is still a Provider wallet ID.
- Existing API-key selection when `walletId` is omitted is preserved.
- No migrations, backfill, new dependencies, feature-flag changes, or Dashboard component changes.
- Invalid selected-key permissions and runtime denials stop the check before external work.

## Pre/post release actions

- Pre: require green CI. No configuration or database changes are needed.
- Post: run **Wallet → Actions → Prove ownership** on an existing wallet; verify Connection-backed execution separately where enabled.
- Rollback: redeploy the previous API image. No database rollback is required; reverting the code also removes these additional checks.

## Result Validation

- After rebasing onto `main`: **95/95 targeted tests passed**, including wallet scope, inactive duplicates, Config compatibility, tenant boundaries, RPC egress/parity, and OpenAPI.
- API typecheck, Biome, module-boundary checks, and playground drift checks passed.
- Local Chrome smoke: **Prove ownership** on an existing Privy wallet showed **“Signer check passed”** and **“Nothing was sent on-chain.”**
- Full API suite before rebase: **5079 passed, 2 failed, 14 skipped**; coverage thresholds passed. Both failures were in unchanged quota tests, which passed **7/7** in isolation. The full suite was not rerun after rebase.

## Does it affect current flows & behavior and how

```text
Before:
walletId → Provider-ID signer lookup alongside Kora/RPC setup
         → sign and simulate

After:
walletId → exact wallet + fresh selected-key permissions
         → ambiguity check → runtime admission
         → same exact signer + Kora/RPC setup
         → sign and simulate
```

The server still creates the memo, verifies the signature, and only simulates the transaction. Nothing is broadcast and no sponsorship is spent. Existing quotas remain unchanged.

## Known gaps

- Connection/BYOK execution has automated coverage but was not exercised in the local live smoke.
- The public Provider-ID selector cannot explicitly choose between duplicate custody records.
- A standalone inactive Connection wallet can return `400 — Custody wallet not found` instead of its former runtime denial.
- Admission checks local eligibility, not provider health or guaranteed signing success.

## Notes

- No policy/Approval flow or durable wallet pin is added to this diagnostic endpoint.
- The guarded RPC transport from `main` is preserved.
- Generated playground changes also refresh pre-existing examples; Issuance runtime behavior is unchanged.

### DIFF
--- a/apps/sdp-api/src/openapi/paths/custody.ts
+++ b/apps/sdp-api/src/openapi/paths/custody.ts
@@ -312,7 +312,7 @@ export function registerCustodyPaths(registry: OpenAPIRegistry) {
         description: "Signer check verified in simulation",
         content: jsonContent(custodySignerCheckResponse),
       },
-      ...errorResponses(errorResponseSchema, [400, 401, 403, 429, 500, 502]),
+      ...errorResponses(errorResponseSchema, [400, 401, 403, 404, 409, 429, 500, 502]),
     },
   });
 
--- a/apps/sdp-api/src/openapi/schemas/custody.ts
+++ b/apps/sdp-api/src/openapi/schemas/custody.ts
@@ -30,7 +30,7 @@ export const switchSigningRequestSchema = withOpenApi(switchSigningSchemaBase, {
 
 export const signerCheckRequestSchema = withOpenApi(signerCheckSchemaBase, {
   description:
-    "Signer-check wallet selection. walletId is optional for an API key with one bound signing wallet and required for session-authenticated dashboard requests.",
+    "walletId is a Provider wallet ID. When omitted for an API key, its authenticated signing-wallet selection is used: the configured preference, otherwise the first resolved binding. Session-authenticated dashboard requests must provide walletId. Ambiguous Provider IDs are rejected; the check uses the resolved exact custody wallet record.",
   example: { walletId: "privy_wallet_123" },
 });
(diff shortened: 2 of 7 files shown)


pashpashkin:
@greptileai review pls

pashpashkin:
@greptileai review pls

resourcefulmind (approved):
Traced the whole path and it holds: lookup and signing land on the same row, the permission re-check only reads the authenticated key, and admission runs before anything touches Kora or RPC. 
Approving. 
Also, the specc now lists 404, but a missing wallet actually comes back as 400 since signing errors map to bad request.

PR #1738, opened 2026-09-09
feat(earn): PRO-1873 address-free positions summary via includeOwnerAddresses=false

keeganthomp:
`GET /v1/earn/external-wallet/positions/summary` hands every end-user wallet address in the project to any `earn:read` key through `totalsByStrategy[].ownerAddresses`, with no wallet-binding narrowing possible on that path. A leaked read-only key enumerates a partner's whole user base ([PRO-1873](https://linear.app/solana-fndn/issue/PRO-1873/threat-model-add-an-address-free-positions-summary-variant), threat model EARN-028). Analytics consumers need totals, not addresses.

**before**: the summary always carried the address list; the only way to get totals was to also hold the address book.

**after**: `?includeOwnerAddresses=false` omits `ownerAddresses` from every strategy total and keeps `walletCount`; a malformed value is a 400, not a guess.

## Reviewer notes

- **Default stays address-bearing, deliberately.** The dashboard drives its per-owner position reads off that list (`embedded-yield-dashboard.tsx`), so flipping the default would break it and every partner already consuming the field. The opt-out is where analytics keys are steered in the partner docs.
- **Omitted, never emptied.** An empty array would read as "no owners"; absence reads as "not requested". The wire type marks the field optional and the one dashboard call site is null-safe.
- Query flag rather than a sibling route: same auth, same scoping, same OpenAPI operation, one new documented parameter.
- Docs artifacts regenerated in the Docs Integrity order (`generate:api` then `generate:ai`).

**Verification**

- 2 new route tests: totals-only response contains neither owner address anywhere in the body, and the malformed flag is rejected
- earn external-wallet route suites and the dashboard unit suite green; `tsc` clean in sdp-api and sdp-web; biome clean

### DIFF
--- a/apps/sdp-api/src/openapi/paths/earn.ts
+++ b/apps/sdp-api/src/openapi/paths/earn.ts
@@ -4,6 +4,7 @@ import {
   earnExternalWalletEarningsQuerySchema,
   earnExternalWalletMovementParamsSchema,
   earnExternalWalletMovementsQuerySchema,
+  earnExternalWalletPositionSummaryQuerySchema,
   earnExternalWalletPositionsQuerySchema,
   earnStrategyIdParamsSchema,
   listEarnStrategiesQuerySchema,
@@ -199,7 +200,7 @@ function registerEarnExternalWalletPaths(
       "grouped by strategy and token. The service pages every stored claim before hydration. " +
       "A total is omitted when any contributing live value is unavailable, never reported as zero or partial.",
     security,
-    request: { headers: projectScopeHeaders },
+    request: { headers: projectScopeHeaders, query: earnExternalWalletPositionSummaryQuerySchema },
     responses: {
       200: {
         description: "Complete external-wallet position aggregate",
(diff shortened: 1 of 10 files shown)


resourcefulmind (approved):
Good fix, and I like that absence means "not requested" rather than an empty list. I checked the flag can't be coerced, the omission builds a fresh object, and no other earn route hands out end-user addresses to a read key. 
Let me quickly ask too...with the default unchanged, every existing read key keeps the exposure until its caller opts out. Is that what the threat model wanted, or is a default flip coming?

keeganthomp:
Not a flip, on purpose. The threat model asked for the variant plus telling partners `earn:read` is PII-bearing, and that's what this does. The residual is bounded: the list is project-scoped (cross-project reads 404, pinned in PRO-1860), so a leaked key only exposes that partner's own users to that partner.

Flipping the default breaks the dashboard (it drives per-owner reads off this list) and every partner integrated against it, so it didn't belong in a Medium ticket. I do think it should happen, at the next Earn API revision, with the dashboard moved to an explicit `includeOwnerAddresses=true`. Filed as [PRO-1908](https://linear.app/solana-fndn/issue/PRO-1908/threat-model-omit-owneraddresses-from-the-positions-summary-by-default) so it's a real decision and not an implication here.

PR #1747, opened 2026-09-10
fix(earn): PRO-1904 submitted-movement expiry corroboration, PRO-1872 public OpenAPI gate, PRO-1871 sponsor key guidance

keeganthomp:
Three small `[threat-model]` tickets from the Earn epic, bundled because none is large enough for its own PR. One commit each.

## What changed

| Ticket | Change |
| -- | -- |
| PRO-1904 | Migration 0092 adds `earn_movements.unknown_signature_observed_at`. The sweep parks a `submitted` movement on its first null-signature observation past the blockhash window and fails it only when a later tick sees it unknown again. `requested` rows keep the one-tick rule; `confirmed` rows are untouched. |
| PRO-1872 | `spec.test.ts` pins the exact public `/v1/earn` operation list (13 today) and asserts each shares its operationId with the internal twin. Growing the list requires named security sign-off (routes/earn/CLAUDE.md, AGENTS.md). |
| PRO-1871 | Embedded-yield guide gains "Protect the sponsor key" and a "Before production" checklist that links it and split-swap recovery. |

## Reviewer notes

- **PRO-1904 costs one extra tick.** A genuinely dead submitted movement now fails a minute later than before. That is the price of never turning a landed-but-forgotten transaction into a terminal false `failed`. Full rationale in the 0092 header and the CLAUDE.md "two unknown-signature observations" section.
- **The mark is sweep-only and write-once.** The interactive read-through passes no block height so it can neither park nor expire; COALESCE stops a burst of ticks counting as two observations; an unavailable height read is not an observation (ADR 0002 exit safety).
- **No PR template exists in this repo**, so the PRO-1872 "PR template note" lives in AGENTS.md and the earn CLAUDE.md instead of a new repo-wide template.
- **Parity is by construction.** Public and preview earn routes are the same Hono router under the same `/v1/*` tracing and rate-limit wildcards, so no gap tickets from PRO-1872.

## Verification

- Reconciliation job suite 31 pass (3 new), spec test 20 pass, withdrawals + repository + tenant-isolation + migration + authz-matrix + movements suites 161 pass.
- `tsc --noEmit` clean. The one biome warning and one docs link-check miss are pre-existing on main.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,6 +31,7 @@ Public docs and AI artifacts should mirror the supported public surface only.
 
 - Public API families: `health`, `api-keys`, `wallets`, `projects`, `issuance`, `payments`, `policies`, `compliance`, `earn`
 - Hidden/internal families stay out of public AI resources unless product policy changes: `rpc`, `admin`, `onboarding`, `auth`, `organizations`, `members`
+- Promoting a route into the public OpenAPI document is a security-relevant scope change, not a docs edit: the PR needs a named security sign-off. For `earn` the pinned list in `apps/sdp-api/src/openapi/spec.test.ts` enforces this (see `apps/sdp-api/src/routes/earn/CLAUDE.md`, "Public OpenAPI promotion").
 
 ## Preferred checks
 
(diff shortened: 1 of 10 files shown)


multipletwigs (approved):
wao nice security 👍 love it

PR #1750, opened 2026-09-10
fix(smoke): run the stage smoke browser on a same-site https origin

keeganthomp:
Fixes the `sdp: stage smoke canary failing` alert that has fired on every main merge since 2026-09-08. Follow-up to #1676 and #1685 (PRO-1856).

## The issue

The stage smoke has never actually passed. Before the `STAGE_SMOKE_READY` flip the build and Playwright steps were skipped, so the job went green on the readiness probe alone. Since the flip, every run dies in `auth.global.setup.ts` on the org-activation poll. The timeout bumps in #1731 and #1735 could not help because it is not a timing problem.

Stage uses the **production** Clerk instance (`clerk.platform.solana.com`), and the smoke runs the Next app on `http://localhost:3100`. That pairing cannot sign in, and the failure trace shows exactly why:

| Clerk cookie | Set as | Effect on a localhost page |
| -- | -- | -- |
| `__client` | `Domain=clerk.platform.solana.com; SameSite=Lax` | cross-site, never sent back; every Clerk call gets a fresh anonymous client, so the session touch behind `setActive` returns 401 `signed_out` |
| `__client_uat` | `Domain=solana.com; Secure` | cannot be stored at all, so `clerkMiddleware` bounces `/dashboard` to `/sign-in` |

The prod canary in sdp-infra uses the identical ticket flow and works only because it drives `https://platform.solana.com`, which is same-site with the Clerk host.

## Why this fixes it

The browser needs to sit on an https `*.solana.com` origin. This PR does that without any DNS or Vercel change:

- `scripts/playwright-same-site-proxy.mjs`: tiny TLS reverse proxy with a throwaway self-signed cert in front of `next start`, preserving `Host` and adding `x-forwarded-proto: https`.
- `playwright.config.ts`: when `PLAYWRIGHT_SAME_SITE_HOST` is set, adds the proxy as a webServer, sets `ignoreHTTPSErrors`, and launches Chromium with `--host-resolver-rules` mapping the host to `127.0.0.1`.
- `playwright/env.ts`: browser `baseURL` becomes `https://stage-smoke.solana.com:3443`; the Next server keeps listening on localhost. Ticket auth now fails loudly without a same-site host.
- `sdp-stage-smoke.yml`: sets the host. Ordinary dev-instance E2E runs are unchanged.

Once the page is same-site, both cookies flow, `setActive` succeeds, and the middleware sees a real session.

## Reviewer notes

- **Cannot be proven before merge.** `stg_ci` is only readable by the main-pinned OIDC identity, so the first real run is the post-merge stage smoke. If it still fails, the workflow is unchanged otherwise and reverting is a one-liner.
- **Clerk FAPI subdomain allowlist.** If one is configured in the Clerk dashboard, `stage-smoke.solana.com` needs adding. The trace shows Clerk echoing `localhost:3100` in CORS headers, so none appears to be set today.
- The cleaner long-term fix is a real stage web deployment on a `*.solana.com` host pointed at via `PLAYWRIGHT_BASE_URL`. That is sdp-infra/Vercel work and out of scope here.

## Verification

- `tsc --noEmit` and `biome check` clean.
- Real Chromium reached `https://stage-smoke.solana.com:3443` through the resolver rule and proxy against a dummy upstream, with `Host` and `x-forwarded-*` intact and response cookies passed through.


multipletwigs (approved):
lfg

keeganthomp:
This is will resolve itself once we have dns setup for staging/dev clerk instance.

PR #1777, opened 2026-09-11
fix(api): deliver workflow webhooks through the guarded egress transport

G1de0n:
Workflow webhook delivery validated the target hostname's resolution and then let the runtime resolve it again for the connect — a window the module's own comment acknowledged. Delivery now goes through `guardedFetch`: the addresses actually dialled are filtered at connect time, literal private hosts are refused, and the response is buffered under a 64 KiB cap, which also removes the manual body drains. The pre-flight lookup stays as the issuer's readable rejection, and the manual one-redirect loop still re-validates each hop. A target the transport refuses maps to the same permanent `BLOCKED_URL` outcome as a pre-flight rejection, so retries are not spent on a connection that fails the same way every time.

Part of HOO-976 hardening follow-ups.

## Tests

- Delivery goes through the guarded transport with a bounded read; the caller's timeout signal is preserved.
- A transport-refused connection fails permanently as `BLOCKED_URL`.
- Existing coverage holds: scheme/literal/hostname pre-flight rejections, redirect re-validation, signing, retry classification.
- Full sdp-api suite green.

### DIFF
--- a/apps/sdp-api/src/services/workflows/actions/webhook.test.ts
+++ b/apps/sdp-api/src/services/workflows/actions/webhook.test.ts
@@ -24,6 +24,33 @@ vi.mock("node:dns/promises", () => ({
       : [{ address: "93.184.216.34", family: 4 }],
 }));
 
+// The delivery transport is delegated to the test's fetch stub so delivery
+// behavior stays assertable without sockets; the transport's own rules
+// (connect-time address filtering, literal refusal) are asserted in
+// guarded-egress.test.ts. The recorder proves delivery goes through the
+// guard rather than a bare fetch.
+const guardedTransport = vi.hoisted(() => ({
+  calls: [] as Array<{ url: string; init: Record<string, unknown> }>,
+}));
+vi.mock("@/services/guarded-egress", async (importOriginal) => {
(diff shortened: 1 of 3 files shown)


multipletwigs (approved):
Wiring is right: `deliver()` goes through `guardedFetch` with the 64 KiB cap and the 10s signal, `EgressBlockedError` maps to a permanent `BLOCKED_URL` failure, and the connect-time DNS filter closes the rebinding window the old comment admitted was open. Approving with nits.

Nits on this diff:
- `webhook.test.ts:35` mocks `@/services/guarded-egress`, so the suite proves the call wiring but not SSRF rejection end to end. Acceptable since the guard has its own suite; an injectable transport would be the house-style shape.
- The new narration comments in `webhook.ts` restate the code — names/JSDoc instead.

Pre-existing, not blockers, but worth follow-ups since you're in this file:
- `webhook-url.ts:93` accepts `https://user:pass@host` and the credential lands in the plaintext `url` param.
- `stored.secret ?? resolveParam(action, "secret")` at `webhook.ts:111` is a legacy fallback.
- Retries carry no `Idempotency-Key` header; `executionId` is in the body so receivers can dedupe, but delivery should be documented as at-least-once.

PR #167, opened 2026-04-10
feat(api): sync Clerk organizations from webhooks

GuiBibeau:
## Summary

- Remove the legacy local `POST /v1/organizations` self-registration path and its registration-token/custody-onboarding plumbing.
- Make Clerk webhooks the simple source of truth for organization, user, and membership sync, including lifecycle cleanup for deleted orgs/users.
- Add webhook lifecycle coverage for Clerk org creation/update, membership creation/deletion, user update/deletion, and org deletion.
- Update setup docs to describe webhook-owned organization sync instead of manual dashboard linking.

## Validation

- `pnpm lint` (passes; existing unrelated integration-test suppression warnings remain)
- `pnpm typecheck`
- `pnpm --filter @sdp/api test:serial`
- `pnpm -C apps/sdp-api exec vitest run src/routes/organizations.test.ts src/routes/webhooks.test.ts src/openapi/spec.test.ts --reporter=verbose --no-file-parallelism --maxWorkers=1`
- `pnpm --filter sdp-docs check:links`
- `pnpm --filter sdp-docs build`

PR #1787, opened 2026-09-11
fix(redaction): scrub end-user owner addresses and raw provider bodies (PRO-1868)

keeganthomp:
Closes the EARN-025 residual: end-user owner addresses and raw provider payloads were the two things the telemetry denylist did not cover. Linear: [PRO-1868](https://linear.app/solana-fndn/issue/PRO-1868).

## What

| Change | Where |
| -- | -- |
| `ownerAddress` / `owner_address` is a PII key. Matched exactly, plus the `ownerAddress=` assignment and quoted-JSON string forms, so a logged URL or an echoed body is covered. | `packages/sdp-redaction` |
| Raw provider body containers (`providerBody`, `providerPayload`, `providerResponse`, `responseBody`, `rawBody`, `rawResponse`, `upstreamBody`) are denylisted whole. | `packages/sdp-redaction` |
| Vendor-call failure event pinned to `error_name` and `error_code` only, with a test that hands it an error carrying a body and an owner. | `apps/sdp-api/src/runtime/vendor-calls.test.ts` |
| Policy doc: new "End-user owner addresses" and "Provider payload bodies" sections, earn provider bodies added to the sinks table. | `docs/security/pii-scrubbing-policy.md` |

## Reviewer notes

- **Classification decision:** owner addresses are PII in telemetry. The threat model already classifies them as PII linkage (EARN-025, EARN-028) and any `earn:read` key can enumerate every owner in its project. This PR records that decision; the ticket asked for it.
- **The rule does not widen.** `walletAddress`, `destinationAddress`, `mintAddress`, `vaultAddress`, `feePayer` and bare `address` still survive, and `NEVER_REDACTED_KEYS` is unchanged. Tests pin both directions.
- **Two visible consequences.** The split-swap orphan events reach Loki with `owner_address` as `[REDACTED]`; triage runs on `advisory_id` (runbook updated in sdp-infra #167). Earn audit rows store the owner as `[REDACTED]` in `metadata`; `resourceId` is the movement id and the ledger row keeps the address, same arrangement the allowlist audit rows use.
- **No code change to the fetch layer was needed.** `providerFetchJson` already lifts only normalizer-named fields plus `provider` and `providerStatus` onto `SdpEarnError.details`; the new container keys are the backstop for a caller that attaches a body anyway.

## Verification

- `@sdp/redaction` 42/42, typecheck and biome clean.
- sdp-api: vendor-calls + logger 24/24; earn external-wallet, orphan detector and audit service suites 83/83 (owner addresses flow through the pino hook, so nothing asserting on the pre-hook payload changed).
- `apps/sdp-api` typecheck clean; zero em dashes in added lines.

### DIFF
--- a/apps/sdp-api/src/runtime/vendor-calls.test.ts
+++ b/apps/sdp-api/src/runtime/vendor-calls.test.ts
@@ -12,6 +12,51 @@ function failureEvents() {
   return logEvent.mock.calls.filter(([, payload]) => payload.event === "sdp_api_vendor_call");
 }
 
+describe("logVendorCallFailure", () => {
+  it("logs the error's name and code only, never a provider body or its details", async () => {
+    // EARN-025: a provider error routinely echoes the request, the owner and
+    // whatever the provider felt like saying. The failure event is the one log
+    // line every vendor call shares, so it carries facts about the FAILURE, not
+    // the payload that failed.
+    const failure = Object.assign(new Error("kamino request failed with status 400"), {
+      name: "SdpEarnError",
+      code: "BAD_REQUEST",
+      details: {
+        provider: "kamino",
+        providerStatus: 400,
+        providerBody: '{"owner":"7Yq3nRbFkMd2pXcLwT9vZs4HaJ1uE6gQoB8iVtNrK5mD","reason":"nope"}',
+        ownerAddress: "7Yq3nRbFkMd2pXcLwT9vZs4HaJ1uE6gQoB8iVtNrK5mD",
+      },
+    });
+
+    await expect(
+      withVendorCall("kamino", "quoteVaultDeposit", () => Promise.reject(failure))
+    ).rejects.toBe(failure);
+
+    const emitted = failureEvents();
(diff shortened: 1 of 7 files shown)


resourcefulmind (commented):
The redaction change is right and the tests pin both directions, nice. One gap before we call PRO-1868 closed: the vault position hydration warn logs still emit the end user's address under owner and snapshotOwner, right next to the provider name, which is the exact linkage this PR describes. The scrubber only matches ownerAddress, so those lines ship as-is. Pre-existing, not introduced here, but same ticket. Rename or drop the field on the external-wallet path (a blanket owner rule would blind the custody caller) and I'll approve.

keeganthomp:
Fixed this in 8ce6662f. External-wallet hydration warnings now drop both the expected and snapshot owner addresses before logging, while custody keeps the useful owner context. Added coverage for the mismatch and failure paths too; 70 focused tests and the API typecheck are passing.

resourcefulmind (approved):
Approving!

PR #1792, opened 2026-09-11
feat(private-channels): authorize the exact custody wallet before signing and replay

pashpashkin:
## What changed

**Authorization order for value movement** — `POST /v1/private-channels/{deposits, withdrawals, channels/:id/transfers, wallets/:id/verify}`
- The exact custody wallet record is resolved (409 when two records match the selector), its API-key wallet bindings are re-read from the DB, it is admitted through the custody runtime (`admitRuntimeExecution`), and the signer is derived for that record — all **before** any SPC session, balance read or idempotency reservation. New seam: `services/private-channels/wallet-access.ts`.
- Coarse access is fresh too: key status/expiry and role-or-explicit permissions are read from `api_keys` on every one of these POSTs, not from the KV snapshot. `permissions = NULL` inherits the *current* role; `[]` grants nothing.

**Replay and recovery** — `routes/private-channels/replay.ts`
- A replay by `Idempotency-Key` requires current `payments:write` and authorizes the *recorded* source: walletId, public key and instance must still resolve for the caller. The recorded result is returned with no new signer and no SPC session.
- Abandoned `pending` rows (>10 min) are recovered under the same rule by the existing resolvers: failed if never signed, otherwise settled from the recorded signature through a gateway read. Still no signer.
- A counterparty `recipient` / `destination` address that matches custody records with different keys → 409 instead of first match.

**Transfers unblocked**
- Migration `0094`: a `pending` transfer row may carry `signature` (signature-before-send). The 0040 CHECK rejected that write.

**Tests**
- Route suites run the real app and authorization seam on Postgres/Redis testcontainers with the Privy HTTP API stubbed; execution below the service boundary stays mocked there.
- New `value-movement.node.test.ts` executes deposit, withdrawal and transfer against real Postgres: the `sendTransaction` stub asserts the signature is already visible in history before the bytes leave; one signer under concurrent identical keys; an unknown send outcome stays `submitted`.

## Why

- **Wrong order.** Deposits and withdrawals derived the signer inside the broadcast step, after the reservation and the gateway session. A provider failure there — or a Connection wallet refused by the BYOK flag — was caught and stored as a `failed` row, answered 200 with `status: failed`, and consumed the caller's idempotency key.
- **Inexact source.** The source was the first match by walletId or address in a provider wallet list, signed through `createOrgSigner(walletId)` rather than the exact custody record. Retained or duplicate records were invisible.
- **Stale authority.** Permissions and bindings came from the cached key (up to 1 h): a key revoked or narrowed in the DB kept moving value until the cache expired or was swept.
- **Replay trusted the retry.** A replay was authorized against whatever record the retry's walletId resolves to *now* — a replacement record with the same walletId passed — then returned the recorded operation.
- **Transfers were broken.** Transfer path records the signature on the still-pending row; the 0040 CHECK rejects that write, so every member transfer failed before send. Service tests mock the repository; the new real-Postgres execution test is what caught it.

## Impact

- **API consumers and dashboard sessions:** same routes and bodies. Failures move earlier and get stricter (table below). Keys with `permissions = NULL` follow the DB role, not the cached one.
- **BYOK:** a Connection wallet enrolled in a Private Channels principal can deposit, withdraw, transfer and verify when `PRIVY_BYOK_ENABLED=true`. With the flag off it is refused with 403 `runtime_execution_paused` before any SPC session. Config wallets are unaffected by the flag.
- **Cost:** a few extra small DB reads per money POST, and the Privy signer round-trip now precedes the balance check, so a request that then fails validation still costs one provider call.
- Feature stays behind `PRIVATE_CHANNELS_ENABLED`, devnet only.

## Pre/post release actions

1. **Migrate.** `0094_private_channel_transfer_signed_reservation.sql` — transactional constraint swap that validates existing rows (small table). The runner applies by filename set; verified it applies on a DB already at 0096.
2. **Audit stored permissions.** Rows with non-array `api_keys.permissions` now fail PC POSTs with 500 `Stored API key permissions are invalid`:
  ```sql
    SELECT id
    FROM api_keys
    WHERE permissions IS NOT NULL
      AND CASE
        WHEN NOT pg_input_is_valid(permissions, 'jsonb') THEN true
        WHEN jsonb_typeof(permissions::jsonb) <> 'array' THEN true
        ELSE EXISTS (
          SELECT 1
          FROM jsonb_array_elements(permissions::jsonb) AS item(value)
          WHERE jsonb_typeof(item.value) <> 'string'
            OR (item.value #>> '{}') <> ALL (ARRAY[
              'tokens:read', 'tokens:write', 'tokens:admin',
              'payments:read', 'payments:write', 'earn:read',
              'earn:write', 'counterparties:read', 'counterparties:write',
              'wallets:read', 'wallets:write', 'compliance:read',
              'compliance:write', 'webhooks:read', 'webhooks:write',
              'transactions:read', 'transactions:write', 'audit:read',
              'org:read', 'org:write', 'org:admin',
              'api-keys:read', 'api-keys:write', 'projects:read',
              'projects:write', 'projects:admin', 'project-members:read',
              'project-members:write', 'sessions:read', 'sessions:write',
              'custody:read', 'custody:write', 'custody:admin',
              '*'
            ]::text[])
        )
      END;
  ```
   Unknown permission names inside the array fail the same way.
3. **After release, watch** 409 `Custody wallet ownership is ambiguous` (duplicate `custody_wallets` for one selector → data fix) and 500 `Stored API key permissions are invalid`.
4. **Rollback** = revert the app. The relaxed CHECK accepts everything the old one did, so 0094 can stay.

## Result validation

| Check | Result |
|---|---|
| `pnpm --filter @sdp/api typecheck` | clean |
| `pnpm --filter @sdp/api lint` | clean for the diff; 1 pre-existing warning in untouched `api-key-scope.service.ts:630` |
| `pnpm --filter @sdp/api exec vitest run private-channels` | 23 files, 314 tests passed (the 6 changed files: 186). Needs Docker: Postgres 16 + Redis 7 testcontainers; SPC gateway and Privy HTTP stubbed |
| `vitest run src/security/value-moving-conformance.node.test.ts` | passed; the contract now points at the handler-level authorization seam |
| Schema before/after | `psql \d` on a throwaway `postgres:16-alpine`: main's migrations, then the branch's |
| Live local API against a real SPC gateway / devnet | **not run** (external services). Route tests exercise the real app end-to-end below the gateway boundary |

## Does it affect current flows & behavior and how

**before** — deposit / withdrawal on `main`:
1. `payments:write` from the cached key; source = first provider-list match; bindings from the cache
2. SPC session, balance read, reserve `pending` row
3. derive signer → sign → record signature → send. A signer failure here becomes a `failed` row under the caller's key.

**after** — every value-moving POST:

```mermaid
flowchart TD
  A[POST deposit / withdrawal / transfer] --> B[requireMovementWrite<br/>key status + role/permissions from DB]
  B --> C{row for Idempotency-Key?}
  C -- yes --> R1[authorize the recorded source<br/>exact record, fresh bindings]
  R1 --> R2[fingerprint match]
  R2 --> R3{abandoned pending?}
  R3 -- yes --> R4[fail if never signed,<br/>else settle from the recorded signature] --> Z[200 recorded result<br/>no signer]
  R3 -- no --> Z
  C -- no --> N1[resolve exact source wallet<br/>409 if ambiguous, fresh bindings]
  N1 --> N2[admitRuntimeExecution + signer for the record<br/>403 BYOK paused · 503 provider]
  N2 --> N3[SPC session, balance read]
  N3 --> N4[reserve pending row<br/>lost race → authorize the winner via onReplay]
  N4 --> N5[sign → record signature on the pending row → send]
```

**Status changes** (reconstructed from the route tests; live API not run):

| Situation | before | after |
|---|---|---|
| Key revoked / narrowed in DB, cache still valid | 200 | 403 |
| Two custody records match the source selector | silently picks one | 409 |
| Address selector collides with a retained record | silently picks the active one | 409 |
| Replay when the original record is inactive and a replacement shares its walletId | replacement authorized | 409 |
| Signer provider unavailable (deposit / withdrawal) | 200 with `status: failed`, key consumed | 503, nothing recorded |
| Connection source, `PRIVY_BYOK_ENABLED=false` | same `failed` row, after reservation | 403 before any session |
| Transfer: signer address ≠ verified source | 400 | 409 |
| Malformed `api_keys.permissions` | not validated | 500 |
| Any member transfer | `failed` before send (CHECK) | submitted / confirmed |

**Schema before/after** — `psql \d private_channel_transfers`, trimmed to the changed constraint:

```text
before  "private_channel_transfers_result_check" CHECK (status = 'pending' AND signature IS NULL AND failure_reason IS NULL OR …)
after   "private_channel_transfers_result_check" CHECK (status = 'pending' AND failure_reason IS NULL OR …)
```

## Known gaps

- No `policyGate` on Private Channels value movement and no private-channel family in `WALLET_OPERATION_FAMILIES` (pre-existing; product decision).
- The four signing POSTs are not behind `meteredQuota`; the Privy call now precedes the balance check (pre-existing; mirror `dvp-fund`).
- `api_keys.permissions` is parsed in three places with different tolerance (cache loader, approved-operation replay, this PR); consolidate into one strict parser.
- Fresh-auth reads repeat up to 3× per POST; thread the refreshed context (perf only).
- Deposits have no abandoned-reservation recovery (pre-existing; withdrawals and transfers do).
- Lost-race re-authorization has a route test for transfers only; deposits and withdrawals share the code path.

## Notes

- Three commits: the migration comes first so the feature commit runs against the relaxed CHECK. The `0094` prefix is shared with `0094_dvp_…`; the runner selects by filename, not by cursor.

## Beyond spec

- `POST /wallets/:walletId/verify` OpenAPI: `security` now lists `apiKeyAuth` (runtime already accepted keys through the router auth; the spec now matches) and adds 409.
- Transfer signer / verified-source mismatch: 400 → 409 (the check moved into `CustodyRuntimeTargets.assertSignerMatchesWallet`).
- Malformed `api_keys.permissions` → 500 `INTERNAL_ERROR` on PC POSTs instead of lenient parsing; failing loudly at the boundary.
- Migration 0094 relaxes `private_channel_transfers_result_check` — forced: transfers were broken
- Counterparty `recipient` / `destination` matching several custody records with different keys → 409 (was first match).

### DIFF
--- a/apps/sdp-api/src/db/migrations/postgres/0094_private_channel_transfer_signed_reservation.sql
+++ b/apps/sdp-api/src/db/migrations/postgres/0094_private_channel_transfer_signed_reservation.sql
@@ -1,0 +1,13 @@
+-- Commit 6f0c3e04e added signature-before-send but left the original CHECK intact.
+-- The transfer service records a signature before send while it still owns the
+-- pending reservation. A crash in that window must remain observable by replay
+-- recovery; the original constraint rejected this write and prevented any send.
+ALTER TABLE private_channel_transfers
+    DROP CONSTRAINT private_channel_transfers_result_check,
+    ADD CONSTRAINT private_channel_transfers_result_check
+        CHECK (
+            (status = 'pending' AND failure_reason IS NULL)
+            OR (status = 'submitted' AND signature IS NOT NULL AND failure_reason IS NULL)
+            OR (status = 'confirmed' AND signature IS NOT NULL AND failure_reason IS NULL)
+            OR (status = 'failed' AND failure_reason IS NOT NULL)
+        );
--- a/apps/sdp-api/src/openapi/paths/private-channels.ts
+++ b/apps/sdp-api/src/openapi/paths/private-channels.ts
@@ -269,7 +269,7 @@ export function registerPrivateChannelsPaths(registry: OpenAPIRegistry) {
     summary: "Create a deposit into the channel escrow",
     operationId: "createPrivateChannelDeposit",
     description:
-      "Builds, server-signs, and broadcasts an escrow deposit from a custody wallet to the instance chain (devnet), crediting `recipient` (defaults to the depositor) in the channel. Returns the deposit with its current status (submitted/confirmed, or failed). The credit (`credited`) is detected asynchronously via the gateway balance. Accepts a dashboard session or an API key. The source `walletId` must be a custody wallet enrolled under the project's Private Channels principal on this instance (and, for a wallet-scoped API key, one the key is bound to), and `recipient` must be an address verified on it. The `Idempotency-Key` header is REQUIRED: an identical retry returns the original deposit without a second escrow transfer, while reusing the key for a different request returns 409.",
+      "Builds, server-signs, and broadcasts an escrow deposit from a custody wallet to the instance chain (devnet), crediting `recipient` (defaults to the depositor) in the channel. Returns the deposit with its current status (submitted/confirmed, or failed). The credit (`credited`) is detected asynchronously via the gateway balance. Accepts a dashboard session or an API key. The source `walletId` must be a custody wallet enrolled under the project's Private Channels principal on this instance (and, for a wallet-scoped API key, one the key is bound to), and `recipient` must be an address verified on it. The `Idempotency-Key` header is REQUIRED: an identical retry returns the original deposit without a second escrow transfer, while reusing the key for a different request returns 409. New execution requires current payments:write, authorization for one unambiguous source custody wallet, and runtime admission before signing setup or an SPC session. Every POST, including a matching replay, requires current payments:write and authorization for the original source. A matching recorded result returns without a new signer or runtime admission; read-only access remains on the history GET endpoints. Separately authorized recovery of an abandoned operation may update its recorded state without a new custody signature.",
     security: [{ apiKeyAuth: [] }, { sessionCookie: [] }],
     request: {
       headers: projectScopeWithRequiredIdempotencyHeaders,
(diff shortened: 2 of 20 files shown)


pashpashkin:
@greptileai review pls

resourcefulmind (approved):
Went through all four signing routes and the replay path. Authorization and admission now sit ahead of every signer and SPC session, replay re-checks the current key against the recorded wallet, and the tests that changed all got stricter replacements. Approving.

One thing for the release step: the body says unknown permission names in a stored array also 500, but the audit query only catches non-array values. Either extend the query to check element names against the enum, or turn that case into a 403, so we're not finding it in prod logs.

pashpashkin:
Thank you for the review @resourcefulmind!
I've updated the audit query; dev, stage, and prod return 0 rows

PR #1796, opened 2026-09-13
feat(dvp): let a party reclaim its funded leg, and fund it again afterwards

tobySolutions:
PRO-1924. Draft until #1795 merges: it builds on that PR's refusal codes and leg footer. It gets rebased onto main and marked ready then, so it never needs reviewing as a stack.

A party that funded a leg had no way to get it back short of the settlement authority cancelling the whole trade. `POST /v1/dvp/trades/:tradeId/reclaim` sends `ReclaimDvp` for one side, signed by the custody wallet at that side's party address (the only signer the program accepts). Same authorization, metered quota and Kora sponsorship path as fund; the leg resolution is now one helper both use.

Before / after:
1. Reclaim a funded leg. Before: not possible. After: escrow drains to the party's own ATA for that mint, created first if missing; the trade stays open.
2. Fund the same leg again. Before (once reclaim existed): 409 "already being funded by another request" forever, because the landed receipt kept `(trade_id, side)` taken. After: the receipt is cleared once the reclaim is on the wire (`deleteReceipt`, guarded on `funding_tx IS NOT NULL` so an in-flight funding is never touched).
3. Trade page: a quiet Reclaim on the footer of a leg the caller holds while its escrow holds something, expired trades included.

Refusals, each with `details.reason`: closed trade, empty escrow, a funding of that leg still in flight, tampered escrow, trade gone, and transfer-hook mints.

Interleavings:

| case | outcome | cost |
| --- | --- | --- |
| funding of the same leg in flight | 409, nothing sponsored | none |
| two reclaims at once | both send; the second drains 0 (program no-op) | one sponsored fee |
| reclaim dropped after broadcast | receipt already cleared; fund still refuses a leg holding its amount (live read), retry drains the rest | fee |
| crash between broadcast and receipt delete | receipt stays; retry reclaims 0 → 409 empty escrow, receipt stays | leg can't be re-funded until a reclaim with a balance clears it |
| preflight rejection | nothing changed, receipt kept | none |

Proven on the deployed program (surfpool forking devnet, `dvp34bdbc…`), built with the same `buildReclaimInstructions` the service uses:
- T22 leg A funded 1000 by its party: reclaim `5kS82XGC…`, escrow 1000→0, party 0→1000
- legacy SPL leg B paid 700 by a third party, party had no ATA: reclaim `3JBHzK7X…` created it, escrow 700→0, party →700
- trade account still open, then re-fund of leg A 400 landed (`63Ra2e9S…`)

Tests: reclaim service 17 (wiring per `reclaim_dvp.rs` layout, legacy token program on leg B, receipt cleared only after broadcast, every refusal), repository 15 (receipt cleared, in-flight lock never), route 43 (custody refusals, body validation) plus the wallet-scoped route inventory, web workspace + hook 63.

Not verified: the Kora-sponsored path end to end (the proof above used a plain fee payer), and staging.

Known gaps:
- Transfer-hook mints are refused, not supported. `settle-instructions.ts:60` says V1 refuses hook mints before settle, but `BLOCKED_MINT_EXTENSIONS` allows TransferHook, so settle and cancel on such a trade would also fail today; not fixed here.
- Reclaim, like every DvP action since #1774, is not wallet-policy gated.
- The reclaim signature is not recorded against the leg; the escrow ledger (PRO-1941) is where that belongs.
- The crash row above can leave a stale receipt; PRO-1891 covers claim/receipt atomicity.


tobySolutions:
Folded into #1795 so the DvP trade page work reviews as one PR. Closing this one.

PR #1800, opened 2026-09-14
ci: scope test detection to any workflow change

G1de0n:
## What

The \`Unit Test Scope\` detect filters in \`ci.yml\` matched only \`.github/workflows/ci.yml\`; they now match any file under \`.github/workflows/\`.

## Why

A pull request that touches only workflow files (for example #1776) left every scope output false, so the \`browser_e2e\` job was skipped before its matrix expanded. The required status checks \`Dashboard E2E (Surfpool Local)\` and \`Dashboard E2E (Transactions / Surfpool Local)\` were then never reported on the head commit — the run shows a single placeholder check literally named \`matrix.name\` — and the pull request stayed blocked indefinitely with every visible check green.

## After

Any change under \`.github/workflows/\` marks all four scopes (\`api\`, \`rest\`, \`e2e\`, \`docker\`) as affected, so the full test matrix runs and every required check reports. Workflow-only pull requests are rare, so the extra CI time is an acceptable trade for never deadlocking on unreported required checks.

## Verification

\`actionlint\` on the changed file reports only the pre-existing SC2086 note on an untouched script block; the diff is four regex lines in the \`detect\` calls.


G1de0n:
Folded into #1776.

PR #1802, opened 2026-09-14
fix(db): lift the wallet-operations family check around the project consolidation deletes

G1de0n:
**Summary**

- Stage deploys have been red since 2026-09-11 16:46 UTC: migration `0096_projects_active_environment_unique.sql` (#1770) aborts with `new row for relation "wallet_operations" violates check constraint "wallet_operations_family_check"`, so every migration after it (0096-helius, 0097, 0098) never applies (executions `sdp-stage-api-public-migrate-djgxf` … `-qjww9`).
- Root cause: `wallet_operations.project_id` and `.api_key_id` are `ON DELETE SET NULL`, and 0096 deletes non-default projects and their api_keys. `wallet_operations_family_check` was rebuilt `NOT VALID` in 0058/0060 — it skips existing rows but re-fires on any UPDATE, and an FK-driven `SET NULL` is an UPDATE. Stage carries historical `wallet_operations` rows with pre-0058 (retired) `operation_family` values attached to consolidated projects, so the cascade trips the check and rolls the whole file back.
- Fix: drop the constraint before the two DELETEs and restore the exact 0060 definition `NOT VALID` afterwards. The runner executes the file in one transaction, so concurrent writers queue against the table lock and the check is back before COMMIT — no window for bad rows.
- 0096 is unapplied on stage and prod, so editing it in place is safe; dev already recorded the filename in `schema_migrations` and will not re-run it.
- Dev passed originally only because it has no retired-family rows attached to doomed projects; prod data should be assumed to have them.

**before** — 0096 on a database with retired-family history:

1. `DELETE FROM api_keys` (doomed projects) → FK sets `wallet_operations.api_key_id = NULL` → check re-fires → abort (or, past that, the same via `DELETE FROM projects` and `project_id`)
2. Transaction rolls back; 0096 stays unapplied
3. Every subsequent migration in the file order never runs; each deploy repeats the failure

**after** — same data:

1. `DROP CONSTRAINT wallet_operations_family_check`
2. Both DELETEs run; FK cascades set `project_id` / `api_key_id` to NULL freely; retired rows keep their historical `operation_family`
3. Constraint re-added `NOT VALID` with 0060's family list; new inserts with retired families are still rejected; COMMIT

**Schema before/after**

Constraint definition is unchanged before vs after the migration — this PR changes only *when* it exists inside 0096's transaction:

```
conname                        | convalidated
wallet_operations_family_check | f
CHECK (operation_family IN ('transfer','payment','ramp','issuance','program')) NOT VALID
```

**Verification**

- Local reproduction on a scratch database: applied migrations through 0095, seeded a doomed archived project + api_key + `wallet_operations` row with retired family `signer_check` (inserted with the check lifted, mirroring pre-0058 history).
- `node scripts/migrate-postgres.mjs` with the **original** 0096 — fails with exactly the stage error, on the `api_keys` DELETE.
- Same command with the **fixed** 0096 (+ 0096-helius + 0097 + 0098) — `Postgres migrations are up to date.`; the historical row survives with `project_id`/`api_key_id` NULL and `operation_family` intact; `pg_constraint` shows the check restored (`convalidated = f`); a fresh insert with a retired family is still rejected.


resourcefulmind (approved):
Approving

PR #1807, opened 2026-09-14
feat(earn): open API to keyless integrations

keeganthomp:
## Summary

- Opens six Earn catalogue, quote, and unsigned build operations to keyless callers, each declared once.
- Keeps supplied credentials strict, authenticated permissions intact, and anonymous traffic outside tenant state and durable writes.
- Adds abuse controls, schemas, tests, and aligned public and internal documentation.

## Validation

- 214 focused API tests, 208 Earn package tests, 16 Ondo tests, and 49 Treasury UI tests
- Full workspace typecheck, module boundaries, API playground drift, generated-artifact drift, docs links, and docs build

## Security and rollout

- Threat-model gate: [PRO-1872 / EARN-027](https://linear.app/solana-fndn/issue/PRO-1872/threat-model-make-openapi-promotion-of-preview-routes-a-security)
- Optional-auth OpenAPI follow-up: [PRO-1948](https://linear.app/solana-fndn/issue/PRO-1948/publish-the-two-tier-earn-api-in-openapi-and-the-embedded-yield-guide)
- Named security reviewer: pending assignment
- Observability: [PRO-1947](https://linear.app/solana-fndn/issue/PRO-1947/rate-limit-and-meter-anonymous-earn-traffic) owns the Grafana anonymous-volume panel and 429 alert

The public OpenAPI remains API-key-only until the named security sign-off is recorded. Before preview smoke testing, non-production Cloud Run must set `SDP_ENVIRONMENT=sandbox`; the API now validates and logs the resolved anonymous Earn environment at startup.

Production Cloud Run ingress being load-balancer-only still needs confirmation from Nikhil because client-IP resolution relies on that trust boundary.

The ticket assumes sponsored keyed builds and movement creation during build. Current main does neither, so this PR preserves caller-paid builds with `sponsored: false` and creates movements only on authenticated submission.

Linear: https://linear.app/solana-fndn/issue/PRO-1943/p0-open-the-earn-api-to-integrators-with-or-without-an-api-key


apps/sdp-api/src/routes/earn/index.ts
```diff
@@ -400,4 +532,7 @@ earn.get(
   getEarnProgramWithdrawal
 );
 
-export default earn;
+earnRoutes.route("/", optionalAuthEarn);
```
keeganthomp, line 555: Hono's `route()` flattens the sub-app's `use("*")` handlers into the parent as `ALL /*`, so because both routers mount at `/`, `optionalEarnAuth`, `optionalEarnProjectContext` and `observeEarnAccessTier` run on **every keyed Earn route too**, ahead of `requireKeyedEarnCredential` and `unifiedAuthMiddleware`. `unifiedAuthMiddleware` doesn't short-circuit on an existing `c.get("apiKey")`, so `authenticateApiKeyRequest` runs twice per keyed request: two KV lookups, two uncached `enforceOrganizationIpAllowlist` reads, two `enforceRateLimit` charges (Clerk/session callers also run the project-membership query twice).

Verified against the real app: with the same key, `GET /v1/earn/strategies` recorded 1 charge on the key's limiter, `GET /v1/earn/programs` recorded 2 (counter went 1 -> 3 -> 5 over two keyed calls). The `sdp_api_earn_tier_request` log also fires with `tier: keyed` for `/v1/earn/programs`, which is the same leak. Net effect: every existing keyed Earn caller's effective rate limit is halved by this PR.

Fix: don't put the optional tier's cross-cutting middleware on a sub-app that shares the mount point. Simplest is to drop the three `optionalAuthEarn.use("*", ...)` lines and spread a shared tuple (`optionalEarnAuth, optionalEarnProjectContext, observeEarnAccessTier`) into each of the six keyless route declarations. The authz-matrix inventory test filters `ALL` routes so it won't catch this; worth a small test that asserts one charge per keyed request.

apps/sdp-api/src/routes/earn/index.ts
```diff
@@ -79,15 +98,163 @@ async function requireEarnFeature(c: Context<{ Bindings: Env }>, next: Next) {
...
+    );
+  }
+}
+optionalAuthEarn.use("*", observeEarnAccessTier);
+
+const EARN_CATALOGUE_CACHE_CONTROL = "public, max-age=30, stale-while-revalidate=30";
+async function cacheEarnCatalogue(c: Context<{ Bindings: Env }>, next: Next) {
+  await next();
+  if (c.res.status >= 200 && c.res.status < 300) {
+    c.header("Cache-Control", EARN_CATALOGUE_CACHE_CONTROL);
```
keeganthomp: `public` here is emitted for credentialed responses too, and the body varies by credential: on a production deployment a `sk_test_` key lists the devnet shelf while an anonymous caller gets the mainnet shelf, and a Clerk caller varies by `X-Project-Id`, all at the same URL. `public` is exactly the directive that lets a shared cache store a response to a request carrying `Authorization` (RFC 9111 §3.5), and there's no `Vary`, so any CDN or proxy in front can hand a sandbox key's shelf to anonymous prod callers for up to 60s. Keep the public directive for anonymous responses only (the `earn.test.ts` case asserting keyed == anonymous cache-control would flip):
```suggestion
    const authenticated = c.get("apiKey") || c.get("clerk") || c.get("session");
    c.header("Cache-Control", authenticated ? "private, max-age=30" : EARN_CATALOGUE_CACHE_CONTROL);
```

apps/sdp-api/src/routes/earn/index.ts
```diff
@@ -79,15 +98,163 @@ async function requireEarnFeature(c: Context<{ Bindings: Env }>, next: Next) {
...
+  createEarnExternalWalletWithdrawalTransaction
+);
+
+// Keyed routes retain dashboard auth, project membership checks, and their
+// existing permission matrix. Give anonymous callers the API-facing contract
+// before projectContextMiddleware can turn a missing project into a 400.
+async function requireKeyedEarnCredential(c: Context<{ Bindings: Env }>, next: Next) {
+  if (
+    !c.req.header("Authorization") &&
+    !getCookie(c, "sdp_session") &&
```
keeganthomp: nit: use `SESSION_COOKIE_NAME` from `@/routes/auth/constants` instead of the literal. If the cookie is ever renamed this check silently turns every dashboard Earn call into a 401 before session auth gets a look.

apps/sdp-api/src/lib/sdp-environment.ts
```diff
@@ -30,5 +30,20 @@ export function resolveSdpEnvironment(c: Context<{ Bindings: Env }>): SdpEnviron
...
+    if (deploymentEnvironment === "sandbox" || deploymentEnvironment === "production") {
+      return deploymentEnvironment;
+    }
+    throw internalError("SDP_ENVIRONMENT must be sandbox or production");
+  }
+
+  if (c.env.ENVIRONMENT === "development") {
+    return "sandbox";
+  }
+  if (c.env.ENVIRONMENT === "production") {
```
keeganthomp: Question on the fallback: if api-dev and api-stage run as production-mode builds (`ENVIRONMENT=production`, the self-hosted compose default) while pointing Earn at devnet, then without `SDP_ENVIRONMENT=sandbox` set on those Cloud Run services in sdp-infra an anonymous call there resolves to `production` and lists/builds against the mainnet shelf, while the guide tells keyless users api-preview is devnet. Is that infra change part of the preview-deployment blocker? If so, worth naming it in the PR body so it doesn't get lost.

apps/sdp-api/src/services/earn/vault-external-wallet.service.ts
```diff
@@ -364,37 +415,38 @@ export async function buildExternalWalletDepositTransaction(
...
+      swapLeg,
+      // The split swap is one of the transactions this flow hands out, so
+      // the partner fee payer covers it too and co-signs before the owner
+      // broadcasts it, exactly like the deposit it precedes.
+      fee,
+    });
+    if (!hasExternalWalletBuildTenant(input)) {
+      return { kind: "swap_required", swap: swapLeg, swapTransaction };
+    }
+    const baseline = await deadline.run("Reading the split-swap baseline balance", () =>
```
keeganthomp: nit: this used to run in `Promise.all` with the compile so the baseline read cost the partner's blockhash window nothing; it's now a serial RPC round trip on the keyed split-swap path. Easy to keep both: `Promise.all([compile, hasTenant ? deadline.run(...) : Promise.resolve(null)])` and branch on the result. Non-blocking.

keeganthomp (commented):
Solid split overall: anonymous builds never touch tenant rows, `rejectInvalid` closes the expired-key downgrade, and the tests pin the tier matrix. Two things I'd fix before this leaves draft: the optional-tier `use("*")` middleware leaks onto every keyed Earn route because both sub-apps are mounted at `/`, so keyed callers now pay two auth passes and two rate-limit charges per request (verified live, details inline at `apps/sdp-api/src/routes/earn/index.ts:535`), and `Cache-Control: public` is emitted on credentialed catalogue responses that vary by credential (`index.ts:167`). One deployment question on the `ENVIRONMENT=production` fallback and two nits inline. Self-authored, so this is a comment rather than a verdict.

apps/sdp-api/src/routes/earn/index.ts
keeganthomp: Addressed in [`5ec2f217e6c9`](https://github.com/solana-foundation/solana-developer-platform/commit/5ec2f217e6c92973f285d71892be6094accd65a3): Scoped optional-auth middleware to the six keyless routes; a regression proves one keyed request records exactly one API-key charge, with 199 focused tests passing.

apps/sdp-api/src/routes/earn/index.ts
keeganthomp: Addressed in [`5ec2f217e6c9`](https://github.com/solana-foundation/solana-developer-platform/commit/5ec2f217e6c92973f285d71892be6094accd65a3): Anonymous catalogue responses remain public while authenticated responses now use private caching; route tests cover both headers.

apps/sdp-api/src/routes/earn/index.ts
keeganthomp: Addressed in [`5ec2f217e6c9`](https://github.com/solana-foundation/solana-developer-platform/commit/5ec2f217e6c92973f285d71892be6094accd65a3): Replaced the session-cookie literal with SESSION_COOKIE_NAME, with API typechecking and module boundaries passing.

apps/sdp-api/src/lib/sdp-environment.ts
keeganthomp: Addressed in [`5ec2f217e6c9`](https://github.com/solana-foundation/solana-developer-platform/commit/5ec2f217e6c92973f285d71892be6094accd65a3): Updated the PR blocker to require SDP_ENVIRONMENT=sandbox on the non-production Cloud Run service before preview smoke testing.

apps/sdp-api/src/services/earn/vault-external-wallet.service.ts
keeganthomp: Addressed in [`5ec2f217e6c9`](https://github.com/solana-foundation/solana-developer-platform/commit/5ec2f217e6c92973f285d71892be6094accd65a3): Restored concurrent standalone-swap compilation and keyed baseline reads while anonymous builds still skip the baseline, with 39 service tests passing.

apps/sdp-api/src/routes/earn/handlers/external-wallet.ts
```diff
@@ -525,25 +528,129 @@ function earnedUnavailableReason(
...
+    };
+  }
+
+  if (!("strategyId" in locator)) {
+    throw badRequest(
+      "strategyId and ownerAddress are required for an anonymous external-wallet exit"
+    );
+  }
+  const strategy = await getEarnRepository(c).getStrategyById(locator.strategyId);
+  if (!strategy || strategy.environment !== environment) throw notFound("Earn strategy");
```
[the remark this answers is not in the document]
keeganthomp: Addressed in [`4aa95393cfaf`](https://github.com/solana-foundation/solana-developer-platform/commit/4aa95393cfafaf21650a60789a518683c0e2e93d): Preserved delisted strategies as sync-owned metadata tombstones so anonymous previews and exits keep resolving. Repository and route regressions cover delisting, exits, and safe relisting.

apps/sdp-api/src/middleware/metered-quota.ts
```diff
@@ -98,3 +108,57 @@ export function meteredQuota(config: MeteredQuotaConfig) {
...
+      await next();
+      return;
+    }
+
+    const actor = `ip:${getClientIp(c) ?? "unknown"}`;
+    try {
+      await enforceRateLimit(
+        c,
+        `metered:${config.name}:anonymous:${actor}`,
+        config.maxRequests(c.env),
```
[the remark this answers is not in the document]
keeganthomp: Addressed in [`4aa95393cfaf`](https://github.com/solana-foundation/solana-developer-platform/commit/4aa95393cfafaf21650a60789a518683c0e2e93d): Made forwarded-header trust opt-in outside Cloud Run. Direct and default self-hosted callers now share the fail-closed unidentified quota bucket, with client-IP and quota regression coverage.

cutbow7 (changes requested):
# Review: feat(earn): open API to keyless integrations (#1807)

**Reviewed at** `0330c6d1` in an isolated worktree. I built, tested, and ran the code — every claim below was executed, not inferred. All commands via pnpm (never npm).

## Verdict

**Approve pending the named security sign-off the PR itself requests.** The implementation is careful, the tier isolation holds under both the test suite and a live server, and the generated artifacts are drift-free. The two items below (#1, #2) should be resolved before or immediately after merge.

## What I verified by running it

**Static checks (all green):**

- `pnpm install --frozen-lockfile` — lockfile intact
- `pnpm typecheck` — 23/23 workspace packages
- 417 tests across the touched suites, incl. real Postgres 16 / Redis 7 testcontainers:
  - `earn.test.ts`, `earn.route-authz-matrix.test.ts`, `earn.external-wallet.test.ts`, `earn.vault.test.ts`, `spec.test.ts` (246)
  - `rate-limit`, `metered-quota`, `client-ip`, `sdp-environment`, `earn.repository`, `earn-catalogue-sync`, `vault-external-wallet.service` (171)
- `pnpm check:module-boundaries`, `pnpm check:api-playground` — pass
- Full regeneration pipeline (`openapi:generate`, `generate:api-playground`, `generate:api`, `generate:ai`) → `git status` **clean**, zero drift
- `check:links` (55 pages / 272 links) + docs build — pass

**Live server proof** (booted the API with `pnpm dev` against real Postgres/Redis containers, real HTTP over the wire):

| Probe | Result |
| --- | --- |
| `GET /v1/earn/strategies` anonymous | `200`, `cache-control: public, max-age=30, stale-while-revalidate=30` |
| `POST /v1/earn/vault-deposits` anonymous (keyed route) | `401 UNAUTHORIZED` — "API key required for this Earn route" |
| Invalid `sk_` key on a keyless route | `401 INVALID_API_KEY` — no downgrade to anonymous |
| `POST /external-wallet/deposit-transactions` `{}` anonymous | `400` validation — reaches the handler, not an auth wall |
| Anonymous withdrawal build with `positionId` form | `400` — "strategyId and ownerAddress are required for an anonymous external-wallet exit" |
| Anonymous catalogue ×130, one IP | exactly 120×`200` then `429` with `retry-after` / `x-ratelimit-*` headers |
| Second IP during that 429 | `200` — per-IP isolation holds |
| `POST` build from the 429'd IP | `400`, **not** 429 — catalogue and build buckets are separate counters |
| Keyed catalogue with a seeded valid key | `200`, `cache-control: private, max-age=30` (the credentialed-cache fix landed) |
| Keyed positions route with a valid key | passes auth + membership into validation (`400` on missing `ownerAddress`) |

## Strengths

- **One declaration per endpoint.** The `OPTIONAL_EARN_ACCESS_MIDDLEWARE` tuple (`routes/earn/index.ts:167`) plus the keyed router's `requireKeyedEarnCredential` (`routes/earn/index.ts:263`) is the correct fix for the `use("*")` double-mount leak found in self-review — verified: keyed callers get one auth pass and their keyed quotas unchanged.
- **The authz matrix ratchet** (`earn.route-authz-matrix.test.ts:39`) forces every new `/v1/earn` route to declare keyless-vs-keyed, and pins invalid-key downgrade rejection per route. This is the test that keeps this tier honest.
- **Anonymous builds are structurally ephemeral.** `ExternalWalletBuildContext`'s anonymous branch makes tenant fields `never`; handler and service both re-assert owner-pays and no persistence; anonymous split-swaps write no advisory.
- **Delisting → tombstone** (`0099` migration + `deprecateUnlistedStrategies`) preserves exit metadata for keyless holders while keeping tombstones out of browse and deposit admission. The upsert `CASE` ordering (`earn.repository.postgres.ts:159`) correctly reactivates only sync-owned tombstones; operator pauses stay sticky.
- **`TRUST_PROXY_HEADERS` fails closed**, with an explicit opt-in field in the generated self-hosted configurator.

## Issues

1. **[Must-address before launch] Public OpenAPI still says `apiKeyAuth`-only on all six keyless routes, while the Embedded Yield guide now teaches keyless usage.** Confirmed `security: [{"apiKeyAuth": []}]` on `/v1/earn/strategies`, `/vault-deposit-previews`, and all three external-wallet build/preview paths in the generated spec. This is the deliberate "conservative contract until PRO-1943 sign-off" state per the PR description and `spec.test.ts` — but it is a live contract mismatch: an integrator generating a client from `openapi.json` believes a key is required. It fails safe (stricter docs than runtime), but the follow-up that flips the `security` declarations to the exact optional-auth matrix must land promptly after sign-off, with docs and spec in the same change. **Named security reviewer still needed here, as the PR requests.**
2. **[Deployment precondition, not enforced anywhere]** Anonymous environment resolution (`lib/sdp-environment.ts:33-49`): a deployment with `ENVIRONMENT=production` and no `SDP_ENVIRONMENT` serves keyless routes against the **mainnet** shelf. The PR description says the non-prod Cloud Run service must set `SDP_ENVIRONMENT=sandbox` before smoke testing — that's a runbook item someone can forget. Suggest logging the resolved anonymous environment once at boot (e.g. `anonymous earn environment: sandbox|production`) so a misconfigured deploy is visible in the first log line rather than in traffic.

## Nitpicks (non-blocking)

3. Unknown paths under `/v1/earn` now answer `401 "API key required"` for anonymous callers instead of 404 (keyed router catches unmatched paths before the optional router 404s — both mount at `/`, `routes/earn/index.ts:555`). Verified live. Doesn't leak route existence; cosmetic only.
4. The 30/min anonymous build tier (`rate-limit.ts:23`) is hardcoded while the RPC budget next to it is env-tunable (`EARN_ANONYMOUS_RPC_*`). Fine for now; consider unifying if self-hosters ask for knobs.
5. Self-hosted default (`TRUST_PROXY_HEADERS=false`) collapses all anonymous callers into one `unknown` IP bucket (120 catalogue/min **shared**). Deliberate and fail-closed, and the env-reference help text covers it — just be aware one noisy client can starve keyless traffic behind a direct exposure.
6. `api-preview.solana.com` appears as string literals in `embedded-yield.mdx` (the site constant only carries production). Coherent with `sdp-web/playwright/env.ts`; consider a shared constant eventually.

## Security sign-off checklist (per AGENTS.md / routes/earn/CLAUDE.md)

- [ ] Named security reviewer recorded in this PR (PRO-1872 / threat model EARN-027)
- [ ] Threat-model row for the keyless subset added or referenced
- [ ] Follow-up tracked to flip the public `security` declarations to the optional-auth matrix

cutbow7:
So, no code changes required, just pls verify those things are done ^^^^ then I will approve this @keeganthomp

keeganthomp:
@cutbow7 Verified on `2acfcbbf`:

- PRO-1872 explicitly maps public OpenAPI promotion to threat-model row EARN-027.
- PRO-1948 tracks the exact follow-up that changes the six keyless operations to the optional-auth OpenAPI matrix.
- A named security sign-off is not yet recorded. PRO-1872's implementation PR #1747 was approved by @multipletwigs, but I am not treating that historical approval as sign-off for this PR. The public OpenAPI remains API-key-only until the current reviewer is named and approves.
- Latest `main` is merged and the PR is conflict-free.

The PR body now records those links and the pending gate. Local validation passed: full workspace typecheck, 214 focused API tests, 208 Earn package tests, 16 Ondo tests, 49 Treasury UI tests, module boundaries, API playground drift, generated-artifact drift, docs links, and docs build.

cutbow7 (dismissed):

resourcefulmind (commented):
Read all 67 files with Thomas's checklist in hand. The code holds up: anonymous requests set no tenant identity and persist nothing, invalid keys still 401, keyed quotas are untouched, and cutbow7's live checks all still stand at this head. Five things before merge:

- The security checklist itself isn't done. No named reviewer in the PR, no threat-model row, no follow-up tracked for flipping the public security declarations. Who's signing this off?
- Non-prod services need SDP_ENVIRONMENT=sandbox before this deploys, otherwise keyless routes on dev and stage serve the mainnet shelf. It's only in the PR body today. A boot-time log of the resolved environment would make a miss visible.
- A stale session cookie, or a Clerk token without an org, proceeds as anonymous on the six routes instead of 401, which is the silent downgrade the auth comment says can't happen.
- The guide's keyed snippet now points at api-preview.
- Question for Nikhil: is prod Cloud Run ingress load-balancer only? The forwarded-for parsing assumes it.

keeganthomp:
Addressed the repository-actionable items in `6011b61f`:

- Stale session cookies and Clerk tokens without organization context now return 401 on the optional Earn tier. The route matrix covers all six keyless routes.
- The API validates and logs the resolved anonymous Earn environment at startup.
- The authenticated guide now defaults to `https://api.solana.com`; preview remains limited to the keyless devnet walkthrough.
- The PR security checklist now maps PRO-1872 to EARN-027, links PRO-1948 for the optional-auth OpenAPI change, and keeps the public OpenAPI API-key-only pending named security sign-off.

Validation passed: 145 focused tests, API lint and typecheck, full workspace typecheck, module boundaries, docs links, and docs build.

Still needs external confirmation: the named security approver and Nikhil’s confirmation that production Cloud Run ingress is load-balancer-only.

multipletwigs:
Thanks for this — one thing I want to push back on before the rest of the review: `SDP_ENVIRONMENT` was deliberately removed from the API, and this PR brings it back (`types/env.d.ts`, `.env.local.example`, `turbo.json`, `secret-keys.mjs`, README, env-reference, plus `resolveAnonymousSdpEnvironment`).

**How environment works on main today**

Environment is a *project* property, not a deployment property. `resolveSdpEnvironment` has exactly two sources: the API key's project environment (via the auth JOIN) or the dashboard's selected project (`x-project-id`, verified against membership). Mainnet access is therefore "hold a production project's API key" — the same deployment serves sandbox and production tenants side by side, and `EARN_PROVIDER_VAULT_DIRECT_DEPOSIT_ENVIRONMENTS` decides per provider which environment may deposit. A request that has no project fails closed on purpose (#1204); the deleted docblock explains why defaulting either way is wrong.

**Why the new var is a problem**

- It re-adds a bespoke "which deployment am I" variable when `ENVIRONMENT` already carries that fact — the PR's own fallback branch proves `ENVIRONMENT` is sufficient.
- Two sources for one fact plus precedence logic (`SDP_ENVIRONMENT?.trim()` then `ENVIRONMENT`) is a fallback chain, which we don't do. It also silently widens `resolveSdpEnvironment` for *every* anonymous caller, not just the Earn routes.
- Ops now has to remember to set `SDP_ENVIRONMENT=sandbox` on non-prod Cloud Run; a missed setting on a dev deploy that runs with `ENVIRONMENT=production` would serve the mainnet catalogue and build mainnet transactions to anyone.

**Suggestion**

- Drop `SDP_ENVIRONMENT` everywhere (env type, examples, turbo, secret-keys, docs, startup log).
- Keep `resolveSdpEnvironment` fail-closed as on main. Give the four keyless Earn routes their own resolver that maps `ENVIRONMENT` → `SdpEnvironment` with an exhaustive `switch` and a `never` default, and pass it only where the handler has no tenant. That keeps "no project ⇒ throw" as the rule and "keyless Earn reads deployment env" as the single, explicit exception.
- The startup log can stay, sourced from `ENVIRONMENT` only.

Happy to talk through it if there's a constraint I'm missing on the Cloud Run side.

keeganthomp:
Addressed in `7a56b46f`: removed `SDP_ENVIRONMENT`, restored project-only fail-closed resolution, and added an Earn-only exhaustive `ENVIRONMENT` mapping used only for anonymous catalogue, quote, and build requests. Validated with API typecheck/build, 6 focused environment tests, Biome, module boundaries, docs generation, and docs link checks.

apps/sdp-api/src/db/repositories/earn.repository.postgres.ts
```diff
@@ -155,18 +156,17 @@ export function createPostgresEarnRepository(db: AppDb): EarnRepository {
...
+             -- lists the same reference again. Operator pauses/deprecations are
+             -- sticky and still outrank the catalogue.
              status = CASE
+               WHEN earn_strategies.catalogue_delisted_at IS NOT NULL
+                 THEN EXCLUDED.status
                WHEN earn_strategies.status IN ('paused', 'deprecated')
                  THEN earn_strategies.status
                ELSE EXCLUDED.status
              END,
+             catalogue_delisted_at = NULL,
```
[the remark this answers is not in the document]
keeganthomp: Addressed in [`41f84ce3d846`](https://github.com/solana-foundation/solana-developer-platform/commit/41f84ce3d8467ea2fd95085f07e9af373f2e732f): Operator status updates now clear the sync tombstone marker, so provider relists preserve both paused and deprecated stops; API build, typecheck, Biome, and module boundaries pass.

multipletwigs (approved):

cutbow7 (approved):

PR #1816, opened 2026-09-15
fix(payments): withdraw funding instructions once a ramp transfer is terminal

multipletwigs:
Re-hosts #1574 by @unsafePtr on an in-repo branch, rebased onto current main (142 commits). Code from the approved fork head `98daa799c`, plus follow-up commits: the component-local terminal set was replaced by `isTerminalRampTransferStatus` from `@sdp/types`, the inline comments were shortened, and the step-content render tests now stub sibling components via `vi.mock` so only the two files under test enter the sdp-web coverage denominator (the original tests pulled the full wizard trees in and broke the ratchet).

**Summary**

- A `canceled`, `failed` or `expired` manual-instructions transfer no longer renders its deposit address, bank details or payment reference; both step contents render `RampStatusPanel` alone once `isTerminalRampTransferStatus(transferStatus.status)` (the canonical `@sdp/types` predicate).
- The terminal check runs ahead of the missing-instructions guard, so a dead transfer reports "Transfer canceled" instead of a quote defect.
- `canceled` gets its own copy per direction instead of `Current provider status: canceled`; SDP cancels locally and never tells the provider, so the old copy misattributed the decision.
- `completed` never reaches the guard: both step contents route it to `RampCompleteScreen` earlier. An absent status stays fundable (quote renders before the first status arrives).
- Provider-agnostic: same path serves BVNK, Mural and Lightspark manual instructions.

**before** — terminal transfer on the manual-instructions step:

1. Only `completed` swapped the screen to `RampCompleteScreen`.
2. `canceled`/`failed`/`expired` fell through to `ManualInstructionsQuote` with copyable funding details beside an error panel.
3. `canceled` rendered as the provider's status, contradicting the provider's own open record.

**after**:

1. `isTerminalRampTransferStatus` → render `RampStatusPanel` only, no instructions.
2. `canceled` → "Transfer canceled" + "instructions no longer apply, create a new quote".
3. Fundable statuses and `undefined` unchanged.

**Verification**

- `pnpm exec tsc --noEmit` (sdp-web) at rebased head — clean.
- `terminal-transfer-instructions.unit.test.tsx` — 20 tests across both directions: every terminal status withdraws `ManualInstructionsQuote`, every fundable status and the pre-status render keep it, and a canceled transfer with absent instructions reports the cancellation, not a quote defect.
- `ramp-status-panel.unit.test.tsx` — 2 new tests: canceled copy per direction never says "Current provider status".
- sdp-web coverage after the change: statements 68.71% / branches 61.42% / functions 65.76% / lines 69.56% against thresholds 68.57 / 61.26 / 65.76 / 69.45.
- Local app walk-through: not run for this re-host.

Supersedes #1574.


resourcefulmind (approved):
Approved.

Walked both rails. The terminal check sits above the missing-instructions guard in each, `completed` still gets intercepted upstream before it can reach the new branch, and a new quote comes back with no status yet so the instructions still render. All holds.

You can ignore the Greptile P2, it predates the render tests and doesn't apply to this head.

`completed` is in the terminal set too, so the completion screen only survives because its branch runs first. Your TERMINAL list leaves it out deliberately, which makes sense, but that also means nothing catches it if someone reorders those branches later. One more case would lock it down. It's not a blocker though

PR #1823, opened 2026-09-15
refactor(types): move status subsets and predicates into @sdp/types

multipletwigs:
Every status vocabulary outside recurring payments is defined once in `@sdp/types` as an `as const` tuple, and every meaning over it (fundable, terminal, in-flight, settled, blocking, historical, …) is a named export beside the tuple. Call sites import the name instead of re-listing literals. No user-visible change; this removes the drift surface where a status added to a type never reaches the `Set` in a service, the `IN (...)` in a repository, or the guard in a component.

**What moved**
- DvP trades, private-channel transfers/deposits/withdrawals, sponsorship reservations, earn strategies/movements, custody + RPC connections (now one shared connection lifecycle), provider credentials, API keys, tokens/issuance transactions, wallet operations, approval requests: tuple + derived type + subsets/predicates in `packages/sdp-types`, each with a one-line JSDoc stating the meaning.
- Provider-owned vocabularies stay in their packages with the same shape: DFNS signature states (`packages/sdp-custody`, module-private), Helius Rings operation states (`packages/sdp-helius-rings`).
- Existing `z.enum([...])` route schemas now take the tuples.

**Call sites**
- Local `new Set<Status>([...])`, `const X: Status[] = [...]`, `=== "a" || === "b"` chains, and `.includes` lists are replaced by the named subset or predicate.
- Repository SQL binds the tuple once with `column = ANY(?::text[])` / `NOT (column = ANY(?::text[]))` instead of literal `IN (...)` lists or placeholder builders.
- Tests import the same names; scenario tables derive from the tuple plus the production predicate under test rather than re-listing statuses.

**Fixes found along the way**
- Helius Rings' terminal-operation set omitted `voided`, which the vocabulary documents as terminal; the no-fail-edge set is now derived from `failEdgeFor` instead of a second hand-maintained list.
- The wallet-operation approval CAS built a dynamic SQL fragment plus a conditional spread; it now binds one from-status array under static SQL.

**before** — adding a status to a vocabulary:

1. Extend the TypeScript union.
2. Find every `Set`, array, `===` chain, and SQL `IN` list that encodes a meaning over that vocabulary.
3. Miss one; the new status is silently "not terminal" / "not fundable" in that one place.

**after** — adding a status to a vocabulary:

1. Extend the tuple; the derived type follows.
2. `Record<Status, boolean>` maps fail to compile until the new status is classified; list subsets are reviewed in one file.
3. Every consumer already imports the name; SQL binds the same array.

## How a subset is declared

```ts
export const FUNDABLE_DVP_TRADE_STATUSES = [
  "created",
  "partially_funded",
] as const satisfies readonly DvpTradeStatus[];

/** Reports whether a trade can receive another funding leg. */
export function isFundableDvpTradeStatus(status: DvpTradeStatus): boolean {
  return FUNDABLE_DVP_TRADE_STATUSES.some((candidate) => candidate === status);
}
```

```sql
WHERE t.status = ANY(?::text[])   -- bound once with [...UNSETTLED_DVP_TRADE_STATUSES]
```

Every replaced literal list was diffed against its new subset; all members are identical (two DvP rejection tests additionally cover `rejected`, which they had omitted).

**Verification**

- `pnpm --filter @sdp/types exec tsc --noEmit`, `@sdp/api`, `sdp-web`, `@sdp/custody`, `@sdp/helius-rings` — all pass
- `git diff --name-only origin/main -- '*.ts' '*.tsx' | xargs pnpm exec biome check` — 100 files, clean
- `git diff --check` — clean
- Status-literal grep gate: remaining hits outside the owning modules are single-status assertions/assignments or exhaustive display maps
- DB-backed suites — not run locally; CI is the gate

Known gaps
- Recurring-payment statuses are handled in the sibling PR for HOO-1638; this PR deliberately does not touch those files.


multipletwigs:
Closing in favour of a stacked split by risk (see Linear HOO-1643 mechanical sweep, HOO-1644 DvP, HOO-1645 Helius Rings voided-terminal fix, HOO-1646 connection lifecycle + approval CAS). Branch kept as the reference implementation.

PR #1827, opened 2026-09-15
feat(issuance): gate freeze, unfreeze, pause, and unpause behind wallet-operation policy

G1de0n:
Freeze, unfreeze, pause, and unpause sign on-chain with custody authority but ran with no wallet-operation policy gate — the deferral noted in #1752, confirmed by the 2026-09-14 STRIDE sweep. An org could not require approval on, or deny, any of these token-admin operations.

This adds four operation types (`issuance_freeze_execute`, `issuance_unfreeze_execute`, `issuance_pause_execute`, `issuance_unpause_execute`) and gates the four routes with `policyGate` using candidate extractors that mirror the burn/seize pattern: token lookup, idempotency-replay short-circuit (`candidate: null`, matching existing replay semantics), authority and wallet resolution, and runtime admission before the candidate is built — so an unavailable wallet still answers 404/409 rather than failing inside enforcement. Handlers now also hold the executing wallet to the policy-judged identity (`assertJudgedCustodyWallet` + approved-operation custody check), the same drift guard burn and seize carry.

Dry-Run on these routes now returns the policy verdict like other gated issuance routes. One behavioral note: a dry-run can fail on wallet availability or provider entitlement, since admission runs during extraction.

Tests: four deny tests proving a denied operation stops before the signer with the wallet operation recorded; conformance entries for all four routes; full sdp-api suite green.


apps/sdp-api/src/routes/issuance/handlers/freeze.ts
```diff
@@ -687,3 +694,117 @@ export const unfreezeAccount = async (c: ValidatedBodyContext<typeof unfreezeSch
...
+    resolved: { judgedCustodyWalletId: wallet.custodyWalletId },
+    candidate: buildIssuancePolicyCandidate({
+      auth,
+      token,
+      custodyWalletId: wallet.custodyWalletId,
+      walletId: wallet.providerWalletId,
+      operationType,
+      amount: null,
+      destination: body.accountAddress,
+    }),
```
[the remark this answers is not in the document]
G1de0n: Fixed in 499d530b — the extractor now resolves the freeze target itself and the candidate's destination is the resolved token account, so policy judges the account the operation will modify.

apps/sdp-api/src/security/value-moving-conformance.node.test.ts
```diff
@@ -110,6 +110,86 @@ const contracts: ValueMovingContract[] = [
...
+      section: '"/tokens/:tokenId/unfreeze",',
+      before: "policyGate({ extract: extractUnfreezePolicyCandidate })",
+      after: "unfreezeAccount",
+    },
+    replay: [
+      {
+        mode: "idempotency_fingerprint",
+        file: "apps/sdp-api/src/routes/issuance.test.ts",
+        evidence: "stops a denied unfreeze before signer and issuance side effects",
+      },
```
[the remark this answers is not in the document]
G1de0n: Fixed in 499d530b — pause/unpause entries now cite the existing exact-wallet replay test, and unfreeze gained its own persisted-account replay test that the entry cites.

apps/sdp-api/src/routes/issuance/handlers/pause.ts
```diff
@@ -99,6 +102,8 @@ export const pauseToken = async (c: ValidatedBodyContext<typeof pauseTokenSchema
     requestedCustodyWalletId: body.signingCustodyWalletId,
     requiredWalletPermissions: ["tokens:admin"],
   });
+  assertJudgedCustodyWallet(c, custodyWalletId);
+  await assertApprovedWalletOperationCustodyWallet(c, custodyWalletId);
```
GuiBibeau, line 118: **[P1] Check the execution lease just before submitting**

These checks verify the wallet, but they do not check whether this worker still owns the approved execution. I expired the lease while the signer loaded: freeze, unfreeze, pause and unpause still called Mosaic. `execution_effect_started_at` also stayed null when each call began.

Please call `beginApprovedWalletOperationEffect(c)` immediately before each Mosaic submission, as the burn handler does. This stops expired workers from submitting and lets recovery detect a worker that disappeared after submission. Both handlers in this file and the two in `freeze.ts` need this guard.

apps/sdp-api/src/routes/issuance/handlers/pause.ts
```diff
@@ -390,3 +397,114 @@ export const unpauseToken = async (c: ValidatedBodyContext<typeof pauseTokenSche
...
+          buildIdempotencyMetadata(idempotencyKey, {
+            tokenId,
+            operation,
+            mode: "execute",
+            params: { ...body, signingCustodyWalletId: custodyWalletId },
+          }).idempotencyFingerprint,
+      })
+    : null;
+  if (replay) {
+    return { ...emptyExtraction, candidate: null };
```
GuiBibeau, line 482: **[P2] Do not mark a pending replay as completed**

`replay` can be a pending transaction with no signature. During approved recovery, this skips the gate and the handler returns HTTP 200 for that pending transaction. The approval runner then marks the wallet operation `completed`, even though the transaction has no confirmed result.

I reproduced this for pause, unpause and unfreeze (with an existing frozen-account record): recovery made no Mosaic call but marked the operation completed.

Please handle approved pending replays like the burn/authority handlers: recover settlement evidence first, then require manual reconciliation if the transaction is still unsettled. Cover both early replay and the `createTransaction` replay path.

apps/sdp-api/src/routes/issuance/handlers/freeze.ts
```diff
@@ -687,3 +694,127 @@ export const unfreezeAccount = async (c: ValidatedBodyContext<typeof unfreezeSch
...
+          buildIdempotencyMetadata(idempotencyKey, {
+            tokenId,
+            operation,
+            mode: "execute",
+            params: { ...body, signingCustodyWalletId: custodyWalletId },
+          }).idempotencyFingerprint,
+      })
+    : null;
+  if (replay) {
+    return { ...emptyExtraction, candidate: null };
```
GuiBibeau, line 772: **[P2] Evaluate current policy for dry-run requests**

This also returns `candidate: null` for `Dry-Run: true`. The policy gate treats a null candidate as ungoverned and returns `allow`.

I configured an always-deny policy and sent the same dry-run with two keys. A fresh key returned `deny`; a key matching a settled transaction returned `allow` with no criteria. This happens for all four actions.

The gate's documented dry-run behavior is to evaluate current policy even when the key matches an old transaction. Please keep the policy candidate for dry-run requests and use this replay shortcut only for execution requests. The matching branch in `pause.ts` needs the same change.

GuiBibeau (commented):
Local QA found three issues, noted inline. The existing targeted suites passed (299 tests), along with API typecheck, lint and module-boundary checks. I added 27 temporary QA cases: 12 passed and 15 reproduced the three issues below. These tests use real PostgreSQL/Redis and the HTTP app, with Solana reads and submissions mocked.

apps/sdp-api/src/routes/issuance/handlers/pause.ts
G1de0n: Fixed in c069b1a3 — `beginApprovedWalletOperationEffect(c)` now runs immediately before each Mosaic submission in all four handlers (freeze, unfreeze, pause, unpause), matching burn.

apps/sdp-api/src/routes/issuance/handlers/pause.ts
G1de0n: Fixed in c069b1a3 — both replay paths (early replay and the createTransaction replay) in all four handlers now fence approved pending replays the way burn/authority do: unsettled transaction + approved operation → effect marker + 409 manual-reconciliation conflict. Your QA cases were the guide here; would appreciate a re-run of them.

apps/sdp-api/src/routes/issuance/handlers/freeze.ts
G1de0n: Fixed in c069b1a3 — the replay shortcut is now execution-only (`idempotencyKey && !isDryRunRequest(c)`), so dry-run always builds the candidate and evaluates current policy; added a regression test where a key matching a settled freeze transaction still gets `deny` on dry-run.

GuiBibeau (dismissed):

GuiBibeau (approved):

PR #1833, opened 2026-09-15
fix(payments): show a payment held for approval as waiting, and send it under one key

tobySolutions:
PRO-1950. A payment a wallet policy holds for approval showed "Transfer failed", and pressing Send again opened a second approval for the same payment. Split down from the original six tickets; the rest moved to #1839 and a separate approvals PR.

**Pay**
- A 202 SIGNING_PENDING is read as held, not failed: "Waiting for approval", the payment summary, and a link to the approval request.
- Every send carries an `Idempotency-Key` kept per payment in the tab (same store as batches). A retry of the same payment reuses it, so if two approvals exist the second replays the first transfer instead of sending again. The dashboard proxy now forwards the header.
- Finished screens (held or sent) have one heading.

**Off-ramp and MoneyGram**
- Off-ramp: a held send says it's waiting, hides the transfer status pill (the row still reads awaiting payment because the gate answers before the handler runs), says the quote can expire first, and links to the approval.
- MoneyGram needs a signature right away, so a held transfer stops the widget with a message to cancel the request in Approvals if it shouldn't go out later.

Tests: api policy route test (same key twice gives one transfer, no key gives two); web 2434 passed.

Known gaps, for the policy thread rather than this PR:
- The API still opens a new approval for each retry, even with the same key. The key stops the second send, not the second approval.
- Replays aren't marked, so repeating an identical approved payment in the same tab returns the earlier transfer once.
- After a reload the off-ramp forgets the hold and offers Send again. Approving both sends once, because the ramp row only moves from awaiting payment once.


tobySolutions:
Replaced by #1870, which carries the held-payment fix alone. The rest of what was here went to #1839 (load times and fee copy) and #1841 (approvals).

PR #1851, opened 2026-09-16
refactor(issuance): retire unused Token-2022 operations

GuiBibeau:
Token creation, minting, freezing, and thawing already use `MosaicService`; `Token2022Service` duplicated those operations for integration fixtures. Move the DvP/custom-mint fixtures to Mosaic and remove the unused methods, types, exports, and `createCustomToken` fallback. Production burn/prepare-burn behavior and sponsorship stay unchanged.

- Removes about 500 lines of obsolete production code.
- Adds eight burn regression cases covering validation, real Kit signing, sponsor handoff, unsigned simulation, and confirmation errors with bigint serialization.
- Renames the custom-mint fixture to reflect Mosaic, documents executing mint authority behavior, and removes the constant burn-error argument.
- Runs both migrated fixtures on Surfpool and adds DvP to the CI matrix. Repairs its local custody setup, distinct external party, markets flag, shim program configuration, and stale initial-balance assertions. DvP coverage is creation, funding, and on-chain/API readback in both directions; it does not settle the trade.

Validation: eight burn tests; real Surfpool DvP cases (2/2), Mosaic custom mint (1/1), and Kora shim tests (2/2), with no skips; four affected package/app typechecks; Biome, module boundaries, YAML fixture-path validation, and diff checks. Independent QA reviewed the deletions and follow-up changes. Earlier local checks also passed 49 Solana/issuance package tests and 295 API tests (nine existing skips).

The project-scoped Privy fixture setup is typechecked but was not executed locally. The live fixture runs use isolated local custody and Kora on Surfpool.


GuiBibeau:
@greptileai review

GuiBibeau:
Ready for QA at `859c2f239`.

Independent review confirmed that deleted operations have no production callers, burn/prepare-burn source checks and signing behavior remain unchanged, and the DvP/custom-mint fixtures preserve amounts, signer identity, and sponsorship through Mosaic. Greptile's latest review is 5/5 with no actionable inline threads, and all required checks are green.

Local checks covered four typechecks, 49 package tests, and 295 API tests (nine existing skips). Live-chain fixtures were not run locally; CI's Surfpool/Kora integration shards passed. The Access shard passed after rerunning a Doppler connection-reset failure before test execution.

QA focus: direct and sponsored burns, unsigned burn simulation, custom-template minting, and both DvP settlement directions. Not merged.

eldarik:
**Approve with nits 🟡 — gated on running the two migrated fixtures once.**

I checked the deletions independently and they hold up: `createCustomToken` has no call site anywhere, and `createMint`/`mintTo`/`freeze*`/`thaw*`/`prepare*` on `Token2022Service` are reachable only from the two fixtures this PR migrates. I also traced the fixture parity through mosaic-sdk — `enableTokenAcl` unset gives `enableSrfc37: false`, `custom-token.ts` returns early before any Token-ACL/ABL setup, so the `aclMode: "allowlist"` the template path adds is inert and `freezeAuthority` passes through unchanged; `uri: ""` keeps `createToken` off its packet-overflow branch. Fee-payer resolution and decimal-amount semantics match on both sides. Nice cut.

**Risk**
- The fixture repointing is the only behavioural change in the PR and neither fixture was executed. Typecheck can't see that `MosaicService.mintTo` discards `options.mintAuthority`, that `resolveTokenAccount` moved ahead of submission, or that `mint` is now optional. One devnet/surfpool run of `dvp-settlement.test.ts` + `token2022.test.ts` and this is done.

**P2**
- `MosaicService.mintTo` (`mosaic/service.ts:541`) accepts `mintAuthority` and then always signs with `this.signer`; only `prepareMintTo` honours it. Correct here because they're the same key, but the migrated call site reads as if it's choosing one — worth a comment or dropping the field from the executing overload.
- `signAndSubmit`'s `confirmation.err` → `"Burn failed: …"` branch is now the class's whole error surface and no new test hits it, on either path. One `confirmTransaction` mock returning `{ err }` per path covers it (and `safeStringify` with it).
- `packages/sdp-api-integration/src/tests/token2022.test.ts` no longer tests `Token2022Service` — filename, case name ("creates mint using service directly") and fixture metadata (`"Token2022 Direct"` / `T2022`) are all stale, and it now shadows `mosaic-templates.test.ts`'s custom-template case. Rename or fold in.

**P3**
- `signAndSubmit(fullTx, sdkRpc, failureMessage)` has two call sites passing the same `rpc` and the same `"Burn failed"` — both params are dead.
- It's a burn service now, but `Token2022Service` / `token-2022.ts` / `Token2022Env` (now just `= RpcEnv`) / `createToken2022Service` keep the old name. Follow-up rename, not this PR.
- Deleting the freeze/thaw short-circuits removes the last *production* read of `SOLANA_MOCK` in the repo; it now survives only as a test-skip flag in `env.d.ts`, `scripts/secret-keys.mjs`, `test/helpers/env.ts` and the integration vitest config.
- The new test spies on the live `@solana/mosaic-sdk` / `@sdp/rpc/solana` namespaces without the `vi.mock(…, importOriginal)` copy the repo uses for this (`issuance.test.ts:50-53`); it only works because `vitest.config.ts` inlines mosaic-sdk under `server.deps.inline`. A comment naming that dependency would help.

@GuiBibeau let me know WDYT, I'll approve on your call

GuiBibeau:
Addressed the points I agree with:

- Added direct and sponsored burn confirmation-error cases, including bigint serialization in the `Burn failed: …` message. All eight burn tests pass.
- Documented that executing `MosaicService.mintTo` uses the service signer; the migrated call uses that same signer.
- Renamed the custom-mint fixture and its case/metadata to Mosaic, updated CI and the README, and documented the Vitest namespace-spy dependency.
- Removed the constant error-message argument from `signAndSubmit`. It has one remaining caller; the RPC argument is still used for submission/confirmation, so I retained it.

The earlier CI already executed the custom-mint fixture successfully. DvP was absent from the CI matrix, so this update adds it and repairs the local fixture setup: enable markets, advertise the DvP program in the Surfpool shim, use a separate local settlement authority, and use a distinct external party address.

Executed against isolated Surfpool with local custody and the Kora shim: both DvP directions pass (2/2, no skips), the renamed Mosaic mint fixture passes (1/1), and the shim tests pass (2/2). Each DvP case receives 201 on creation and 200 on funding/readback, then checks the exact on-chain escrow balance. All four affected typechecks pass. The fixture also initializes project-scoped Privy custody when selected; that provider branch was typechecked but not executed locally.

The existing DvP cases check creation, funding, and on-chain balance/readback in both directions; they do not call settle. Their descriptions now say what they actually cover. The service-wide rename and remaining `SOLANA_MOCK` cleanup stay as separate follow-ups.

GuiBibeau:
@greptileai review

GuiBibeau:
Ready for QA at `6aa07012b` after addressing the review.

Greptile reviewed this commit at 5/5 with no inline findings, independent QA found no remaining issues, and all 26 required checks are green.

Both migrated fixtures ran locally and in [Surfpool / Issuance Core](https://github.com/solana-foundation/solana-developer-platform/actions/runs/35073246500/job/104719489703): Mosaic custom mint 1/1 and DvP 2/2, with no skips. DvP covers creation, funding, exact on-chain balances, and API readback for both legs; it does not test final settlement. Local checks also passed eight burn cases (including both confirmation-error paths), two shim tests, and all four affected typechecks.

The project-scoped Privy fixture branch is typechecked but was not exercised locally. The service-wide rename and remaining `SOLANA_MOCK` cleanup remain separate follow-ups. Not merged.

API unit shard 6/8 initially stalled without producing test results and hit the 20-minute job limit. The same commit passed an independent local run of all 49 files / 780 tests with coverage; only that CI shard and its dependent checks were retried.

eldarik (approved):

PR #1867, opened 2026-09-16
refactor(markets): dedupe quote notices, floors, tones and route helpers

cutbow7:
Behavior-preserving cleanups in the Markets module:

- earn: one VaultQuoteNotices component and one derivedMinOut floor derivation shared by the vault deposit and withdraw modals (the two copies differed only in field names and translation keys)
- earn: one vaultOutcomeTone mapping and one TransactionLink instead of the deposit/withdraw modal twins
- earn: one blocking-issues schema fragment behind both preview envelopes; named the two refresh cadences the hooks repeat as bare literals; delete the dead findProgram helper and the never-adopted EarnStrategyIdentity component
- dvp: shared next-step builders for the closed/creating/settled legs every trade kind restates; dvpTimestampToIso replaces three inline u64-seconds conversions; CopyableAddress copies through the shared useCopy hook; delete the dead partiesStepSchema/termsStepSchema exports and un-export the internal isDvpTradeOpen
- BFF: forwardedIdempotencyHeaders replaces the five copies of the upstream Idempotency-Key forwarding dance


tobySolutions (approved):
LGTM ser

PR #1871, opened 2026-09-16
feat(payments): move BVNK individual onboarding to v1 customers with agreement sessions

multipletwigs:
- BVNK individual onboarding moves from the v2 customer API to the platform v1 chain: agreement session → signed consent → customer create → JIT customer GET. Webhooks and ledger v2 wallets stay on v2.
- The pre-customer link row reserves our `externalReference` (counterparty id minus the `cpty_` prefix) and is CAS-flipped to BVNK's native reference on create, or claimed by the `platform:customer:update` webhook if the row lost the race.
- Payment rules send the individual's PII read just-in-time from the v1 customer GET (`bvnkRuleEntityFromCustomer`). Nothing PII-shaped is persisted; without the entity BVNK rejects the rule with `bvnk:0000 Validation failed`, which is why every on-ramp stalled at provisioning before.
- Webhook processing parses with per-event zod schemas (`bvnk.schemas.ts`), handles the `bvnk:platform:customer:status-change` VERIFIED event that was previously acked as ignored, and routes wallet events by their parsed name.
- The on-ramp rule request is persisted before the KYC gate, so the VERIFIED webhook provisions the funding wallet itself instead of waiting for another manual advance.
- Shared ramp status sets (`SETTLEABLE_RAMP_TRANSFER_STATUSES`, `TERMINAL_RAMP_TRANSFER_STATUSES`) replace SQL literals in both settlement handlers.
- Address requirement sub-keys follow the Places field names (`line1`, `subdivisionCode`) so autofill fills them; the collector maps to BVNK's wire names. Autofill skips option values the provider does not offer.
- Residence list is US-only until a merchant legal entity is verified (DE/SE are rejected by the sandbox entity).

**before** — v2 customer create:

1. `POST /platform/v2/customers` with agreements accepted inline.
2. Customer GET returned a sticky `verificationUrl`; KYC phase derived from it.
3. Rule create sent `customerIdentifier` only → 400 `bvnk:0000`; `provisioningError` stored; every counterparty stuck at `customer_funding_account_provisioning`.
4. `bvnk:platform:customer:status-change` fell into the default branch and was acked as ignored, and the advance returned `customer_verifying` before persisting the rule request, so even a handled VERIFIED webhook had nothing to provision.

**after** — v1 agreement sessions:

1. `POST /platform/v1/customers/agreement/sessions` → agreements surfaced in the dashboard card; `PUT …/sessions/{ref} {status:"SIGNED", ipAddress}` on consent.
2. `POST /platform/v1/customers` with `externalReference` = counterparty uuid; link row CAS-flipped from the reservation to the native reference.
3. `GET /platform/v1/customers/{ref}` refreshes `status`/`verification.status`; Sumsub URL surfaced JIT only when `INFO_REQUIRED`.
4. On VERIFIED (either status-change event) the webhook finds the persisted rule request → wallet create → rule create with the individual entity → ACTIVE. No further dashboard action needed.
5. `ACCOUNTS-2000` with `errors.externalReference` recovers the existing customer; other `ACCOUNTS-2000` codes (session reuse, legal entity) fail loudly.

## How the reservation claim works

```text
persistBvnkAgreementSession   INSERT customer_link  provider_customer_reference = <counterparty uuid>   (the claim)
createBvnkCustomer            POST v1/customers {externalReference: <uuid>} → {reference: R}
assignCustomerLinkReference   UPDATE … SET provider_customer_reference = R WHERE provider_customer_reference = <uuid>  (CAS)
webhook platform:customer:update {reference: R, externalReference: <uuid>}
  lookup by R  → hit  : refresh + provision
             → miss : parseBvnkCustomerExternalReference(<uuid>) → cpty_<uuid> → same CAS → refresh + provision
```

## API before/after

`POST /v1/counterparties/{id}/requirements` (bvnk, onramp) — agreement stage response:

```jsonc
// before
{ "status": "customer_agreement_required", "agreements": [{ "id": "…", "name": "…", "description": "…", "downloadUrl": "…" }] }
// after
{ "status": "customer_agreement_required", "agreements": [{ "name": "…", "displayName": "…", "description": "…", "url": "…", "privacyPolicyUrl": "…" }] }
```

Webhook `POST /v1/webhooks/ramps/bvnk` — new event handled:

```jsonc
{ "event": "bvnk:platform:customer:status-change", "data": { "status": "VERIFIED", "reference": "<native ref>" } }
// before: 200 {received:true}, logged "ignoring event"
// after : 200 {received:true}, customer refreshed, pending on-ramps provisioned
```

## Schema before/after

Migration `0111_bvnk_customer_link_sessions.sql` archives v2-era `customer_link` rows (no `status` and no `session` in metadata) and strips the vestigial `agreements` key. No DDL.

**Screenshots**

**Verification**

- `tsc --noEmit` — api, web, sdp-payments, sdp-types: pass
- `biome check` on all changed files — clean
- `pnpm -C apps/sdp-api playground:check` — pass
- hermes-verify browser run (third pass) — PASS: dashboard flow reaches `customer_funding_account_provisioning` with zero 500s
- Live sandbox probes (fake person data): v1 session → sign → create → GET chain; rule create ACTIVE on the app's exact wallet reference and entity shape; wallet create webhook echoes the wallet name
- Not run locally: vitest (CI). `apps/sdp-api/src/openapi/__snapshots__/spec.test.ts.snap` needs `vitest -u` for the agreements shape change.

Known gaps
- Consent `ipAddress` falls back to `0.0.0.0` when `getClientIp` resolves nothing (local, no trusted proxy). Temporary until client-IP plumbing lands.
- Off-ramp settlement now writes `fiat_amount` from `walletAmount` only; the `displayAmount` fallback was removed.
- Session `expiresOn` is not persisted; an expired session surfaces as an `ACCOUNTS-2000` on create.
- `simulatePayin` sandbox helper keeps its `??` defaults.
- **`counterparties.provider_data` is legacy and must not grow.** BVNK on-ramp wallet + rule state (`walletId`, `walletStatus`, `bankAccount`, `ruleId`, `ruleStatus`, `request`, `provisioningError`) still lives in `provider_data.bvnk.onramp.wallets[key]`, read/written by `ensureBvnkPaymentRule`, the webhook route, `ramps.ts`, and `provider-data.ts`. PRO-1823 moves it onto a `counterparty_provider_accounts` row of kind `funding_wallet`: the row insert is the claim before the BVNK wallet create, the row id (prefix stripped) becomes the wallet name, and the webhook reconciles by id with a `customerReference` cross-check. The `TODO(PRO-1823)`/`TODO(PRO-1824)` markers in the webhook route point at the two writers.
- `openBvnkCustomerLink` duplicates `openExternalRampUrl` with an allowlist; five call sites outside this PR.
- `ensureBvnkPaymentRule` GETs the customer that the caller just refreshed (one extra provider call per advance).

Agreement consent card

Verification state


multipletwigs:
closing in favor of virtual wallets for now - no actual payout

PR #1881, opened 2026-09-17
perf(web): stop Home activity wallet fan-out

GuiBibeau:
## What changed

- load Home payment activity with one organization-wide persisted transfer query instead of one observed-chain query per wallet
- keep observed chain history on wallet-specific surfaces where its RPC cost is intentional
- reduce Home activity polling from every 20 seconds to every 60 seconds
- explicitly pause interval refreshes while the tab is hidden or offline
- add regressions for the single-request path and refresh policy

This complements #1839: that PR limits how long Home waits for a wallet, while this PR removes the Home wallet fan-out entirely.

## Local verification

- `vitest` targeted Home route/workspace tests: 12/12 passed
- `pnpm --filter sdp-web typecheck`: passed
- Biome check on all changed files: passed
- Full sdp-web suite: 2,450 tests passed; 25 unrelated tests failed because jsdom exposed no `localStorage` under the local Node runtime
- ESLint could not run because `eslint-plugin-react` crashes while loading `react/display-name` under ESLint 10, before linting changed files

## UI

No visual layout or copy changes; this changes only data-fetch behavior and refresh cadence.


GuiBibeau:
@greptileai review

GuiBibeau:
QA-ready. Greptile is 5/5 with no review threads, and all required CI checks are green, including dashboard E2E, Docker builds/smoke, typecheck, lint, and Vercel preview.

Suggested QA focus: confirm Home Recent Transactions still shows persisted payment and issuance activity, then leave the tab open/backgrounded to verify the activity feed refreshes at the new one-minute visible-only cadence.

resourcefulmind (approved):
Checked the row sets, the org-wide query was already the first leg of the old
path, so the only thing that disappears is observed-chain rows. Looking good on this side. 

However, i have noticed that today's volume now counts only SDP-persisted transfers, so an org
with inbound deposits made outside SDP sees that number drop with no copy
change. Do you think we should tweak the labeln Gui?

PR #1899, opened 2026-09-17
fix(release): stop the breaking-change detector flagging its own commit

resourcefulmind:
Blocks the 0.79.0 release notes. Raised at the release gate.

The first changelog the new renderer produced opens with this:

```
### ⚠ BREAKING CHANGES

* **release:** surface breaking changes in the changelog (HOO-1662) (#1888)
```

0.79.0 has no breaking change. The `v0.78.0..main` range carries no `!` subject marker, and the only commit whose body matches "breaking change" is #1888 itself, where the phrase is prose describing what the feature does:

> a breaking commit keeps its entry in its type section and also surfaces in a BREAKING CHANGES section on top

The detector matches that anywhere in the body rather than requiring a real `BREAKING CHANGE:` footer, so the commit that added breaking-change reporting announced itself on its first run.

**What this changes**

One entry added to `nonBreakingCommitOverrides`, the set you already keep for this case. It sits next to the existing entry for a commit that carried an inaccurate footer, and the same set already feeds both the bump selection and the new renderer (`release-changelog.mjs:69`).

**Verification**

- `buildSectionMarkdown` run over the real `v0.78.0..main` range, before and after: section emitted `true` before, `false` after. Version unchanged at 0.79.0, since the range has real features.
- `node --test scripts/*.test.mjs`: 144 pass.
- `biome check` on the changed file: clean.

**What this deliberately does not fix**

The detector still matches prose instead of a footer, so the next commit that writes about breaking changes gets flagged the same way. That is the real bug and it wants its own change, not a release-day patch. Happy to file it if you agree with the read.


resourcefulmind:
Superseded by #1900, which fixes the detector itself rather than silencing this one commit.

Adding f87bb92 here would have misattributed the fault: that commit's footer is fine, the detector was matching prose anywhere in the body. #1900 makes it read a real BREAKING CHANGE footer, which drops #1888 out of the breaking set on its own and leaves the override list meaning what it says.

PR #1902, opened 2026-09-17
refactor(markets): delete dead earn/dvp code and name shared rules once

cutbow7:
Behavior-preserving cleanups in the Markets module:

- earn: delete the never-called wallet-activity toast hook, the retired custodial create/retarget write path, the unused program-deposits feed (hook + fetcher), hasPrograms, and the withdrawal tracking reset helper no test ever called; one requestJsonOk gate replaces the five copies of the non-2xx guard the remaining readers restated
- earn: one isPositiveDecimal predicate replaces the eight inline compareUnsignedDecimals(x, "0") spellings across the deposit and withdraw money surfaces
- dvp: one pastedAddressOption builder for the four copies of the paste-an-address combobox option; export isDvpTradeOpen so the detail page's canReclaimLeg reads the OPEN set instead of re-spelling it; one dvpToastAction for the two identical explorer-link success toasts; one dvpErrorEnvelopeSchema for the three copies of the refusal envelope
- dvp: delete the unused DvpPartySlotMode type, the never-imported DVP_TRADE_SIDES/DvpTradeSide re-exports, and the unused DvpPartyWire/ partySlotSchema re-exports; delete the production-dead fromBaseUnits adapter


keeganthomp (approved):
No regressions found. I verified each deleted export against the full base tree; none has a production caller, and the Treasury-owned Earn seams remain intact. The shared helpers preserve the prior response validation, decimal, toast, and DvP open-state behavior. Focused Earn/DvP tests (494), web typecheck, and Biome checks pass.

tobySolutions (approved):
LGTM to me chief @cutbow7. I checked and verified that the integrations do not disrupt anything on my ends.

PR #1908, opened 2026-09-17
fix(custody): unblock Privy credential checks for empty applications

pashpashkin:
**What changed / why**

Use `GET /v1/wallets` without query parameters for Privy credential validation. The previous request failed for an empty application despite valid credentials, blocking connection setup.

Direct Privy probes with the same credentials and empty app:

| Request | Observed result |
|---|---|
| Before: `/v1/wallets?limit=1&chain_type=solana` | `500` after ~55s; SDP times out at 10s |
| After: `/v1/wallets` | `200`, empty `data` array, ~440ms |

`limit=1` alone also failed. The check now uses Privy's default first page.

**Impact**

- Applies to the shared credential check used by connection setup, rotation and rollback.
- `chain_type` is unnecessary for authentication; wallet creation still explicitly requires Solana.
- Preserves the 10s timeout, `401 → failed`, and other errors → `retry_unknown`.
- No changes to secret storage, audit behavior or public API contracts.

**Validation**

- Direct API comparison above; invalid-secret control returned `401`.
- Local dashboard: user confirmed Add connection succeeds with the same API fix.
- `pnpm -C apps/sdp-api exec vitest run src/services/custody/privy-credential.test.ts src/routes/internal-custody/provider-credential-check.test.ts src/routes/internal-custody/provider-credential-lifecycle.test.ts` — 134 passed.
- API typecheck and Biome on all three changed files — passed.


pashpashkin:
@greptileai review pls

G1de0n (approved):

PR #1917, opened 2026-09-17
refactor(payments): discriminate the on-ramp quote body by provider

resourcefulmind:
First of four for Coinbase on-ramp quotes, which fail today before any HTTP call because nothing collects the buyer contact Coinbase requires. This PR only makes room.

- The quote body is now discriminated by provider: seven arms, **field-for-field identical** to the single object they replace.
- **Every arm is a `strictObject`.** An arm that was not strict would silently accept keys the object rejected, which is the only way this could change behaviour.
- Arms are written out rather than generated, because PR 2 edits one of them.
- The documented schema follows the runtime one, so the spec renders a `oneOf` with every field description intact.
- One consumer changed: covering all arms narrows the request to `never` in its exhaustiveness branch.

Verification: payments.ramps 45, payments routes 335, openapi 56 including the spec snapshot. The five new tests were run against the previous flat schema first and passed identically. No generated artifact changed. Not run: the dashboard, which this does not touch.


resourcefulmind:
Closing to redo through the repo's own process. These were opened without running sdp-review over the change set and without sdp-make-pr, so the body and the review evidence were improvised rather than produced the standard way. Reopening both once that has actually been done. No code is changing; the branches stand.

PR #1918, opened 2026-09-17
feat(payments): accept buyer email and phone on the Coinbase on-ramp quote

resourcefulmind:
Second of four, stacked on #1917. Coinbase's headless create-order requires buyer contact details, so its client refused every quote before any HTTP call. The Coinbase arm now carries them.

- **Both fields are optional.** The provider client already refuses a quote without them, so omitting them fails exactly as today rather than changing the error a caller sees. Every other arm rejects them as unknown keys.
- **Neither value is persisted, and that needed a fix.** The policy gate embeds the request into `rawPayload.executionRequest`, and `rawPayload` is written to the wallet-operation row. This extraction set no `executionRequestBody`, so the gate fell back to the whole body: a buyer's email and phone into the policy audit on every quote.
- The extraction now names what a decision may remember, as an allowlist. A field added to the quote body joins the audit record only when someone adds it there too.

Verification: payments routes and openapi, 27 files, 398 tests. Four packages typecheck. Not run: the dashboard, which PR 3 covers.


resourcefulmind:
Closing to redo through the repo's own process. These were opened without running sdp-review over the change set and without sdp-make-pr, so the body and the review evidence were improvised rather than produced the standard way. Reopening both once that has actually been done. No code is changing; the branches stand.

PR #1921, opened 2026-09-17
feat(dvp): record fund, reclaim, settle and cancel in the audit ledger

tobySolutions:
Four DvP actions spend from a custody wallet and none of them left an `audit_logs` entry; the only audited part of the module was the settlement wallet's provisioning. Fund, reclaim, settle and cancel now each land one hash-chained event against the trade, naming the principal that authenticated the request.

Funding is money into escrow, so it goes through a fail-closed `beginCritical` intent: if the ledger cannot record the intent, the deposit is refused and nothing is sent. Reclaim, settle and cancel are exits, so they log after the effect and cannot block it. A ledger outage must never be the reason a deposit sits in escrow (ADR 0002 exit safety, the same asymmetry PRO-1866 settled for Earn).

The bracket is in `apps/sdp-api/src/routes/dvp/trade-audit.ts`, with the failure postures pinned in `apps/sdp-api/src/routes/dvp.audit.test.ts`. All nine tests were checked against the unwired handlers first and all nine fail there, so none of them is decoration.

Known gap: `trade-audit.ts` parallels `routes/earn/handlers/movement-audit.ts`, and a shared money-audit helper is the right end state. Folding it in now would mean editing Earn's audited seams from a DvP PR.

Not in here: none of these four actions goes through `policyGate`. That is PRO-1975.

PRO-1992


tobySolutions:
@greptile-apps review

G1de0n (approved):

PR #1922, opened 2026-09-17
feat(dvp): accept an Idempotency-Key on settle and cancel

tobySolutions:
Fund and reclaim answer a retry with the first request's signature. Settle and cancel took no key at all, so their only protection was the close lock, and that lock is taken after the blockhash and the authority signature. A retry while it is live gets a 409 and never learns the signature that did settle the trade. A retry after it expired on a close that DID land signs and sponsors a second close against an escrow the chain has already emptied, which fails and reports an error for a settlement that actually happened.

Both now run through the same machinery, generalized rather than copied: the sponsored transaction is written onto the request row before broadcast, and a retry asks the chain what it did. Migration 0112 widens the table to a trade-level action and makes the two absences exact, so a leg action must name its side and a close must not.

One deliberate difference from the leg path, in `services/dvp/leg-action-idempotency.ts`. A leg action frees its key on any 4xx, because every leg refusal is raised before broadcast. A close breaks that premise: one that was broadcast and then failed on chain throws 400 `dvp_close_failed_on_chain` (`settle.ts` closeOutcome). Freeing on the code would happen to be right there, but it rests on knowing which 4xx are post-broadcast, so the close frees its key only when nothing was ever recorded and otherwise lets the chain decide.

Interleavings, all reachable today: concurrent same key 409s; a reused key on another trade or action 409s; a refusal before signing frees the key; a crash between the close lock and the recording frees the key but `claimClose` 409s until the reconciler releases the expired lock; a recorded attempt that can still land 409s, one that landed replays, one that provably moved nothing runs again.

Sits on #1921, so it is a draft until that merges; the diff narrows to the second commit then.

PRO-1993


tobySolutions:
@greptile-apps review

tobySolutions:
@greptile-apps review

G1de0n (dismissed):

G1de0n (approved):

tobySolutions:
Closing as redundant.

#1927 was built on this branch and contained these commits, so PRO-1993 shipped in its squash (`72806ecc`).

Verified on `main`: migration `0112_dvp_action_requests_close_actions.sql`, `runDvpCloseOnce` in `closeTrade` (`apps/sdp-api/src/routes/dvp/handlers.ts:505`), `recordAttempt` in `apps/sdp-api/src/services/dvp/settle.ts`, the retry-replay conformance case, and the `Idempotency-Key` documentation on the DvP paths.

This branch is frozen at the pre-#1927 `handlers.ts`, so merging it would remove the settle policy gate rather than add anything.

PR #1929, opened 2026-09-18
feat(payments): BVNK on-ramp via reference-matched pay-ins and a reconciler payout

multipletwigs:
BVNK on-ramp: one funding wallet per counterparty stays; payment rules are gone. The payer puts the transfer id in the bank reference, the pay-in webhook matches it, and a BVNK-specific reconciler creates the payout explicitly with the transfer id as BVNK's unique `reference`.

- **Quote makes no provider write**: prebook `xfr_` → one bank-details GET → CAS `pending → awaiting_payment`. Instructions carry `paymentReference = "SDP-ONRAMP <xfr_id>"`, routing number, account, BIC.
- **Pay-in (v1 event)** regex-matches the `xfr_` id across `paymentReference + additionalRemittanceInformation` (the rail splits the string inconsistently), then ONE UPDATE: `awaiting_payment → settling`, gross fiat, `provider_data.bvnk.payin` first-write-wins; unique index on the pay-in id. **No match parks the event** (`TerminalRampWebhookError`).
- **Reconciler** (`reconcileBvnkOnrampPayouts`, 1-min cron): claim on `provider_data.bvnk.payout` → dry-run → `POST /api/v1/pay/summary` with `reference = xfr_id`; amount = received − fees. BVNK rejects a duplicate reference, so ambiguous outcomes recover by list-by-reference + `GET …/summary` before recording uuid + PROCESSING settlement atomically. Aged rows are polled.
- **Payout webhook** (`reference` = our id, `type: OUT`): PROCESSING no-op; COMPLETE/FAILED go through one shared terminal operation shared with the poll path, guarded by stored payout uuid; divergent terminal facts park, identical replays no-op.
- **Read model**: `providerReference` = BVNK payout uuid once it exists (omitted before); `settlement` is a `PROCESSING | COMPLETE` union with estimates and actuals; the wizard shows the success screen at PROCESSING with receipt link, fees, rate, pay-in id.
- **Removed**: rule create/list/deactivate, funding-wallet lock/release/repair, `funding_wallet_reserved/settling`, wizard cancel prompt. **Live bug fixed**: `GET /payment/v1/rules?walletId=` was ignored by BVNK (platform-wide list); moot now.

```mermaid
sequenceDiagram
  autonumber
  participant W as Dashboard wizard
  participant API as sdp-api
  participant DB as Postgres
  participant B as BVNK
  participant J as reconcileBvnkOnrampPayouts (cron)
  W->>API: POST requirements (advance)
  API->>DB: claim funding_wallet row (unique per counterparty+fiat)
  API->>B: create ledger wallet sdp:onramp:<row id>
  B-->>API: wallet:status-change ACTIVE (+FIAT instrument)
  API->>DB: CAS provisioning → provisioned
  W->>API: POST quote
  API->>DB: prebook xfr_ (pending, provider_data {bvnk:{}})
  API->>B: GET wallet (bank details, JIT)
  API->>DB: CAS pending → awaiting_payment
  API-->>W: instructions: paymentReference "SDP-ONRAMP xfr_…"
  B-->>API: payin:status-change COMPLETED (paymentReference + additionalRemittanceInformation)
  API->>DB: match xfr_ id → ONE UPDATE awaiting_payment → settling (payin id unique)
  J->>DB: claim payout (first-write-wins) + persist intent
  J->>B: dry-run, then POST /api/v1/pay/summary reference=xfr_
  B-->>J: uuid, PROCESSING, economics, redirectUrl
  J->>DB: record payout uuid + PROCESSING settlement (one UPDATE)
  B-->>API: crypto:status-change PROCESSING / COMPLETE / FAILED (reference=xfr_)
  API->>DB: terminal op: CAS settling → completed|failed guarded by uuid
  J->>B: GET /api/v1/pay/{uuid}/summary for aged settling rows (lost webhook)
  W->>API: GET transfer (poll) → providerReference=uuid, settlement, receiptUrl
```

**Environment boundaries.** Webhooks arrive on a per-environment route and are verified with that environment's HMAC secret; the environment then travels with every lookup. Transfer-by-id from the remittance, pay-in-by-id, funding wallet by BVNK reference, customer link, ACTIVE wallet and off-ramp channel handlers all join `projects.environment = <webhook environment>`; a sandbox-signed event naming a production row resolves to nothing and is parked as terminal. The reconciler builds its BVNK runtime per candidate from the owning project's environment, so sandbox rows use sandbox credentials. Receipt URLs are stored only when host matches the environment (`pay.sandbox.bvnk.com` / `pay.bvnk.com`) and path is `/payout/<uuid>`. Migration 0113's destructive updates join `projects.environment = 'sandbox'`; the sandbox simulate route claims the transfer before its POST.

| Migration 0113 | |
|---|---|
| `payment_transfers_bvnk_onramp_payin_id_unique` | unique on `provider_data->'bvnk'->'payin'->>'id'` where provider=bvnk, type=onramp |
| sandbox funding rows in the retired lock status | reset to `provisioned_funding_wallet`, metadata `{}` |
| sandbox in-flight BVNK on-ramps | `failed`, error "BVNK on-ramp payment rules retired" |

Ceilings (PR notes): crash after prebook leaves a `pending` transfer (cancelable); a late deposit for a terminal transfer is parked for manual refund; payout fees are an estimate, insufficient funds → `failed` + manual; sandbox never completes payouts, so COMPLETE is covered by tests and the poll path only.

Verification: tsc clean (types, payments, api, web); biome clean; OpenAPI + playground regenerated, no drift; vitest not run locally (CI). Sandbox end-to-end on this branch: VERIFIED → wallet → quote → simulated pay-in with `SDP-ONRAMP xfr_…` → `settling` → reconciler payout (9.9 USD + 0.1 fee = 10) → PROCESSING webhook with `reference = xfr_…`. Audits: Greptile P1s + four astra rounds (read-only) applied; two live BVNK probes in `docs/_devlog/HOO-1658` (local). local app (Hermes browser run, counterparty V4): requirements ready → instructions with `SDP-ONRAMP xfr_…`, routing 101019644 → simulate → success screen at PROCESSING (payout uuid, receipt link, fee 0.10, rate 0.998, pay-in id, no tx hash) → counterparty details modal — PASS 6/6, console clean.

**Screenshots**

funding instructions with the payment reference
success screen while the payout is processing at BVNK
transfer details modal with payout uuid and receipt

funding instructions with the payment reference

success screen while the payout is processing at BVNK

transfer details modal with payout uuid and receipt


tobySolutions:
Read the whole branch, this is solid. All from reading, nothing observed live, nothing blocking.

**A FAILED payout with no address never gets recorded.** `bvnkPayoutObservationMismatches` null-guards everything except `destination`/`network` (`settlement.ts:274-279`), which `buildObservation` leaves null on a failed status with no address (`settlement.ts:157-158`) and the superRefine only requires for COMPLETE (`bvnk.schema.ts:126-141`). Reads as a conflict, parks, transfer stays `settling`, poll retries into it every 10 min. The fixture always sets an address (`test/helpers/bvnk.ts:271`) so it's untested. Does BVNK omit it on FAILED?

**0112 archives every active funding wallet, no provider or environment filter**, while every destructive statement in 0113 joins `environment = 'sandbox'`. Takes production rows too, and makes 0113's third UPDATE a no-op. Intended? It's the one that isn't reversible.

Three smaller:
- `bvnkWalletBankAccount` takes `paymentInstruments?.[0]` (`bvnk.ts:221`, same at `1383`) where you filter `type === "FIAT"` at `1512` and `bvnk.schema.ts:156`.
- `adoptBvnkFundingWalletByName` reads currency only off `balance` (`bvnk.ts:1185`) and throws, though the schema has both `currency` and `balance` optional, so provisioning wouldn't recover.
- Unclaimed branch is `updated_at ASC LIMIT 20` and a pre-write failure doesn't touch `updated_at`, so a permanently failing row stays at the head. The poll branch guards this with `markPolled` in the catch.

multipletwigs:
@greptile-apps mural is not live yet

apps/sdp-api/src/db/migrations/postgres/0112_bvnk_funding_wallet_per_fiat.sql
```diff
@@ -0,0 +1,11 @@
+UPDATE counterparty_provider_accounts
+SET status = 'archived',
+    metadata = '{}'::jsonb,
+    updated_at = sdp_iso_now()
+WHERE kind = 'funding_wallet'
```
keeganthomp: [P1] Do not archive production wallets before preserving their transfers

This archives every active funding wallet across providers and environments. In the next migration, 0113 intentionally preserves production `funding_wallet_locked` rows and in-flight production transfers, but those wallets have already been archived here. The new payout reconciler only joins active funding wallets, and the legacy `payment:v2:payin:status-change` webhook is now acknowledged without processing, so a production transfer can remain in `settling` after its wallet is made unreachable, potentially with deposited fiat. Scope this reset to the intended BVNK sandbox rows, or retire production wallets and transfers together. Please add a sequence-level migration test that seeds production state before applying 0112 and 0113.

keeganthomp (changes requested):
Reviewed all 60 changed files at `39b1412`, including the migration sequence, BVNK client and schema updates, webhook attribution and idempotency, payout reconciliation and recovery, public types and OpenAPI, and dashboard states.

The transaction claiming and duplicate-payout recovery are well defended, but the production migration interaction below can orphan an in-flight money movement, so I’m requesting changes.

Local validation passed: payments 118 tests; focused web 38 tests; API, web, and payments typechecks; Biome; module boundaries; API playground drift. DB-backed API tests were unavailable locally because no container runtime was present; all eight API unit-test shards are green in CI.

apps/sdp-api/src/db/migrations/postgres/0112_bvnk_funding_wallet_per_fiat.sql
multipletwigs: Fixed in 82805aeb0.

- 0112 now archives only BVNK funding wallets whose project is `sandbox` (same `projects` join 0113 uses); the index swap stays global. Production wallets and their in-flight transfers are untouched.
- Added the sequence test you asked for in the 0113 suite: seeds a production locked wallet + `settling` transfer and a sandbox pair under the pre-0112 index, applies 0112 then 0113, asserts production is unchanged and the sandbox pair is archived/failed.
- The sequence test also caught that 0113's locked-wallet reset lacked `status = 'active'` and rewrote the row 0112 had just archived; added the predicate.
- 0112's own test now asserts a production row survives with its metadata intact.

apps/sdp-api/src/db/migrations/postgres/0112_bvnk_funding_wallet_per_fiat.sql
```diff
@@ -0,0 +1,15 @@
...
+WHERE prj.id = cpa.project_id
+  AND prj.environment = 'sandbox'
+  AND cpa.provider = 'bvnk'
+  AND cpa.kind = 'funding_wallet'
+  AND cpa.status = 'active';
+
+DROP INDEX counterparty_provider_accounts_active_funding_wallet_unique;
+CREATE UNIQUE INDEX counterparty_provider_accounts_active_funding_wallet_unique
+    ON counterparty_provider_accounts(counterparty_id, provider, fiat_currency)
+    WHERE status = 'active' AND kind = 'funding_wallet';
```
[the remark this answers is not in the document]
multipletwigs: Fixed in 980f3ea76. Before the index swap, 0112 now archives every active funding wallet that has a more recently updated active sibling for the same counterparty, provider and fiat (ties broken by id), keeping the newest one active and leaving the archived rows' metadata intact. The 0112 test seeds two active production wallets for one counterparty/fiat under the old onrampKey index and asserts the older one is archived, the newer stays active, and the new unique index builds.

resourcefulmind:
- The internalError to badRequest change does not do what the bot asked. It sits in process(), which runs in waitUntil after the 200, so it never becomes a 400 and the row still burns ten retries then parks. The superRefine you wrote for the crypto payout is the shape that works, it rejects in parse() pre-ack.
- counterparties/handlers.ts, gateOnCollectedFields went from `stage === null ||` to `stage !== null &&`. Only the stage === null case changes, and it turns a collect response into a 400. Which state was that fixing?

keeganthomp (approved):

PR #1931, opened 2026-09-18
perf(payments): cache immutable observed-transfer transaction bodies

cutbow7:
Every includeObserved=true transfers-list read fetched the parsed body of each non-persisted on-chain signature with an individual getTransaction RPC call (up to historyLimit = 200 per request at concurrency 5), and third-party inbound transfers are never persisted, so the same immutable bodies were re-downloaded on every page view and every 20s dashboard poll. Memoize positive getTransaction results per signature in a bounded in-memory cache (FIFO, 1000 entries, 1h TTL) and coalesce concurrent lookups of the same signature. Signature history still comes fresh from getSignaturesForAddress on every request, and null results (transaction not yet indexed at the confirmed commitment) and RPC failures are never cached, so a just-submitted payment still appears as soon as the chain reports it - no user-visible freshness is traded for the speedup.

Measured with a simulated-latency harness (25ms per RPC, 200 signatures/request): 1018ms and 200 billed RPC calls cold (origin/main behavior on every request) vs 26ms and 5 calls for a rolling poll with 5 new signatures, and ~1ms with 0 calls for a warm repeat.

I also had this dialogue with the clanker:

how will scale? do we expect 1000 cache size to be good enough? can you estimate various scenarios?
Short answer: 1000 is comfortably sized for the common cases and degrades gracefully — worst case reverts to today's behavior, never worse. Here's the math....


keeganthomp (changes requested):
Requesting changes for the unresolved P1 in the existing inline thread. The cache holds the complete `confirmed` response for an hour, so a fork can leave an observed payment with stale slot, timestamp, status, inner instructions, or token balances even though signature history is refreshed. Cache only finalized responses, or leave confirmed results uncached until they finalize.

multipletwigs:
Reviewed the cache path. Code is good: null results and RPC rejections never get cached, in-flight lookups coalesce and drain in `finally`, the FIFO bound can't be exceeded since the write path has no await, and tenant scoping still happens per request in `buildObservedTransferRows` so a shared signature-keyed cache adds no cross-tenant surface. In-process is the right call for immutable bodies; multi-instance only lowers the hit rate, never correctness.

multipletwigs (dismissed):

keeganthomp (dismissed):

keeganthomp (approved):

keeganthomp (dismissed):

keeganthomp (approved):
No blocking findings. The cache now populates only from fresh finalized history, keeps fork-sensitive slot and timestamp metadata fresh, and preserves retry behavior for nulls and RPC failures. Focused cache tests, API and RPC type checks, Biome, and required CI all pass.

multipletwigs (commented):
Design and the Solana-side reasoning check out (history `confirmationStatus` as the finality gate, `getTransaction` carrying no finality field, fresh slot/blockTime over the cached body). Two type tightenings before merge:

1. `packages/sdp-rpc/src/solana.ts` — `SignatureInfo.confirmationStatus?: Commitment | null` is optional only so older mocks compile. Kit types the RPC field as `Commitment | null` (never absent) and the mapper always sets it, so make it required `Commitment | null` and update the mocks. The "never caches when the history entry carries no confirmation status" test then goes away with it, since that state can't be produced by the RPC layer.

2. `observed-transfers.ts` — `resolveObservedSlot(freshSlot: bigint | null | undefined, cachedSlot)` guards a case the type rules out: `SignatureInfo.slot` is a non-optional `bigint`. Drop the helper and use `Number(signatureInfo.slot)` directly; the cached-body slot fallback is dead.

PR #1932, opened 2026-09-18
fix(earn): scrub provider error prose at the sdp-earn fetch seam

keeganthomp:
PRO-1995: `@sdp/earn`'s provider fetch helper now scrubs what it lifts out of a failure body, as the ramps twin already does.

- **The provider's message goes through `scrubTelemetryString`** before it becomes `SdpEarnError.message`, which API callers receive and the envelope only credential-redacts.
- `errorDetails` output goes through `scrubTelemetry` inside the existing swallow guard: a scrub failure drops the details, never passes a body fragment through.
- **`@sdp/redaction` is now sdp-earn's second dependency.** No runtime deps of its own, so the catalogue cron still loads no chain SDK. Allowlisted in `check-module-boundaries.mjs`, module map regenerated.
- PII scrubbing policy and sdp-earn CLAUDE.md updated at the seam.

Verification: sdp-earn 148 tests pass (5 new); tsc clean for sdp-earn, kamino, veda, jupiter-lend, ondo, sdp-api; Biome, module-boundary and pinned-dependency checks pass. Not run: sdp-api vitest (no app code changed).


cutbow7 (approved):
LGTM

PR #1939, opened 2026-09-18
feat(earn): harden mainnet embedded yield execution

keeganthomp:
## What

Makes Embedded Yield cluster-safe on devnet and mainnet, and hardens Earn transaction builds for live mainnet execution without Kora.

- Validates the RPC genesis hash, selects cluster-aware strategies and mints, and links receipts to the correct Explorer cluster.
- Passes a partner payer through Jupiter's build contract while keeping the owner as swap authority; adds compact memo and split-swap fallbacks for packet limits.
- Publishes Kamino deposit minimums and derives `minSharesOut` from the live share quote with the shared 10 bps policy.

## Reviewer notes

- Kamino deposits now require `minSharesOut` in sandbox and production. Integrators must quote with `POST /v1/earn/vault-deposit-previews` and submit the derived floor.
- The payer is admitted only as the exact ATA rent funder; unexpected signer slots are rejected.
- Some USDG vault plans cannot fit a separate fee payer. The API returns a clear owner-paid retry instead of an oversized transaction.
- The current mainnet catalog has no PYUSD strategy; the PYUSD Token-2022 path remains covered by Jupiter contract tests.

## Verification

- API focused: 110 tests; typecheck
- Web modal: 48 tests; typecheck
- Embedded Yield: 60 tests; typecheck; production build
- Kamino: 128 tests, 3 skipped; typecheck
- Live mainnet: all 6 catalog strategies, 18 confirmed transactions, 16 finalized movements, zero open positions


cutbow7 (changes requested):
# Review: PR #1939 — feat(earn): harden mainnet embedded yield execution

**Verdict: Request changes** — the core mechanics are sound and well-tested, but the Kamino floor-policy change breaks **71 tests across 4 API route suites**. CI will be red as-is. Everything else (payer hardening, compact-memo retry, local oversize classification, cluster guards) verified working.

**Verification performed** (isolated worktree at `0431572b`, base `f37fd04c`):

| Check | Result |
| --- | --- |
| React Doctor (`--scope changed --base origin base --project sdp-web,sdp-docs`) | ✅ 100/100, no diagnostics |
| `pnpm install` + typecheck (sdp-api, sdp-web, sdp-kamino, sdp-types, example) | ✅ all pass |
| Biome lint | ✅ only pre-existing `apps/sdp-web/src/app/icon.svg` a11y error (untouched by this PR) |
| Unit tests: sdp-kamino / example / api earn services / web earn / other earn routes | ✅ 128/128, 60/60, 110/110, 188/188, 174+71 pass |
| Route tests vs base (both branches run) | ❌ 71 failures caused by this PR (below) |
| Genesis-hash constants vs live RPC (`api.devnet`/`api.mainnet-beta`) | ✅ match; live `assertRpcCluster` accepts devnet, refuses devnet-when-mainnet configured |

---

## 🔴 Merge-blocking: Kamino floor policy change was not applied to the existing suites

`packages/sdp-types/src/provider-access.ts:248` flips `kamino: null` → `{ defaultToleranceBps: 10 }`, so `assertDepositFloorPresent` (`apps/sdp-api/src/routes/earn/handlers/admission.ts:184`) now **400s every sandbox Kamino deposit without `minSharesOut`**. The PR adds a new test asserting this and updates the docs, but leaves four suites asserting the old contract. I ran each file on both branches to confirm attribution:

| File | Base | This PR | Failing tests |
| --- | --- | --- | --- |
| `apps/sdp-api/src/routes/earn.vault.test.ts` | 66/66 ✅ | 20/66 ❌ | 46 |
| `apps/sdp-api/src/routes/earn.external-wallet.test.ts` | 57/57 ✅ | 40/57 ❌ | 17 |
| `apps/sdp-api/src/routes/earn.vault-exposure-cap.test.ts` | 10/10 ✅ | 3/10 ❌ | 7 |
| `apps/sdp-api/src/routes/earn.test.ts` | 27/27 ✅ | 26/27 ❌ | 1 |

Failure mode (verified in the logs): `400 "minSharesOut is required: kamino deposits carry a share floor in every environment. Quote the deposit with POST /v1/earn/vault-deposit-previews and derive the floor from sharesOut."` where the tests expected `202` or a different 400.

**Fix** (pick one, then green the suites to match):

1. If every-environment flooring is intended (the PR's stated intent): update the seeded-deposit requests in `earn.vault.test.ts` (e.g. the `postVaultDeposit` bodies around line 479+), `earn.external-wallet.test.ts` ("money-in gates" block) and `earn.vault-exposure-cap.test.ts` (bodies at lines 379, 434, 496, 559) to pass `minSharesOut`, keeping the tests that assert *other* 400s focused on their own error; and fix the stale assertion + comment in `earn.test.ts:490` (expects sandbox `depositSlippage: null`; must now be `{ quoteRequired: true, defaultToleranceBps: 10 }` — the comment at lines 473–476, "Kamino declares no floor of its own", is also now wrong).
2. If sandbox should stay floorless, scope the map entry instead of updating tests — but that contradicts the PR description and the new test added in this PR, so option 1 is the consistent path.

## 🟡 Non-blocking

1. **Breaking API change worth announcing** — every existing integrator posting sandbox Kamino deposits without `minSharesOut` starts receiving 400s on deploy. The dashboard is safe (it reads the same catalogue `depositSlippage`), but the PR body should call out the contract change and the migration ("quote, then send the floor").
2. **Second attempt could also compact the memo** — in the swap-funded split path (`vault-external-wallet.service.ts:456`, `attempt = await attemptBuild(RETRY_SWAP_MAX_ACCOUNTS)`), the narrowed-route retry keeps the full-size memo. A build that only overflows because of the 96-byte fee-payer signer might fit a narrower route with the compact memo, saving the partner the extra split-swap transaction. Safe as-is (the `swapLeg && TooLarge → fit:false` conversion inside `attemptBuild`'s try means this attempt can't throw), just an optimization opportunity.
3. **Comment nit** — `vault-execution.service.ts:79` says the compact memo carries "128 bits"; `digest("base64url").slice(0, 22)` is 132 bits. The claim holds a fortiori; the number is just off.
4. **Cosmetic** — `examples/embedded-yield-bank/src/App.tsx:297` labels a missing-cluster toast "devnet" (optional chaining falls through). Fine for a demo.

## ✅ What I verified works

- **Payer hardening (`jupiter-swap.service.ts`)**: the fee/rent payer is admitted as a signer **only** on the exact 6-account idempotent ATA-create instruction funding the *owner's* ATA (`isExpectedAtaCreate` pins accounts[0]=payer signer, accounts[2]=owner); the swap instruction — including Jupiter's opaque route tail — stays owner-only via the `ownerOnly` signer set. The new negative test ("never admits the requested payer as swap authority") covers the escalation I went looking for. When `payer` is unset the URL param is absent and behavior is byte-identical to before.
- **Local oversize classification (`vault-execution.service.ts:570`)**: `simulateVaultPlan` now measures the compiled bytes and throws `VaultTransactionTooLargeError` before the RPC, so a keyed build gets the typed verdict instead of an opaque `-32602`. All existing `VaultTransactionTooLargeError` handlers (custody re-route at `vault-deposit.service.ts:393`, external deposit/withdrawal retries) catch it wherever it now originates; the try in `attemptBuild` was widened to cover the probe simulation, which is what makes the swap-funded split still reachable (tested by "splits when the composed probe is oversized before RPC simulation").
- **Compact memo retry**: 22 base64url chars = 132-bit truncated SHA-256 of the request id; retries of the same id stay byte-identical, and nothing downstream parses the memo format (the idempotency fingerprint is built from request fields, not the memo).
- **Cluster safety in the example**: `assertRpcCluster` genesis hashes match the live chains (checked against real `getGenesisHash` responses; mainnet USDC `EPjFW…Dt1v` is correct), the guard is invoked on every entry point (loadDashboard/deposit/withdraw) and cached per (rpcUrl, cluster); `pickSavingsStrategy` is cluster-aware and refuses cross-cluster pins; Explorer links correctly drop `?cluster=devnet` on mainnet.
- **Kamino deposit minimums**: the new `DEPOSIT_BELOW_MINIMUM` issue compares post-crank tokens (`tokensForSharesBaseUnits`) against `state.minDepositAmount` and reports the inclusive requested minimum — arithmetic checks out (0.1 after 0.005 crank → "at least 0.105"). The withdrawal slippage map correctly stays `null` for Kamino, matching the new `withdrawalSlippage: null` assertion.

keeganthomp:
Addressed in b0290b72b: aligned the Kamino sandbox fixtures and replay fingerprints with the required quote-derived floor, documented the sandbox and production migration, compacted the narrowed Jupiter retry, corrected the memo bit count, and fixed the demo toast fallback. The affected API suites pass 233/233, the embedded demo passes 60/60, and both typechecks pass.

cutbow7 (approved):

PR #1942, opened 2026-09-18
feat(earn): support Veda queued withdrawals end to end

cutbow7:
## Summary

- Add complete Veda instant and queued withdrawal support for custody and external wallets.
- Persist queued requests, actions, reservations, and transaction builds with tenant RLS, idempotency, and reconciliation.
- Add Treasury withdrawal route selection, request tracking, cancellation/recovery, and settlement refresh.
- Add Embedded Yield async-withdrawal examples and locked-share availability details.
- Keep async-withdrawal capabilities and UI provider-neutral and reusable.

## Behavior

- Reports instant and queued withdrawal options independently; SDP never auto-selects a route.
- Builds, submits, reads, lists, and cancels queued withdrawal requests.
- Reconciles finalized request, fulfillment, and cancellation events through a provider-owned lifecycle decoder.
- Projects solver fulfillment into the Earn movement feed with the actual assets paid.
- Preserves withdrawal access when deposits, provider surfacing, or strategy admission are paused.
- Centralizes queue status, terminal-state, copy, and idempotency behavior behind the generic async-withdrawal UI contract.
- Allows another provider to reuse the queue flow without adding provider-ID branches.

## Surfpool proof

Adds a real-program Surfpool test that forks Veda’s deployed devnet state and verifies:

1. Deposit shares into the vault.
2. Create a queued withdrawal request.
3. Confirm the request PDA and exact escrow/share balance changes.
4. Confirm the deployed Veda queue program rejects the same cancellation instruction before its deadline.
5. Advance only Surfpool’s clock beyond the deadline.
6. Land that same cancellation instruction.
7. Confirm the request closes and the exact shares are returned.

The test does not rewrite Veda programs or Veda-owned state. Solver fulfillment is not fabricated because it requires Veda’s private solve-authority signer; fulfillment parsing, reconciliation, and accounting are covered at their real SDP boundaries.

## Validation

- [x] API: 420/420 files, 6,392 passed, 14 expected skips
- [x] Remaining workspace: 4,160 passed, 10 expected environment-gated skips
- [x] Full workspace typecheck: 24/24 tasks
- [x] Veda Surfpool lifecycle: 1/1 passed
- [x] Changed-file Biome: 89 supported files, zero diagnostics
- [x] Docs link check: 55 pages, 274 links
- [x] Docs production build
- [x] Module-boundary check
- [x] API-playground drift check
- [x] Generated artifact drift check
- [x] PostgreSQL migration and tenant-isolation coverage

No live Privy calls were made during validation, so the current Privy outage did not affect these results.

## Release gates

- The new runtime routes are intentionally excluded from public OpenAPI pending named EARN-027 security approval.
- Production remains fail-closed while `VEDA_DEPLOYMENTS["mainnet-beta"]` is unset. Veda must name approved production vaults before mainnet activation.
- Mainnet activation should include a corresponding mainnet-fork Surfpool proof.

## Tooling notes

- Repository-wide `pnpm lint` still reports existing Biome configuration/schema drift and diagnostics in unchanged files; every changed supported file passes Biome.
- Standard API coverage reporting fails under Vitest 4.1.11 while parsing the unchanged coverage-threshold configuration. The complete coverage-disabled API runtime suite passes.
- React Doctor was unavailable inside the Docker environment and was not installed.


apps/sdp-api/src/db/repositories/earn-movements.repository.ts
```diff
@@ -974,6 +974,102 @@ function mapMovementRow(row: Record<string, unknown>): EarnMovementRow {
...
+  return `${QUEUED_FULFILLMENT_MOVEMENT_PREFIX}${requestId}`;
+}
+
+function queuedRequestIdFromMovementId(movementId: string): string | null {
+  return movementId.startsWith(QUEUED_FULFILLMENT_MOVEMENT_PREFIX)
+    ? movementId.slice(QUEUED_FULFILLMENT_MOVEMENT_PREFIX.length)
+    : null;
+}
+
+function mapFulfilledQueueMovement(row: Record<string, unknown>): EarnMovementRow {
```
keeganthomp, line 994: [P1] Record fulfilled payouts in the authoritative ledger

This synthetic row is only visible to repository methods that manually union `earn_vault_withdrawal_requests`. The published `/v1/transactions` Earn source still selects only `earn_movements`, so a solver-paid withdrawal appears in `/earn/movements` but disappears from the platform transaction feed, contrary to this repository's single-ledger contract. Persist one idempotent fulfillment movement keyed by the request and closing signature, or make every consumer use one shared projection; the current hand-merged model has already drifted.

apps/sdp-api/src/services/earn/vault-queued-withdrawal-reconciliation.service.ts
```diff
@@ -0,0 +1,754 @@
...
+    throw new Error(`Queued withdrawal ${request.id} closing event has a foreign identity`);
+  }
+}
+
+function assertLiveRequest(
+  request: EarnVaultWithdrawalRequestRow,
+  lookup: Exclude<EarnVaultQueuedWithdrawalRequestLookup, { status: "closedOrUnknown" }>
+): void {
+  if (
+    lookup.request.providerReference !== request.vault_address ||
```
keeganthomp, line 616: [P2] Bind provider reads to the requested address

The provider-neutral lookup returns both `lookup.requestAddress` and `lookup.request.requestAddress`, but this guard never binds either one to `request.request_address`. A future adapter can therefore return another same-owner, same-vault request and have its nonce and timing persisted; the real close event may then be rejected by the nonce check forever. Validate both returned addresses, and keep the top-level and request status coherent, before projecting provider state.

apps/sdp-api/src/db/repositories/earn-vault-withdrawal-requests.repository.ts
```diff
@@ -0,0 +1,1499 @@
...
+
+    async claimUnsettledActions(limit) {
+      if (!Number.isInteger(limit) || limit < 1 || limit > 256) {
+        throw new Error("Queued withdrawal action claim limit must be from 1 to 256");
+      }
+      const result = await db
+        .prepare(
+          `WITH candidates AS MATERIALIZED (
+             SELECT id FROM earn_vault_withdrawal_request_actions
+              WHERE status IN ('requested', 'submitted', 'confirmed')
```
keeganthomp: [P2] Stop reclaiming actions whose request is terminal

Recovery can intentionally leave an action `submitted` after PDA or closing-event truth has already moved its parent request to `fulfilled` or `cancelled`. Because this claim has no parent-status predicate, those actions enter signature reconciliation on every sweep forever; as history grows they consume the 128-row batch and make new recovery progressively slower. Join the parent request and exclude terminal parents, or record a distinct no-more-reconciliation action state.

keeganthomp (changes requested):
The provider capability split is reusable, and the migration's RLS, lease, and idempotency model looks sound. Requesting changes for one platform-visible ledger gap and two recovery/extensibility issues. Greptile's existing P2 on validating caller shares against instruction truth is also valid and should be fixed without trusting the provider quote on both sides of the comparison.

cutbow7:
Addressed all outstanding review items in d223f65e:

**React Doctor (9 warnings → 0 on changed scope)**
- Extracted a shared `EarnVaultApprovalResult` view used by both the instant and queued withdrawal modals (kills the duplicated JSX subtree).
- Queued-withdraw modal: preview fetch and request submission now live in focused hooks (`useQueuedWithdrawalPreview`, `useQueuedWithdrawalSubmission`), busy flags reset in `finally`, and the settled-request view model is derived during render instead of reconciled by an effect.
- Exit modal: split into a capability-reading hook, a forced-choice derivation, and a focused route chooser.
- The `no-pass-data-to-parent` / `no-pass-live-state-to-parent` warnings in `useEarnVaultWithdrawalRequestOutcome` are a deliberate pattern — the hook forwards server-poll lifecycle events (settlement) to the caller; annotated with `react-doctor-disable-next-line` and a reason.

**Greptile P2:** `assertQueuePlan` now validates the built share quantity directly against the caller's `terms.shares` with `compareDecimalAmounts` (decimal-aware), no longer trusting the provider's own quote on both sides.

**@keeganthomp:**
- **P1 ledger gap:** `advanceRequest` now persists one idempotent `earn_movements` fulfillment row — keyed by the request id and carrying the closing signature — in the same transaction that records `fulfilled`, so `/v1/transactions` and every direct ledger consumer see solver-paid payouts. Migration 0114 admits the observed-fulfillment shape (closing signature present, no signed outbox bytes), and the merged read paths dedupe by movement id.
- **P2 provider-read binding:** `assertLiveRequest` now binds both returned addresses (`lookup.requestAddress` and `lookup.request.requestAddress`) to the durable `request_address` and requires top-level/request status coherence before projecting provider state.
- **P2 recovery fairness:** `claimUnsettledActions` joins the parent request and excludes terminal parents (`fulfilled`/`cancelled`/`failed`), so terminal history can no longer consume the bounded recovery batch.

keeganthomp (approved):

PR #1946, opened 2026-09-19
fix(dvp): catalogue assets in the create wizard, and payouts named like parties

tobySolutions:
PRO-2016 and PRO-2015, both on the DvP create form, from feedback on the latest UI.

**The asset picker came up empty.** It listed only the org's own issued tokens, which holds while both sides are one org and breaks the moment a party is a pasted address, the case DvP exists for. It now offers the cluster's well-known tokens too, issued first, deduped by mint. Each leg already inspects its chosen mint and names what rules it out, so the list does not decide eligibility.

**A payout is now named the same three ways a party is.** It was one combobox of registered counterparties, so an org with none opened it on an empty list, and the pasted-address option only existed if you thought to type into a box labelled as a search. The payout is the party picker now, given the payout's slot, so the modes, the base58 correction and the resolver are shared and cannot drift. The two pickers stack rather than sharing a row: at half width the mode control ran into the next column.

**A chosen payout address was silently overwritten.** Editing a party re-seeded that side's payout unconditionally, so a deliberate redirect was replaced by the party address and proceeds went somewhere the form had stopped showing. Re-seeding now applies only while the payout is still the seeded default. Covered by a test that fails without the change.

The mint search wording is #1945's and is untouched here, so the two land in any order. English catalog only, per the translation catalog policy.


tobySolutions:
Superseded by #1948, which carries the same change plus the payout selector rework. Closing this one.

PR #1956, opened 2026-09-21
perf(dvp): read a trade page's funding claims in one query

cutbow7:
GET /v1/dvp/trades fanned out to one funding-claims query per trade on the page (up to 100 parallel queries against a 10-connection pool, each paying the identity-stamping BEGIN/COMMIT round trips). Add listForTrades to the funding-claim repository — one trade_id IN query, mirroring the leg-transfer ledger's batch — and use it from listTrades.

Semantics are unchanged: claim rows are RLS-filtered per row (a funding org reads its own, and the trade's owner reads every organization's claims on its trades per migration 0110), so one batched read answers exactly what the per-trade reads answered. Measured against a local testcontainer page of 100 funded trades: ~19.5ms median for the fan-out vs ~1.7ms for the batch, and the gap widens over real network latency.


multipletwigs (approved):
thanks

tobySolutions (approved):

